PR #2, opened 2019-11-07
Channel flow init

michaeljbrazell:
Adds a new initial condition and input file. Mahesh sent this to me from his implementation in IAMR


src/setup/init_fluid.f90
```diff
@@ -380,5 +381,49 @@ subroutine tuscan(lo, hi, vel, density, tracer, slo, shi, dx, dy, dz, domlo, dom
...
+         z = zlo + dz*(real(k,rt) + 0.5d0) 
+         do j = lo(2), hi(2)
+            y = ylo + dy*(real(j,rt) + 0.5d0)
+            do i = lo(1), hi(1)
+               x = xlo + dx*(real(i,rt) + 0.5d0)
+
+               vel(i,j,k,1) = 8.0d0*(5.0d0*(1.0d0-z**4)/4.0d0 + 0.3d0*cos(5.0d0*pi*y)*exp(0.5-32.4*(1.0d0-abs(z))**2)*(1.0d0-abs(z)))
+               vel(i,j,k,2) = 8.0d0*21.6d0*sin(4.0d0*pi*x)*exp(-32.4d0*(1.0d0-abs(z))**2)*(1.0d0-abs(z))
+               vel(i,j,k,3) = 0.0d0
+               tracer(i,j,k,1:ntrac) = 0.0d0
```
sayerhs: What are these for?

michaeljbrazell: the initial conditions for velocity that Mahesh used... similar to this page 9: https://www.mcs.anl.gov/~fischer/nek5000/sprague_nek5000_dec2010.pdf

michaeljbrazell: tracers would be for debugging tracers... Ann did a recent push that checks for ntrac>0

sayerhs: @michaeljbrazell I was asking about the tracers, it wasn't clear what the values `1.0 - 3.0` stood for.

michaeljbrazell: ohh, sorry. I just stole that from plane_poiseuille. It has no meaning

sayerhs: If we don't need the tracers, can we delete these lines?

michaeljbrazell: there will always be at least 1 tracer allocated... the flags come in through the input file to use more tracers incflo.ntrac and to advect the tracers incflo.advect_tracer. I was going to debug the tracer too so I'll leave just 1 tracer in there. If that's ok?

PR #3, opened 2019-11-07
Initial commit for CMake framework

jrood-nrel:
There is still a ton of stuff for me to fix with this, but it can get me started with the ability to hook up CDash, RTD, etc. Note, more of this code comes from PeleC than it does from Nalu-Wind so it's a little unorthodox. Such as almost all the CMake being confined to the `cmake` and `test` directories (Ray prefers I keep CMake confined to only a few directories in PeleC). I will fix this next and use `add_subdirectory` like a good CMake citizen. Also, each test will build its own executable in this framework, so I need to fix that next as well.

There is a bunch more we will need to discuss with how we ultimately want this organized, but I am hoping to iterate on it rather than try to nail it in the first commit. I also don't want any bad practices to hang around from this initial commit, so I'm open to fixing anything deemed important before merging.


jrood-nrel:
Oh I forgot to mention the worst part about this at the moment is that I aggregate the AMReX source files manually, which is a nightmare. I do this in PeleC to optimize the build time of all the AMReX library configurations I need, but I need to figure out the best way to build using AMReX's built-in CMake so I don't have to track their source files. I don't particularly like how they do CMake in AMReX though.

CMakeLists.txt
```diff
@@ -0,0 +1,136 @@
...
+    set(AMR_WIND_ENABLE_FCOMPARE ON)
+  endif()
+  if(AMR_WIND_TEST_WITH_FEXTREMA)
+    set(AMR_WIND_ENABLE_FEXTREMA ON)
+  endif()
+endif()
+
+########################### MASA #####################################
+
+if(AMR_WIND_ENABLE_MASA)
```
sayerhs: Can we add these as we go along instead of leaving this stuff unused?

CMakeLists.txt
```diff
@@ -0,0 +1,136 @@
...
+  set(CMAKE_PREFIX_PATH ${MASA_DIR} ${CMAKE_PREFIX_PATH})
+  find_package(MASA QUIET REQUIRED)
+  if(MASA_FOUND)
+    message(STATUS "Found MASA = ${MASA_DIR}")
+  endif()
+endif()
+
+########################### AMR-Wind #####################################
+
+if(AMR_WIND_ENABLE_VERIFICATION AND NOT AMR_WIND_ENABLE_TESTS)
```
sayerhs: Same here.

sayerhs:
The thing that I dislike most about this is what you commented on – stashing all the logic in `cmake` just to avoid `CMakeLists.txt` in the source directories. We should change that before we proceed any further. 

I am not yet sure of `GlobalSourcesList` instead of breaking down into smaller groups that have their own includes/libs to enable more parallelism during builds and less recompiles when upstream changes. But that seems like a lower priority at the moment.

jrood-nrel:
Understandable. I will fix some of this stuff today.

jrood-nrel:
I updated this to use `target_sources` and includes in each local source directory. I also removed the multiple executable functionality from Pele and now this will only test one exe in each build directory. Testing seems to work as expected.

The next thing I will have to do is hook into amrex's cmake to have it build its own library rather than me bypassing their cmake. I am find trying to do that in this PR, or merge this. I would kind of like to merge this first since there have already been some merges to `development` due to work from Ann.

sayerhs:
Yeah, merge what you have and then do the enhancements as follow-on pull requests.

PR #8, opened 2019-12-06
fixed race condition for plane average

michaeljbrazell:


michaeljbrazell:
the main change suggested by Weiqun is to do an Atomic Add 
`HostDevice::Atomic::Add(avg_fab_arr.ptr(0,0,k,n), mfab_arr(i,j,k,n)/area);`

He also said that if it is too slow we can work on a better solution

jrood-nrel:
Is this for OpenMP then? I will use the thread sanitizer sometime to check for data races. I also plan on hooking it up as a nightly test to complete my trello card.

michaeljbrazell:
The ABL case was failing on GPU's due to this. I haven't tested OpenMP yet, but hopefully it works there too!

PR #11, opened 2020-02-05
Compute kinetic energy turned back on

michaeljbrazell:
Do we want level_mask to be stored and reset during regrid? Or create it on the fly when needed? Will level mask be used for overset?


sayerhs:
@michaeljbrazell based on the discussions today, it appeared that we won't be using level mask for overset. We still need to test that plan.

PR #12, opened 2020-02-07
added MOL and Godunov TGV regression tests

michaeljbrazell:


test/CTestList.cmake
```diff
@@ -99,6 +99,8 @@ endfunction(add_test_u)
 #=============================================================================
 add_test_r(test_1 4)
 add_test_r(test_2 4)
+add_test_r(test_tgv 4)
```
sayerhs: Probably a nitpick, but we can drop the `test_` prefix. Instead call the tests `tgv_mol` and `tgv_godunov`.

jrood-nrel:
You can probably remove or rename `test_1` and `test_2` and remove the `test_` prefix from the new tests.

michaeljbrazell:
Thanks for the suggestions, I'll wait for travis and then close this.

PR #13, opened 2020-02-07
Line plot

michaeljbrazell:
This pull request cleans up the line plotting, it's still not in the greatest location and has some ABL specific stuff that needs to be fixed. This also add the ability to output a single text or binary file for line data. The text file is now default.


sayerhs:
I would suggest that the entire averaging/plotting logic be taken out of `incflo` class.

michaeljbrazell:
ok, I'll see what I can do!

src/incflo_advance.cpp
```diff
@@ -42,9 +43,19 @@ void incflo::Advance()
         }
     }
 
-    // fixme this has some specific abl stuff in it that will break for non-abl cases
-    // make this generic so plane averages can be done without abl
-    if(m_probtype==35) spatially_average_quantities_down(true);
+   if(m_plane_averaging){
+       const int axis=2;
+       PlaneAveraging *pa = new PlaneAveraging(*this, get_velocity_new(), get_tracer_new(), axis);
```
sayerhs: Why is this a pointer? And where is this pointer deallocated? Can you confirm there is no memory leak here? 

```C++
PlaneAveraging pa(*this, get_velocity_new(), get_tracer_new(), axis);
````

Comment (michaeljbrazell) [reply src/incflo_advance.cpp]: nice catch, thank you

File: src/prob/prob_init_fluid.cpp [hunk 2]
```diff
@@ -481,7 +460,8 @@ void incflo::init_abl (Box const& vbx, Box const& gbx,
               theta = m_temperature_values_d[t] + (z-m_temperature_heights_d[t])*slope;
             }
         }
-        
+#if 0 
+ figure out how to get on device
```
sayerhs: Regarding this, since this is a one-time initialization just do it on host and push it to device? That's what we do in nalu-wind

src/utilities/CMakeLists.txt
```diff
@@ -5,6 +5,10 @@ target_sources(${amr_wind_exe_name}
       incflo_build_info.cpp
       incflo_steady_state.cpp
       io.cpp
+      PlaneAveraging.cpp
+      #Headers
+      PlaneAveraging.H
```
sayerhs: headers don't need to be part of targets

src/utilities/PlaneAveraging.cpp
```diff
@@ -0,0 +1,393 @@
...
+            break;
+        default:
+            amrex::Abort("axis must be equal to 0, 1, or 2");
+            break;
+    }
+    
+}
+
+
+//void PlaneAveraging::single_level_line_plot(int s, int b, int axis, int ncomp, Vector<Real> &line){
```
sayerhs: Remove?

src/prob/prob_init_fluid.cpp
michaeljbrazell: good idea, I'll fix this on a different PR

PR #14, opened 2020-02-12
Pass some AMR-Wind options to AMReX in CMake

jrood-nrel:
This adds some code in cmake which can set the options for amrex before it gets configured. I think I have all the amrex cmake options listed.


sayerhs:
@jrood-nrel I don't think there is ever a situation/use-case where we would say `AMR_WIND_ENABLE_MPI:BOOL=OFF`. Do we even need that option?

jrood-nrel:
I almost exclusively debug AMReX applications with MPI off, but it's just my preference.

PR #21, opened 2020-03-19
Refactor ABL implementation to remove global variables in `incflo`

sayerhs:
Extract ABL implementation into its own class. Introduce the notion of an abstract Physics class/driver and implement ABL as a concrete implementation. 

This is not perfect, but at least attempts to refactor without breaking everything. Ideally, we would want the concrete implementation be the equation systems and then terms implemented as source terms.


src/wind_energy/CoriolisForcing.H
```diff
@@ -0,0 +1,42 @@
...
+        const amrex::Array4<amrex::Real const>& velocity,
+        const amrex::Array4<amrex::Real>& vel_forces) const;
+
+private:
+    ///@{
+    /** Orientation of the grid w.r.t. to planetary coordinate system
+     */
+    amrex::Vector<amrex::Real> m_east{{1.0, 0.0, 0.0}};
+    amrex::Vector<amrex::Real> m_north{{0.0, 1.0, 0.0}};
+    amrex::Vector<amrex::Real> m_up{{1.0, 0.0, 1.0}};
```
michaeljbrazell: 0,0,1

PR #22, opened 2020-03-22
Enable GPU builds via CMake

sayerhs:


sayerhs:
@jrood-nrel I removed it because one of the headers `prob_I.H` caused build errors because `nvcc` tries to process it and complains about missing classes. You bring up a good point about IDEs, I propose we revisit the design to allow `AMR-wind` itself being added as a library to a subsequent hybrid solver.

PR #24, opened 2020-03-23
tracer and velocity at n+1/2 for more accurate forcing

michaeljbrazell:
this adds a multifab that holds tracer^{n+1/2} so that the forcing functions do not have two inputs, it also uses mac velocities and a face to cell average to get veloicity at n+1/2. With the Coriolis term on this will cause differences in an ABL regression test


michaeljbrazell:
small update: I'm waiting in the queue to run some full length ABL simulation to test this branch out

PR #28, opened 2020-03-25
Updated make files 

lawrenceccheung:
I updated the `Make.package` and the `make.incflo` files in case anybody is unable to use cmake for compilation.

Lawrence

### DIFF
--- a/src/utilities/PlaneAveraging.cpp
+++ b/src/utilities/PlaneAveraging.cpp
@@ -63,7 +63,7 @@ void PlaneAveraging::plot_line_average_text(std::string filename, int step, Real
     }
 
     for(int i=0;i<ncell_line;++i){
-        outfile << step << ", " << std::scientific << time << ", " << line_xcentroid[i];
+        outfile << step << ", " << std::scientific << std::setprecision(10) << time << ", " << line_xcentroid[i];
         for(int n=0;n<navg;++n){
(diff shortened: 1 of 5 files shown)


sayerhs:
@lawrenceccheung we are looking to deprecate the `Make` file pathway. Can you explain the issue with CMake builds for you?

src/utilities/PlaneAveraging.cpp
```diff
@@ -63,7 +63,7 @@ void PlaneAveraging::plot_line_average_text(std::string filename, int step, Real
     }
 
     for(int i=0;i<ncell_line;++i){
-        outfile << step << ", " << std::scientific << time << ", " << line_xcentroid[i];
+        outfile << step << ", " << std::scientific << std::setprecision(10) << time << ", " << line_xcentroid[i];
```
sayerhs: unrelated change to the pullrequest

lawrenceccheung:
There's a requirement in the `CMakeLists.txt` for cmake >= 3.14 (also is in the amrex submodule).  The currently installed cmake at Sandia goes up to 3.12, so I'm currently using make while I ask the HPC people to upgrade cmake.

src/utilities/PlaneAveraging.cpp
lawrenceccheung: I undid the change to `PlaneAveraging`, will split it to a different pull request.

PR #32, opened 2020-03-27
added unit tests for strainrate function

michaeljbrazell:


sayerhs:
@michaeljbrazell 

Nitpick: Can we move this to `unit_tests/derive` to maintain similarity with the corresponding source code in `src/`?

PR #36, opened 2020-03-31
plane average unit tests

michaeljbrazell:
this adds unit tests for plane averaging. It tests a constant, linear interpolation, and average of a periodic function in each direction. 

I'm still having trouble compiling on Summit :( 

/ccs/home/mbrazell/amr-wind/unit_tests/utilities/test_plane_averaging.cpp(86): error: The enclosing parent function ("TestBody") for an extended __device__ lambda cannot have private or protected access within its class

/ccs/home/mbrazell/amr-wind/unit_tests/utilities/test_plane_averaging.cpp(261): error: The enclosing parent function ("TestBody") for an extended __device__ lambda cannot have private or protected access within its class

### DIFF
--- a/src/utilities/PlaneAveraging.cpp
+++ b/src/utilities/PlaneAveraging.cpp
@@ -231,8 +231,13 @@ Real PlaneAveraging::eval_line_average(Real x, int comp)
         const Real x1 = xlo + (ind+0.5)*dx;
         c = (x-x1)/dx;
     }
+    
+    if( ind+1 >= ncell_line){
+        ind = ncell_line-2;
+        c = 1.0;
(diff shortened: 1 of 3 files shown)


WeiqunZhang:
You have to make `TestBody` `public`.  It's a CUDA restriction.

sayerhs:
@michaeljbrazell you need to create a helper function to execute the actual test on GPU when running it within a GTest object. See here for an example https://github.com/Exawind/amr-wind/blob/f824cd197ff00fec976f1111d45eea63151cee04/unit_tests/wind_energy/test_abl_src.cpp#L126

PR #42, opened 2020-04-04
Possible fixes for errors in tests on the GPU

jrood-nrel:
I'm making this a draft because I'm not sure if this is the best way to solve this. Mainly I have created this PR to illustrate the errors with the tests currently occurring on the GPU. These changes allow all the tests to pass for me. I also updated the AMReX submodule to tag 20.04. I propose we update AMReX every month to their tags. You will notice some changes that had to occur for the AMReX update as well.

One note on how I found the cause of these errors on the GPU for developers new to AMReX GPU development. I saw an illegal memory access on device as the error. This usually means a global variable is trying to be read on the device. So I ran `cuda-memcheck` and it will point you to the location of the global read error. Typically I solve this by capturing the value before the `ParallelFor` and using that value in the `ParallelFor`.

### DIFF
--- a/src/diffusion/incflo_diffusion.cpp
+++ b/src/diffusion/incflo_diffusion.cpp
@@ -148,7 +148,7 @@ incflo::wall_model_bc(const int lev,
 #pragma omp parallel if (Gpu::notInLaunchRegion())
 #endif
     for (MFIter mfi(density,mfi_info); mfi.isValid(); ++mfi) {
-        const Box& bx = mfi.validbox();
+        const Box& bx = mfi.tilebox();
         auto vel = velocity.array(mfi);
         auto den = density.array(mfi);
         auto bc_a = bc.array(mfi);
@@ -264,7 +264,7 @@ incflo::heat_flux_model_bc(const int lev, const int comp, amrex::MultiFab& bc) c
 #endif
     for (MFIter mfi(bc,mfi_info); mfi.isValid(); ++mfi) {
         
-        Box const& bx = mfi.validbox();
+        Box const& bx = mfi.tilebox();
         Array4<Real> const& bc_a = bc.array(mfi);
         
         int idim = 0;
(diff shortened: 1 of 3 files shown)


jrood-nrel:
Hmm, maybe I need to make it a real PR for Travis to run.

jrood-nrel:
Oh come on. Where's Travis?

sayerhs:
@jrood-nrel Travis runs the test but seems to have lost the connection to GitHub again. 

The unit tests were running fine on GPUs on eagle and summit. Do I understand correctly that the change was due to upgrading the AMReX version?

jrood-nrel:
I was seeing the unit tests failing https://my.cdash.org/testDetails.php?test=56139336&build=1801306 . I thought this might have been due to the `setVal` happening on the device when the strainrate looks like it doesn't run on device. I have made these updates in Pele so I thought it might fix it by specifying the host for these `setVal`s. Honestly I don't fully understand the AMReX update, but this appears to have solved the unit test errors I was seeing.

jrood-nrel:
Hmm, those are for the abl_src. When I ran them manually, I was seeing strainrate tests failing. I'm a little lost here in what is actually going on, so that's why I originally made this a draft.

sayerhs:
We should discuss these changes over a call I think. Given MultiFab is shared memory this RunOn specification seems unnecessary. The failing test on CDash seems to be because it is missing GPU synchronize before the values were checked on host.

jrood-nrel:
Sounds good. Now I'll try to fix Travis again in the meantime.

jrood-nrel:
Oh I also want to make a note here. We should discuss the use of validbox and tilebox. I feel like all validboxes should likely be changed to tileboxes.

sayerhs:
@jrood-nrel I see that you're launching the GPU tests with `-j 36` and this is causing all tests to launch at the same time and attempt to run on GPUs. While I agree that this should not cause cases to fail, can you do what we did for nalu-wind and allow CMake to run only one test at a time on GPUs (so that we are not oversubscribing the GPUs)? This probably explains why the ABL unit tests are failing through CDash but weren't failing when run manually on eagle/summit.

jrood-nrel:
I use `ctest -j 36` manually when I have been testing these changes and everything passes.

jrood-nrel:
Goodbye Travis! Pull request testing looks like it's reporting again.

PR #46, opened 2020-04-05
Add option to build with MASA

jrood-nrel:
In preparation for adding verification tests, this adds the option to build with MASA. I haven't added any CI code for this yet, but we know how to install and use MASA on Travis. Hopefully we can have some lightweight verification in CI in the future.

I talked with @marchdf about this and we thought it would be best to have him help me learn enough to add the MMS verification code for this project myself.


cmake/FindMASA.cmake
```diff
@@ -0,0 +1,35 @@
...
+  masa.mod
+  HINTS ${MASA_DIR} ${CMAKE_INSTALL_PREFIX}
+  PATH_SUFFIXES lib)
+
+find_library(MASA_LIBRARY
+  NAMES masa
+  HINTS ${MASA_DIR} ${CMAKE_INSTALL_PREFIX}
+  PATH_SUFFIXES lib)
+
+find_library(MASA_FORTRAN_LIBRARY
```
sayerhs, line 27: Is this needed  for our use case? Can MASA work as a pure C/C++ library? If so, we shouldn't error out if Fortran is not available.

sayerhs:
Also I have looked at the MASA implementation in PeleC and I would like to do things a bit differently in amr-wind (e.g. global variables inside a namespace etc.).  So we should discuss before we start implementing code.

jrood-nrel:
Sounds good. We will all discuss before I start adding code. I will look at updating `FindMASA.cmake` for this and removing Fortran.

jrood-nrel:
I followed https://pabloariasal.github.io/2018/02/19/its-time-to-do-cmake-right/ and tried to create a namespace target in `FindMASA.cmake`. It seems to work. I am skeptical if I am using `INTERFACE` correctly.

cmake/FindMASA.cmake
```diff
@@ -0,0 +1,47 @@
...
+
+include(FindPackageHandleStandardArgs)
+find_package_handle_standard_args(MASA
+    REQUIRED_VARS MASA_INCLUDE_DIRS_TEMP MASA_LIBRARY_TEMP
+    VERSION_VAR MASA_VERSION
+)
+
+if(MASA_FOUND)
+    set(MASA_INCLUDE_DIRS ${MASA_INCLUDE_DIRS_TEMP})
+    set(MASA_LIBRARY ${MASA_LIBRARY_TEMP})
```
sayerhs, line 38: I am curious. What is the reason for this indirection? i.e., from TEMP variable to the actual variable?

jrood-nrel: In the example I got it from, it looks like it uses pkgconfig to populate temporary variables for headers and libraries and if they are found, then it sets the actual variables to these values. I assumed he had a reason for it, but I'm not sure why he had this intermediate step. I guess it doesn't really look necessary.

PR #48, opened 2020-04-07
remove leveldata from advance/ApplyPredictor/ApplyCorrector

michaeljbrazell:


src/incflo_advance.cpp
```diff
@@ -24,32 +25,24 @@ void incflo::Advance()
...
-            fillpatch_velocity(lev, m_t_new[lev], m_leveldata[lev]->velocity, ng);
-            fillpatch_density(lev, m_t_new[lev], m_leveldata[lev]->density, ng);
-            if (m_advect_tracer) {
-                fillpatch_tracer(lev, m_t_new[lev], m_leveldata[lev]->tracer, ng);
-            }
-        }
+
+        m_repo.get_field("velocity",amr_wind::FieldState::New).fillpatch(m_time.current_time());
+        m_repo.get_field("density",amr_wind::FieldState::New).fillpatch(m_time.current_time());
+        m_repo.get_field("tracer",amr_wind::FieldState::New).fillpatch(m_time.current_time());
```
sayerhs, line 45: To be fully consistent with the old code, we should use `m_time.new_time()` instead of `m_time.current_time()` as the old code is using `m_t_new` and not `m_t_old`

src/incflo_advance.cpp
```diff
@@ -147,110 +140,101 @@ void incflo::ApplyPredictor (bool incremental_projection)
...
-            tra_forces.emplace_back(grids[lev], dmap[lev], m_ntrac, nghost_force(),
-                                    MFInfo(), Factory(lev));
-        }
-        vel_eta.emplace_back(grids[lev], dmap[lev], 1, 1, MFInfo(), Factory(lev));
-        if (m_advect_tracer) {
-            tra_eta.emplace_back(grids[lev], dmap[lev], m_ntrac, 1, MFInfo(), Factory(lev));
-        }
-    }
+    // Allocate scratch space for tracer eqns forces and viscosity
+    // fixme, should this have an if (m_advect_tracer)  statement on it like before?
```
sayerhs, line 178: If we move to registering equations, we won't have to check `m_ntrac`

src/incflo_advance.cpp
```diff
@@ -147,110 +140,101 @@ void incflo::ApplyPredictor (bool incremental_projection)
...
-                                                  GetVecOfConstPtrs(vel_eta));
-        if (m_godunov_include_diff_in_forcing)
-            for (int lev = 0; lev <= finest_level; ++lev)
-                MultiFab::Add(vel_forces[lev], m_leveldata[lev]->divtau_o, 0, 0, AMREX_SPACEDIM, 0);
+        get_diffusion_tensor_op()->compute_divtau(divtau.vec_ptrs(),
+                                                  velocity_old.vec_const_ptrs(),
+                                                  density_old.vec_const_ptrs(),
+                                                  (*vel_eta).vec_const_ptrs());
+        if (m_use_godunov)
+            amr_wind::field_ops::add(*vel_forces, divtau, 0, 0, AMREX_SPACEDIM, 0);
```
sayerhs, line 213: So I see that you're using `divtau` instead of `divtau_old` here, do you know what was the reason to have old/new states for `divtau` and `laplacian scalar`?

src/incflo_advance.cpp
michaeljbrazell: good catch, if we are moving to the version of fill patch that does not interpolate in time, should we just remove time as an input?

src/incflo_advance.cpp
michaeljbrazell: I used the old state above but didn't label it with _old, the same goes for some other variables. I could add if that is more clear. Godunov will only need one state, MOL needs both states of divtau (and laps)

sayerhs: No, I saw your comment about `_old` for velocity, etc. but somehow didn't catch that for `divtau`. What you have is good.

src/incflo_advance.cpp
sayerhs: The AMReX routines still require a time input even if we are passing just a single field. We can provide `Field::fillpatch` overload that takes no `time` argument and then use `m_time.current_time()` for AMReX routines, current not sure if that is the best approach.

PR #50, opened 2020-04-07
Capture values into local variables in diffusion bc

jrood-nrel:
This should solve the segfaults in the regression tests.

### DIFF
--- a/src/core/FieldUtils.H
+++ b/src/core/FieldUtils.H
@@ -26,10 +26,8 @@ inline amrex::IndexType index_type(const FieldLoc floc)
         return amrex::IndexType(amrex::IntVect::TheDimensionVector(2));
     }
 
-#ifdef AMREX_USE_GPU
     // Suppress warnings when compiling with CUDA
     return amrex::IndexType::TheCellType();
-#endif
 }
 
 inline std::string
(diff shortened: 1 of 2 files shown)


src/core/FieldUtils.H
```diff
@@ -26,10 +26,8 @@ inline amrex::IndexType index_type(const FieldLoc floc)
         return amrex::IndexType(amrex::IntVect::TheDimensionVector(2));
     }
 
-#ifdef AMREX_USE_GPU
```
sayerhs, line 29: Just curious, why did you have to remove this? I had put this in because the explicit return value seemed only needed for GPU builds.

jrood-nrel: The CI system wouldn't compile because it would complain that the function was not void and not returning a value and treated it as an error.

jrood-nrel: Are you fine with this change? I can take it out and see if CI still passes. That's the only reason I changed it.

jrood-nrel: It was really surprising it was an error and not a warning.

sayerhs: Sorry, I am totally fine with this. I was just curious for the reason behind and now I understand.

PR #51, opened 2020-04-07
Update AMReX to development branch from 4-9-2020.

jrood-nrel:
This fixes the unit tests on the GPU. It should also fix the fcompare exit code when running with MPI. I am making this a draft because I don't want to merge it until I get the nightly tests which test the latest AMReX working. I want to see those build lines fail and see this PR resolve them.


jrood-nrel:
Alright, I have AMReX getting updated on the two new build lines now and it seems to be sufficiently failing. I planned to merge this but now I'm seeing an error in the CI from AMReX which I have reported in the past here https://github.com/AMReX-Codes/amrex/issues/731 . So I might need to wait for that to get fixed. I only see it on Github CI though.

jrood-nrel:
I think this is all good now. I'm going to merge it so I can work on turning `fcompare` on for CUDA.

PR #52, opened 2020-04-07
remove leveldata

michaeljbrazell:


src/core/Physics.H
```diff
@@ -16,8 +15,8 @@ public:
 
     //! Initialize the fields for the physics/problem
     virtual void initialize_fields(
-        const amrex::Geometry& geom,
-        LevelData& leveldata) const = 0;
+        int level,
+        const amrex::Geometry& geom) const = 0;
```
sayerhs, line 19: `amrex::Geometry` can go away too, the classes can access `amrex::AmrMesh`. `level` is needed because we call this from within `MakeNewLevelFromScratch`

src/incflo.H
```diff
@@ -439,8 +432,6 @@ private:
     int m_plt_divu        = 0;
     int m_plt_vfrac       = 1;
 
-    amrex::Vector<std::unique_ptr<LevelData> > m_leveldata;
-
     amrex::Vector<std::unique_ptr<amrex::FabFactory<amrex::FArrayBox> > > m_factory;
```
sayerhs, line 435: `m_factory` is no longer used anymore either.

PR #53, opened 2020-04-08
Allow Field fillpatch to handle different interpolators and use it for force fillpatch

sayerhs:
- Allow FieldFillPatchOp to handle different field interpolators
- Replace `fillpatch_force` with Field fill patch operators
- Remove unused code from `incflo`
- Add ability to let user control which fields require fillpatch operations on regrid
- Add ability to create new states after a field has been registered/declared


sayerhs:
@michaeljbrazell This should address the two issues you encountered in #52 

- Only fields with explicit fillpatch_on_regrid flag set will be updated during regrid
- Can create `n + 1/2` state directly through field and not require the use of `ScratchField`

PR #54, opened 2020-04-09
move bc's to field bc's, move Boussinesq, other clean up

michaeljbrazell:


michaeljbrazell:
I haven't tested this on GPU's yet

test/test_files/boussinesq_bubble_mol/boussinesq_bubble_mol.i
```diff
@@ -23,8 +23,11 @@ geometry.prob_hi        =  0.5 0.5 1.0  # Hi corner coordinates
 
 geometry.is_periodic    =   1   1   0   # Periodicity x y z (0/1)
 
-incflo.probtype         =  113
-incflo.gravity          = 0. 0. -0.5
+incflo.probtype         =  11
+incflo.gravity          = 0. 0. 0.5
+abl.thermal_expansion_coeff = 1.0
+abl.reference_temperature = 0.0
+bb.use_boussinesq = 1
```
sayerhs, line 31: Hmm... this is a design issue we should fix.

src/wind_energy/BoussinesqBubble.cpp
```diff
@@ -0,0 +1,71 @@
...
+{
+    amrex::ParmParse pp("bb");
+
+    pp.query("use_boussinesq", m_has_boussinesq);
+
+    // Instantiate the BoussinesqBubble field initializer
+    m_field_init.reset(new BoussinesqBubbleFieldInit());
+
+    if (m_has_boussinesq)
+        m_boussinesq.reset(new BoussinesqBuoyancy());
```
sayerhs, line 21: We should probably have `BoussinesqBuoyancy` take a `std::string` argument that allows use to pass `BoussinesqBuoyancy("abl")` in `ABL.cpp` and `BoussinesqBuoyancy("bb")` here and use the appropriate namespace to lookup parameters.

test/test_files/boussinesq_bubble_mol/boussinesq_bubble_mol.i
michaeljbrazell: do you want it fixed in this PR or later?

sayerhs: We can revisit this. just create a github issue or equivalent so that we don't forget.

PR #55, opened 2020-04-10
Update GNU makefiles

lawrenceccheung:
Still waiting for a new version of `cmake` on the Sandia machines.  In the meantime, updating the makefiles to work with GNU make.

Lawrence


sayerhs:
@lawrenceccheung Ooops... sorry. We will try to be more conscious in the future. 

@jrood-nrel I suspect if we are going to support GNUMakefile pathway, we should at least have a CI line that tests the build through those?

PR #58, opened 2020-04-12
Prepare codebase for introduction of turbulence equations

sayerhs:
Introduce the notion of PDE systems so that we can enable/disable different equation systems for wind relevant problems. Motivation is to be able to run just momentum with k-omega SST turbulence model for certain applications, while being able to run ABL simulations with 1-eqn k-sgs model and the temperature equation. 

This pull request introduces the basic skeleton and replaces all field creation logic and tracer source terms as an example. More work is needed to start replacing bits of `Apply{Predictor,Corrector}` with this interface.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -37,8 +37,8 @@ BraceWrapping:
   BeforeCatch:     false
   BeforeElse:      false
   IndentBraces:    false
-  SplitEmptyFunction: true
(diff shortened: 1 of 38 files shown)


sayerhs:
@michaeljbrazell I think this is ready for merge and then we can continue

PR #62, opened 2020-04-14
Build AMReX plotfile tools using AMReX build system

mic84:


sayerhs:
Thanks @mic84 This will need an update to the amrex git submodule to work I presume?

mic84:
@sayerhs  No, it does not. This fixes an issue with exa-wind CMake system. No need to upgrade AMReX.

PR #66, opened 2020-04-16
moved Godunov and MOL into AdvectionOp

michaeljbrazell:


src/convection/incflo_godunov_advection.cpp
```diff
@@ -85,7 +85,8 @@ godunov::compute_advection(int lev, Box const& bx, int ncomp,
         });
 
     // Use PLM to generate Im and Ip */
-    } else {   
+    } else {
+        amrex::Print() << "im in plm 2" << std::endl;
```
sayerhs, line 89: Is this leftover from debugging?

src/equation_systems/icns/icns_ops.H
```diff
@@ -80,6 +86,414 @@ struct SrcTermOp<ICNS> : SrcTermOpBase<ICNS>
...
+            pp.query("use_ppm", godunov_ppm);
+            pp.query("godunov_use_forces_in_trans", godunov_use_forces_in_trans);
+        }
+
+        { // Prefix mac
+            amrex::ParmParse pp_mac("mac_proj");
+            pp_mac.query( "mg_rtol", mac_mg_rtol);
+            pp_mac.query( "mg_atol", mac_mg_atol);
+            pp_mac.query( "mg_max_coarsening_level", mac_mg_max_coarsening_level);
+        }
```
sayerhs, line 106: Replace this with `MLMGOptions`

src/incflo_field_repo.cpp
```diff
@@ -14,9 +14,8 @@ void incflo::declare_fields()
         "ICNS-" + scheme, m_time, m_repo, m_probtype);
     m_scalar_eqns.emplace_back(amr_wind::pde::PDEBase::create(
         "Temperature-" + scheme, m_time, m_repo, m_probtype));
-
-    const int nstates = 2;
-    m_repo.declare_cc_field("conv_density", 1, 0, nstates);
+    m_scalar_eqns.emplace_back(amr_wind::pde::PDEBase::create(
```
sayerhs, line 17: This only needs to happen if `m_constant_density == false` correct?

src/equation_systems/icns/icns_ops.H
```diff
@@ -80,6 +86,414 @@ struct SrcTermOp<ICNS> : SrcTermOpBase<ICNS>
...
+
+        // TODO: Need iconserv flag to be adjusted???
+        iconserv.resize(ICNS::ndim, 0);
+    }
+
+    void operator()(const FieldState fstate, const amrex::Real dt)
```
sayerhs, line 112: Eventually we will want to break this into smaller pieces. When we do that we can address the MAC projection duplication between Godunov/MOL.

src/incflo_field_repo.cpp
michaeljbrazell: I tried that and it didn't work... I think because it creates density_conv_term, should I put if statements to use conv_density when m_constant_density = false, and density_conv_term otherwise?

src/convection/incflo_godunov_advection.cpp
michaeljbrazell: yep

src/incflo_field_repo.cpp
sayerhs: Ah I see. I think this will go away when we move more stuff into equation systems.

PR #68, opened 2020-04-17
Use AMReX's CMake function for generating build info

jrood-nrel:
Remove python dependency and use AMReX's CMake function for AMReX_buildInfo.cpp.


sayerhs:
@jrood-nrel Wanted to check with you to see if you were waiting for something else in AMReX before merging this.

PR #71, opened 2020-04-17
WIP moving source terms into SrcTermOp

michaeljbrazell:
WIP I didn't get around to making source terms subclasses of momentumSource, I also commented out the ABL source term unit tests temporarily

### DIFF
--- a/src/wind_energy/BoussinesqBuoyancy.cpp
+++ b/src/wind_energy/BoussinesqBuoyancy.cpp
@@ -1,4 +1,5 @@
 #include "BoussinesqBuoyancy.H"
+#include "FieldUtils.H"
 
 #include "AMReX_ParmParse.H"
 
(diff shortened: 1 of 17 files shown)


src/wind_energy/BoussinesqBuoyancy.cpp
```diff
@@ -39,15 +41,20 @@ BoussinesqBuoyancy::BoussinesqBuoyancy()
...
+    const FieldState fstate,
     const amrex::Array4<amrex::Real>& vel_forces) const
 {
     const amrex::Real T0 = m_ref_theta;
     const amrex::Real beta = m_beta;
     const amrex::GpuArray<amrex::Real, AMREX_SPACEDIM> gravity{
         m_gravity[0], m_gravity[1], m_gravity[2]};
 
+    FieldState tstate = field_impl::phi_state(fstate);
+    const auto& scalars = m_repo.get_field("temperature", tstate)(lev).const_array(mfi);
```
sayerhs, line 56: So the idea would be that we would declare a class member 

```C++
// in class header
Field& m_temperature;

BoussinesqBuoyancy()
  : m_repo(repo)
  , m_temperature(repo.get_field("temperature"))
```

So that we can just use the field directly without doing a `get_field` within the `MFIter` loop

michaeljbrazell: to do this we'll have to move the physics out of the read parameters and after the declare_fields, but I'm guessing you already have a plan for that too...

PR #72, opened 2020-04-18
Refactor predictor/corrector code to use equation system interface

sayerhs:
- Refactor MAC projection code to eliminate duplication in Godunov/MOL advection operators

- Create an RHS computation operator to deal with update (for explicit) and RHS terms for implicit/Crank-Nicolson, update Predictor/Corrector with the new functions

- Convert calls to `tracer` update with loops of scalar equations.


sayerhs:
I am going to resubmit this pull-request after fixing the conflicts introduced due to `plm` modifications.

PR #92, opened 2020-04-28
Rayleigh Taylor fix and unit tests

michaeljbrazell:
-changed sign in DensityBuoyancy which will cause major diff in RayleighTaylor regression test
-converted unit tests to new source terms

### DIFF
--- a/unit_tests/wind_energy/test_abl_src.cpp
+++ b/unit_tests/wind_energy/test_abl_src.cpp
@@ -288,7 +288,7 @@ TEST_F(ABLMeshTest, boussinesq_nph)
 
     auto& src_term = pde_mgr.icns().fields().src_term;
 
-    auto& temperature = sim().repo().get_field("temperature").create_state(amr_wind::FieldState::NPH);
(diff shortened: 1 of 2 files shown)


michaeljbrazell:
we should coordinate this with @jrood-nrel  since it will break regression tests

unit_tests/wind_energy/test_abl_src.cpp
```diff
@@ -282,48 +223,107 @@ void init_abl_temperature_field(int kdim,
...
+    auto& pde_mgr = sim().pde_manager();
+    pde_mgr.register_icns();
+    pde_mgr.register_transport_pde("Temperature");
+    sim().init_physics();
+
+    amr_wind::pde::icns::BoussinesqBuoyancy bb(sim());
+
+    auto& src_term = pde_mgr.icns().fields().src_term;
+
+    auto& temperature = sim().repo().get_field("temperature").create_state(amr_wind::FieldState::NPH);
```
sayerhs, line 291: This is not needed anymore. `register_transport_pde("Temperature")` already creates `n + 1/2` states for you.

PR #94, opened 2020-04-30
variable viscosity CFL

michaeljbrazell:
- modified the CFL to account for variable viscosity
- added an input that turns on/off using variable forcing in CFL calculation
- preliminary testing looks like with explicit diffusion we can run at a max CFL=2 and with implicit diffusion at a max CFL=5, just because we can doesn't mean we should since Godunov is probably robust enough to go beyond what we want for accuracy. So currently we still require CFL <= 1, unless further testing is done. 
- warning this will tweak the abl_godunov_explicit regression test since it does not use a fixed dt, need to rebless gold files and time this with @jrood-nrel

### DIFF
--- a/src/incflo_compute_dt.cpp
+++ b/src/incflo_compute_dt.cpp
@@ -22,74 +22,103 @@ using namespace amrex;
 //
 // WARNING: We use a slightly modified version of C in the implementation below
 //
-void incflo::ComputeDt (bool explicit_diffusion)
+void incflo::ComputeDt(bool explicit_diffusion)
 {
     BL_PROFILE("amr-wind::incflo::ComputeDt")
 
     Real conv_cfl = 0.0;
     Real diff_cfl = 0.0;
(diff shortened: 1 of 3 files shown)


src/incflo_compute_dt.cpp
```diff
@@ -22,74 +22,103 @@ using namespace amrex;
...
+                rho, mu, 0,
+                [=] AMREX_GPU_HOST_DEVICE(
+                    Box const& b, Array4<Real const> const& rho_arr,
+                    Array4<Real const> const& mu_arr) -> Real {
+                    Real mx = -1.0;
+                    amrex::Loop(b, [=, &mx](int i, int j, int k) noexcept {
+                        mx = amrex::max(
+                            2.0 * mu_arr(i, j, k) *
+                                (dxinv[0] * dxinv[0] + dxinv[1] * dxinv[1] +
+                                 dxinv[2] * dxinv[2]) /
```
sayerhs, line 71: Could be precomputed?

PR #95, opened 2020-04-30
plane averaging improvements

michaeljbrazell:
- added viscosity line averaging back in
- changed plane averaging to output header for time step 1 instead of 0

### DIFF
--- a/src/utilities/PlaneAveraging.H
+++ b/src/utilities/PlaneAveraging.H
@@ -20,7 +20,9 @@ public:
     ~PlaneAveraging(){};
 
     void operator()(const amrex::Vector<amrex::Geometry>& Geom,
+               amrex::Vector<amrex::MultiFab*> const& density,
                amrex::Vector<amrex::MultiFab*> const& velocity,
+               amrex::Vector<amrex::MultiFab*> const& mueff,
                amrex::Vector<amrex::MultiFab*> const& tracer);
 
     /** evaluate line average at specific location for velocity xdir*/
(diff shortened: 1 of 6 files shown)


src/utilities/PlaneAveraging.H
```diff
@@ -20,7 +20,9 @@ public:
     ~PlaneAveraging(){};
 
     void operator()(const amrex::Vector<amrex::Geometry>& Geom,
+               amrex::Vector<amrex::MultiFab*> const& density,
                amrex::Vector<amrex::MultiFab*> const& velocity,
+               amrex::Vector<amrex::MultiFab*> const& mueff,
```
sayerhs, line 25: Is it time for `PlaneAveraging` to become aware of `FieldRepo`?

src/utilities/PlaneAveraging.cpp
```diff
@@ -100,7 +100,11 @@ void PlaneAveraging::plot_line_binary(std::string filename, int step, Real time)
 }
 
 template<typename IndexSelector>
-void PlaneAveraging::fill_line(const IndexSelector &idxOp, const amrex::MultiFab& velocity,  const amrex::MultiFab& tracer)
+void PlaneAveraging::fill_line(const IndexSelector &idxOp,
+                               const amrex::MultiFab& density,
+                               const amrex::MultiFab& velocity,
+                               const amrex::MultiFab& mueff,
+                               const amrex::MultiFab& temperature)
```
sayerhs, line 107: I think it will be useful to split off `PlaneAveraging` utility into two parts: 1. that handles velocity and its first/second moments, and 2. a new class that can take any scalar and compute its averages, first/second moment... not just tempertaure.

src/utilities/PlaneAveraging.H
michaeljbrazell: I think so!

src/utilities/PlaneAveraging.cpp
michaeljbrazell: sounds good, I'll put it on the todo list

PR #96, opened 2020-05-01
Adding utilities to compute the gradient & curvature of a scalar function

gdeskos:
- Added ABL & TGV examples + make files
- Added compute_gradient + compute curvature in derive_K.H


sayerhs:
@gdeskos Also would it be possible to add unit tests for these two functions?

PR #98, opened 2020-05-01
cleanup inputs

michaeljbrazell:
- removed more unused inputs 
- and functions
- added a FreeStream Physics class to help eliminate init fluid
- now probtype=0 by default and not required in input file unless probtype=31
- amr.line_plot_int -> io.line_plot_int
- amr.KE_int -> io.KE_int


src/wind_energy/FreeStream.H
```diff
@@ -0,0 +1,44 @@
+#ifndef FreeStream_H
```
sayerhs, line 1: I think we should move physics stuff out of `wind_energy` into its own `physics` directory. Thoughts?

PR #99, opened 2020-05-02
input docs

michaeljbrazell:
updated example input file in docs and made some changes to inputs


docs/sphinx_doc/user/inputs.rst
```diff
@@ -159,14 +163,530 @@ This section deals with parameters that control the simulation.
...
+   
+   Height to evaluate the log law for the wall model. 
+   Currently, if this parameter is not specified in the input file, the first half cell height is calculated
+   and is used to set the log law height. 
+   Therefore the log law height depends on the domain size and number of elements in the normal direction. 
+   If this parameter is set to a number the log law is evaluated at a fixed height. 
+   Note: currently the fluctuating velocity terms in the shear stress model are only 
+   available at the first cell center above the wall. This limitation will be removed soon.
+   
+.. input_param:: abl.ntemperature
```
sayerhs, line 448: Technically `ParmParse` doesn't need this. It will automatially resize the `amrex::Vector`, we should probably add a check to ensure that `temperature_heights` and `temperature_values` are the same length.

src/wind_energy/ABL.cpp
```diff
@@ -86,11 +86,26 @@ void ABL::pre_advance_work()
...
         // TODO: This should be handled by PlaneAveraging
         int output_interval = 1;
+        int plot_type = 0;
         amrex::ParmParse pp("io");
         pp.query("line_plot_int", output_interval);
+        pp.query("line_plot_type", plot_type);
 
         if ((output_interval > 0) && (time.time_index() % output_interval == 0)) {
-            m_pa.plot_line_text("line_plot.txt", time.time_index(), time.current_time());
+            switch (plot_type) {
```
sayerhs, line 95: Why not consolidate this logic in `PlaneAveraging` and just pass `plot_type` to the `m_pa.process_outputs()` method?

michaeljbrazell: good idea, I can do that

docs/sphinx_doc/user/inputs.rst
michaeljbrazell: I was worried about the 0 length vectors inside the field initialization. But maybe those are ok too. I'll check

PR #102, opened 2020-05-06
Adding MMS capability using MASA

marchdf:
In addition to the MMS implementation, this Includes:
- unit tests
- regression tests
- documentation


docs/sphinx_doc/dev/verification.rst
```diff
@@ -0,0 +1,81 @@
...
+
+Performing the MMS verification
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+For an MMS verification study, one performs a convergence study using
+an MMS input file, several examples of which can be found in the
+regression test suite. The following results for the Godunov FV method
+exhibit the expected second order accuracy:
+
+- Velocity (u, v, w) :math:`L_2` error norms:
```
sayerhs, line 68: Nitpick. Might be useful to mention here that the error norms are output to a text file and the default name of the text file?

src/mms/MMS.cpp
```diff
@@ -0,0 +1,188 @@
...
+        }
+
+        const auto& dx = m_mesh.Geom(lev).CellSizeArray();
+        const auto& problo = m_mesh.Geom(lev).ProbLoArray();
+        const amrex::Real cell_vol = dx[0] * dx[1] * dx[2];
+
+        using FuncDef = double (*)(double, double, double);
+        auto compute_error = [this, &lev, &level_mask, &problo, &dx,
+                              &cell_vol](int idx, FuncDef f) {
+            amrex::Real error = amrex::ReduceSum(
```
sayerhs, line 142: We should revisit this reduce call. If the innermost function is going to run on CPU, launching reduction kernels on device don't make sense.

test/CTestList.cmake
```diff
@@ -125,6 +125,11 @@ add_test_r(boussinesq_bubble_mol 4)
 add_test_r(boussinesq_bubble_godunov 4)
 add_test_r(abl_mol 4)
 add_test_r(abl_godunov 4)
+if(AMR_WIND_ENABLE_MASA)
```
sayerhs, line 128: We should exclude these tests from CI. Currently, CI builds don't enable MASA

unit_tests/mms/mms_test_utils.H
```diff
@@ -0,0 +1,21 @@
+#ifndef MMS_TEST_UTILS_H
+#define MMS_TEST_UTILS_H
+
+#include "aw_test_utils/MeshTest.H"
+#include "aw_test_utils/pp_utils.H"
+
+namespace amr_wind_tests {
+
+class MMSTest : public AmrexTest
```
sayerhs, line 9: Can remove this. MMS test will always require an underlying AmrCore instance.

jrood-nrel:
I've been working on adding MASA to PeleC's CI in Github actions, so I will add that here after I get it settled.

marchdf:
Actually please don't merge quite yet. There's one more test I would like to add.

src/mms/MMS.cpp
marchdf: Any suggestions for that @sayerhs ? I moved things around a bit...

marchdf:
ugh meant that as a comment... the boxes look the same

@sayerhs and @michaeljbrazell I just added a new MMS test for the error computation.

src/mms/MMS.cpp
sayerhs: I think we can do this entirely on host instead of going through the `ReduceSum` kernel launch. That means using an accumulation variable, MFIter loops with `amrex::Loop` within that handles the per-rank logic and then doing MPI reductions on the variable. I think the `compute_error` function can go away :(

docs/sphinx_doc/dev/verification.rst
```diff
@@ -0,0 +1,83 @@
...
+   $ spack install masa
+
+Linking MASA to AMR-Wind
+########################
+
+The following CMake options enable MASA in AMR-Wind:
+
+.. code-block:: bash
+
+   -DAMR_WIND_ENABLE_VERIFICATION=ON -DAMR_WIND_ENABLE_MASA=ON -DMASA_DIR=$MASA_DIR
```
michaeljbrazell, line 56: it might be less confusing if you use MASA_ROOT_DIR like above. Or vice-versa so they are consistent 

mkdir build
cd build
cmake -DAMR_WIND_ENABLE_VERIFICATION=ON -DAMR_WIND_ENABLE_MASA=ON -DMASA_DIR=$MASA_ROOT_DIR ..

src/mms/MMS.cpp
marchdf: Ok I took a stab at it. Let me know if you see ways to improve it.

docs/sphinx_doc/dev/verification.rst
marchdf: good idea, let me do that

PR #105, opened 2020-05-08
Handle restart/output properly for the PDE/turbulence interface

sayerhs:
This pull-request introduces a new IOManager that will handle I/O

- Introduce `io.outputs` that takes a list of strings and uses that to determine fields for output
- Allow PDE, Physics, and Turbulence instances to register default fields for output and restart
- Integrate IOManager with CFDSim to allow other classes to access and update default I/O 
- Replace old I/O interface with the IOManager calls
- Fixes restart with arbitrary number of fields (e.g., turbulence models)

Merging this will break nightly tests as the field names have changed, and requires an update of the gold files.


sayerhs:
@jrood-nrel Let us know when you're ready to merge this, will require reblessing of gold files.

jrood-nrel:
We can merge this.

sayerhs:
@jrood-nrel This is merged... will need a re-bless of golds for nightly.

PR #109, opened 2020-05-11
Add wall model capability to x and y boundaries

michaeljbrazell:
- adds wall model capability to x and y boundaries
- removes extrapolate for velocity and density since we want to evaluate velocities at cell center
- this function will be removed once we add the general version that works for arbitrary heights

### DIFF
--- a/src/diffusion/incflo_diffusion.cpp
+++ b/src/diffusion/incflo_diffusion.cpp
@@ -121,12 +121,12 @@ void wall_model_bc(
     auto& viscosity = repo.get_field("velocity_mueff");
     const int nlevels = repo.num_active_levels();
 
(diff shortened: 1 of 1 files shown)


michaeljbrazell:
What should we do with density?

sayerhs:
Is there any way to reduce the code duplication? Not critical for this pull request as it would be eventually replaced by the new algorithm, but we will have to deal with it for the new one and other places.

sayerhs:
Also, by density do you mean the nph stuff?

michaeljbrazell:
I'll take a stab at reducing duplication. I was thinking about the density in the mu*dudz = rho*utau^2 term

PR #111, opened 2020-05-12
Compile fix for MMS unit test

marchdf:
Something broke the unit test compiles for MMS... Here's the fix.


sayerhs:
Sorry, this was my mistake. We need coverage with MMS turned on. I'll switch over to an MMS-enabled build type

PR #116, opened 2020-05-13
Add utilities to compute scalar and vector gradients

gantech:
Replaces https://github.com/Exawind/amr-wind/pull/96

Also added unit tests for scalar and vector gradient computations on both interior and boundary nodes


gantech:
There seems to be a memory leak in VectorGradientTest. Please hold until I fix it.

gantech:
Ok... this pull request is good to go now I think

PR #122, opened 2020-05-15
fix restart

michaeljbrazell:
adding post_init_actions to restart so that ABL restarts correctly


michaeljbrazell:
do we want this or create a post_restart_actions?

PR #124, opened 2020-05-16
Add Debug build to CI matrix

jrood-nrel:
Adding `Debug` build to CI test matrix which runs only the unit tests. Making this a draft because I will likely need to iterate on it to get CI working as expected.


sayerhs:
@jrood-nrel just curious, is there a specific reason to exclude debug build the on Mac? Was it just an optimization?

jrood-nrel:
Yeah it felt a little redundant. I don't mind leaving it in, but since I think we get 20 total cores, I think the build might take resources from the rest of the jobs. I'm not certain though.

PR #125, opened 2020-05-18
new plane average

michaeljbrazell:
changes:
- added new plane average that takes a field
- removed old plane average and replaced with new one except for line plot output
- added a second moment class that takes two plane averages as inputs

questions:
- It might make sense to separate out the line plot files for each field, but I wasn't sure what was happening with netcdf. What do think?
- Second moment has some redundancy when you make Reynolds stress by inputing plane average twice you get <u'v'> and <v'u'> etc. But this makes it generic for any two planes... keep it this way?
- if we like this make a third moment?


src/utilities/FieldPlaneAveraging.H
```diff
@@ -0,0 +1,81 @@
...
+#ifndef FieldPlaneAveraging_H
+#define FieldPlaneAveraging_H
+
+
+#include <AMReX_AmrCore.H>
+#include "DirectionSelector.H"
+#include "Field.H"
+
+class incflo;
+
```
sayerhs, line 15: Use `amr_wind` namespace.

src/utilities/FieldPlaneAveraging.H
```diff
@@ -0,0 +1,81 @@
...
+class FieldPlaneAveraging
+{
+public:
+    FieldPlaneAveraging(amr_wind::Field &field, int axis_in = 2);
```
sayerhs, line 19: This should probably be templated to accept both `Field` and `ScratchField`

src/utilities/FieldPlaneAveraging.H
```diff
@@ -0,0 +1,81 @@
...
+//    /** output a single text file with averages and fluctuations */
+//    void plot_line_text(std::string filename, int step, amrex::Real time);
+//    /** output a single binary file with averages and fluctuations */
+//    void plot_line_binary(std::string filename, int step, amrex::Real time);
+//
+//    /** output a line plot file based on the plot type */
+//    void plot_line(int step, amrex::Real time, int plot_type = 0);
+
+    /** change precision of text file output */
+    void set_precision(int p) {precision = p;};
```
sayerhs, line 40: Separate I/O concerns from logic.

src/utilities/FieldPlaneAveraging.H
```diff
@@ -0,0 +1,81 @@
...
+
+    int ncell_plane; /** number of cells in plane */
+    int ncell_line; /** number of cells along line */
+
+    int precision = 4; /** precision for line plot text file */
+    const int level = 0; /** level for plane averaging for now fixed at level=0 */
+
+
+    amr_wind::Field& field;
+    const int axis;
```
sayerhs, line 70: Use `m_<varname>` syntax for member variables.

src/utilities/FieldPlaneAveraging.H
```diff
@@ -0,0 +1,81 @@
...
+//    void plot_line_binary(std::string filename, int step, amrex::Real time);
+//
+//    /** output a line plot file based on the plot type */
+//    void plot_line(int step, amrex::Real time, int plot_type = 0);
+
+    /** change precision of text file output */
+    void set_precision(int p) {precision = p;};
+
+    amrex::Real get_dx(){return dx;};
+    amrex::Real get_xlo(){return xlo;};
```
sayerhs, line 43: For all methods 

- Use `const` qualifiers 
- Prefer `dx()` instead of `get_*`

src/utilities/FieldPlaneAveraging.H
```diff
@@ -0,0 +1,81 @@
...
+    int get_axis(){return axis;};
+    int get_level(){return level;};
+    int get_line_avg_size() {return line_average.size();};
+    int get_ncell_plane(){return ncell_plane;};
+    int get_ncell_line(){return ncell_line;};
+    amrex::Real * get_line_avg_data() {return line_average.data();};
```
sayerhs, line 49: Why not return `amrex::Vector<amrex::Real>` here? That way the calling site automatically can query the size.

src/utilities/FieldPlaneAveraging.cpp
```diff
@@ -0,0 +1,143 @@
...
+
+
+#include "FieldPlaneAveraging.H"
+#include <algorithm>
+#include "incflo.H"
+
+using namespace amrex;
+
+
+Real FieldPlaneAveraging::eval_line_average(Real x, int comp) const
```
sayerhs, line 13: For me this method name is ambiguous. It is not computing the averages on the line, instead it is interpolating a previously computed average at a given point along the line.

src/utilities/FieldPlaneAveraging.cpp
```diff
@@ -0,0 +1,143 @@
...
+            break;
+        default:
+            amrex::Abort("axis must be equal to 0, 1, or 2");
+            break;
+    }
+
+}
+
+template<typename IndexSelector>
+void FieldPlaneAveraging::avg_line(const IndexSelector &idxOp,
```
sayerhs, line 112: `compute_averages` or `evaluate_averages` might be a better name here.

src/utilities/SecondMoment.H
```diff
@@ -0,0 +1,63 @@
...
+    int ncell_plane; /** number of cells in plane */
+    int ncell_line;  /** number of cells along line */
+
+    int precision = 4; /** precision for line plot text file */
+
+    int ncomp1;
+    int ncomp2;
+
+    FieldPlaneAveraging& pa1;
+    FieldPlaneAveraging& pa2;
```
sayerhs, line 46: Provide a more descriptive name here.

src/utilities/SecondMoment.H
```diff
@@ -0,0 +1,63 @@
...
+    amrex::Vector<amrex::Real> second_moments_line; /** line storage for fluctuations */
+
+    int ncell_plane; /** number of cells in plane */
+    int ncell_line;  /** number of cells along line */
+
+    int precision = 4; /** precision for line plot text file */
+
+    int ncomp1;
+    int ncomp2;
+
```
sayerhs, line 44: Can access this from `FieldPlaneAveraging::field`?

src/utilities/SecondMoment.H
```diff
@@ -0,0 +1,63 @@
...
+    /** change precision of text file output */
+    void set_precision(int p) {precision = p;};
+
+private:
+
+    int nfluc; /** outer product of components */
+    amrex::Vector<amrex::Real> second_moments_line; /** line storage for fluctuations */
+
+    int ncell_plane; /** number of cells in plane */
+    int ncell_line;  /** number of cells along line */
```
sayerhs, line 38: Can access from `FieldPlaneAveraging`?

src/utilities/SecondMoment.H
```diff
@@ -0,0 +1,63 @@
...
+    int precision = 4; /** precision for line plot text file */
+
+    int ncomp1;
+    int ncomp2;
+
+    FieldPlaneAveraging& pa1;
+    FieldPlaneAveraging& pa2;
+
+    const int axis;
+    const int level;
```
sayerhs, line 49: Don't need these variables?

src/utilities/SecondMoment.cpp
```diff
@@ -0,0 +1,138 @@
...
+            break;
+        default:
+            amrex::Abort("axis must be equal to 0, 1, or 2");
+            break;
+    }
+
+}
+
+template<typename IndexSelector>
+void SecondMoment::fill_line(const IndexSelector &idxOp, const amrex::MultiFab &mfab1, const amrex::MultiFab &mfab2)
```
sayerhs, line 53: `compute_moments`?

src/utilities/SecondMoment.H
```diff
@@ -0,0 +1,63 @@
...
+#define SecondMoment_H
+
+
+#include <AMReX_AmrCore.H>
+#include "DirectionSelector.H"
+#include "Field.H"
+#include "FieldPlaneAveraging.H"
+
+class incflo;
+
```
sayerhs, line 17: In `amr_wind` namespace

michaeljbrazell:
Thanks Shreyas!, I think I got everything except for templating for field and scratchfield

PR #128, opened 2020-05-20
restart reg test

michaeljbrazell:
added a reg test that uses a restart from a previously run reg test


michaeljbrazell:
One down side to this approach is that it requires running abl_godunov before abl_godunov_restart

jrood-nrel:
We can set a dependency on the test, or we can make a restart test category.

jrood-nrel:
I think we can add this to get the dependency, but I haven't been able to test it yet:
```
# Regression test and excluded from CI with dependency
function(add_test_red TEST_NAME NP TEST_DEPENDENCY)
    add_test_re(${TEST_NAME} ${NP})
    set_tests_properties(${TEST_NAME} PROPERTIES FIXTURES_REQUIRED ${TEST_DEPENDENCY})
endfunction(add_test_red)
```
and
```
add_test_red(abl_godunov_restart 4 abl_godunov)
```

PR #129, opened 2020-05-21
bug fix for average and moments

michaeljbrazell:
use local variables inside parallel_for


michaeljbrazell:
@lawrenceccheung can you please try this?

PR #132, opened 2020-05-22
field plane average and moments

michaeljbrazell:
- trying field plane average again starting only with unit tests
- I think I resolved the eagle gpu unit test failure
- I still haven't found the bug causing a NaN Lawrence was seeing with his intel build or what was causing small diffs in regression tests for certain builds


src/utilities/FieldPlaneAveraging.cpp
```diff
@@ -0,0 +1,201 @@
+#include "FieldPlaneAveraging.H"
+#include <algorithm>
+#include "incflo.H"
+
+using namespace amrex;
+using namespace amr_wind;
```
sayerhs, line 6: use 
```
namespace amr_wind {

}
``` instead

File: src/utilities/FieldPlaneAveraging.cpp [hunk 2]
```diff
@@ -0,0 +1,201 @@
+#include "FieldPlaneAveraging.H"
+#include <algorithm>
+#include "incflo.H"
+
+using namespace amrex;
```
sayerhs, line 5: Is it possible to eliminate this?

src/utilities/FieldPlaneAveraging.cpp
```diff
@@ -0,0 +1,201 @@
...
+        Box bx = mfi.tilebox();
+
+        auto fab_arr = mfab.const_array(mfi);
+
+        amrex::ParallelFor(
+            bx, ncomp,
+            [=] AMREX_GPU_DEVICE(int i, int j, int k, int n) noexcept {
+                const int ind = idxOp(i, j, k);
+                HostDevice::Atomic::Add(
+                    &line_avg[ncomp * ind + n], fab_arr(i, j, k, n) * denom);
```
sayerhs, line 194: `denom` can be taken outside?

src/utilities/SecondMomentAveraging.cpp
```diff
@@ -0,0 +1,249 @@
...
+        m_plane_average1.ncell_plane() == m_plane_average2.ncell_plane());
+    AMREX_ALWAYS_ASSERT(
+        m_plane_average1.ncell_line() == m_plane_average2.ncell_line());
+
+    // this could be relaxed if we wanted one plane to be at a different time
+    // step than another plane
+    AMREX_ALWAYS_ASSERT(
+        m_plane_average1.last_updated_index() ==
+        m_plane_average2.last_updated_index());
+
```
sayerhs, line 86: this shouldn't be in the constructor. The user could hold references to this object and call `operator()(...)` repeatedly

src/utilities/FieldPlaneAveraging.cpp
michaeljbrazell: yes but it will cause a small order of operations diff, I'll make a note and fix it later

PR #141, opened 2020-05-28
Prepare codebase for DPC++

sayerhs:
Replace `std::` math functions with `amrex::Math` versions in preparation for DPC++


sayerhs:
@WeiqunZhang these were the only ones I could find. I've updated them all with `amrex::Math` versions. 

@asalmgren Let me know which variables you were thinking of as global variables.

PR #142, opened 2020-05-28
growntilebox

michaeljbrazell:
- changed the compute_* functions to use tilebox and not a grown one
- this requires a fillpatch afterwards. 
- Added a two level static refinment ABL regression test
- upgraded strain rate unit test to quadratic polynomials
- removed random variable from analytical function and loosened scalar laplacian tolerance 
- tilebox/fillpatch approach gives a different solution than using grown tile box when number of levels > 0


src/turbulence/LES/Smagorinsky.H
```diff
@@ -31,6 +31,9 @@ private:
 
     const Field& m_vel;
     const Field& m_rho;
+
+    const amr_wind::SimTime& m_time;
```
sayerhs, line 35: The base class TurbModelBase already has a reference to CFDSim

gdeskos:
So do we still have to lower the tolerance for the laplacian to pass after you removed the random number factor from the grid spacing ? Also while there, can you apply the attached patch ? This is to fix a bug in field_algebra.cpp

diff --git a/src/derive/field_algebra.cpp b/src/derive/field_algebra.cpp
index 3982c28..0b50a78 100644
--- a/src/derive/field_algebra.cpp
+++ b/src/derive/field_algebra.cpp
@@ -23,7 +23,7 @@ void normalize_field(FType& Field)
                 }
                 //Normalize field
                 for (int icomp=0;icomp<ncomp;++icomp){
-                    field_arr(i,j,k)=field_arr(i,j,k,icomp)/std::sqrt(mag);
+                    field_arr(i,j,k,icomp)=field_arr(i,j,k,icomp)/std::sqrt(mag);
                 }
 
                 });

michaeljbrazell:
@gdeskos yes I needed to remove random and adjust tolerance. fixed the icomp bug, thanks!

PR #144, opened 2020-05-30
Add a Geostrophic Wind Forcing term

gantech:
Create a forcing term to enforce a Geostrophic Wind

### DIFF
--- a/src/equation_systems/icns/source_terms/GeostrophicForcing.cpp
+++ b/src/equation_systems/icns/source_terms/GeostrophicForcing.cpp
@@ -12,9 +12,8 @@ namespace icns {
 
 /** Geostrophic forcing term for ABL 
  *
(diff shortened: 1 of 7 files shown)


src/equation_systems/icns/source_terms/GeostrophicForcing.cpp
```diff
@@ -0,0 +1,77 @@
...
+        // Read the geostrophic wind speed vector (in m/s)
+        amrex::ParmParse pp("GeostrophicForcing");
+        pp.getarr("geostrophic_wind", m_target_vel);
+    }
+
+    amrex::Real rho=1.0;
+    {
+        // Read the density
+        amrex::ParmParse pp("incflo");
+        pp.query("density", rho);
```
sayerhs, line 49: Change this to use the density field within the operator

PR #146, opened 2020-05-31
Make AMR-Wind installable

sayerhs:
This pull-request makes the header includes work both inside source directory as well as when used from an installed location for use with amr-wind as a library for other codes, e.g. hybrid amr/nalu coupling. 

- Adds headers to include rules
- Adds custom target that allows IDEs to discover header files
- Adds CMake config rules that allows `find_package(AMR-Wind)` for other codes using CMake


sayerhs:
@gdeskos @gantech I tried merging this with both your WIP branches and it was a straightforward merge. So, hopefully, this doesn't disrupt your workflow too much. 🤞

PR #150, opened 2020-06-02
Add tolerance for fcompare when using CUDA

jrood-nrel:
Since GPU results are nondeterministic, this adds a 1e-12 tolerance bound when CUDA is enabled to fix test failures from the gold file comparison. This also fixes the default override for test checkpoint and plot intervals I initially had wrong.

### DIFF
--- a/test/CTestList.cmake
+++ b/test/CTestList.cmake
@@ -33,10 +33,13 @@ function(add_test_r TEST_NAME)
     # Copy files to test working directory
     file(COPY ${TEST_FILES} DESTINATION "${CURRENT_TEST_BINARY_DIR}/")
     # Set some default runtime options for all tests in this category
-    set(RUNTIME_OPTIONS "time.max_step=10 amr.plot_file=plt amr.checkpoint_files_output=0 amr.plot_files_output=1 amrex.throw_exception=1 amrex.signal_handling=0")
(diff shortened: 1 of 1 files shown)


test/CTestList.cmake
```diff
@@ -33,10 +33,13 @@ function(add_test_r TEST_NAME)
     # Copy files to test working directory
     file(COPY ${TEST_FILES} DESTINATION "${CURRENT_TEST_BINARY_DIR}/")
     # Set some default runtime options for all tests in this category
-    set(RUNTIME_OPTIONS "time.max_step=10 amr.plot_file=plt amr.checkpoint_files_output=0 amr.plot_files_output=1 amrex.throw_exception=1 amrex.signal_handling=0")
+    set(RUNTIME_OPTIONS "time.max_step=10 amr.plot_file=plt time.plot_interval=10 time.checkpoint_interval=10 amrex.throw_exception=1 amrex.signal_handling=0")
```
sayerhs, line 36: did you mean to disable `checkpoint` files here? Previously that was 0 now it is 10.

jrood-nrel: Now that we have a restart test, I can't globally disable checkpoints. (that option didn't work anyway since it was for PeleC and names have changed here) I would like to disable checkpoints for tests though.

jrood-nrel: Actually I realize I need to redesign the restart test because it complicates these runtime options even further.

jrood-nrel: Ok I think I figured it out and will just rely on each test to appropriately set the checkpoint intervals if a restart test depends on it, but still force plot intervals.

PR #156, opened 2020-06-04
Add verfication test category and add MMS 2nd order verification test

jrood-nrel:
This allows anyone to run @marchdf 's MMS 2nd order verification. This test does require many Python packages to be available at run time, such as matplotlib and pandas. The test produces a `plots.pdf` file and is uploaded to CDash. Currently the python script doesn't appear to automatically check for 2nd order, so the test passes if the PDF is generated. Probably should add such a check in the future.

### DIFF
--- a/test/CTestList.cmake
+++ b/test/CTestList.cmake
@@ -5,6 +5,10 @@ if(AMR_WIND_TEST_WITH_FCOMPARE)
   message(STATUS "Test golds directory for fcompare: ${FCOMPARE_GOLD_FILES_DIRECTORY}")
 endif()
 
+if(AMR_WIND_ENABLE_MASA AND NOT AMR_WIND_ENABLE_MPI)
+  message(WARNING "Running verification tests without MPI enabled will require long run times")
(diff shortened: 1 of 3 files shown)


sayerhs:
So I suppose this means I've to train myself to add `-LE verification` with my CTest runs when developing on Mac?

jrood-nrel:
Yeah I feel the same way. Is there a way to not run a test by default? Let me look.

sayerhs:
I think that would require an `AMR_WIND_ENABLE_VERIFICATION_TESTS` flag?

jrood-nrel:
Yeah the only thing for test properties I see is explicitly setting a test to disabled, which won't work. I guess that's why I have an `ENABLE_MASA` and `ENABLE_VERIFICATION` in Pele. I can add it back in if you would rather do it that way.

sayerhs:
Let's try out the pull-request as it stands for a while. If it feels too painful, we can revisit and add the CMake option to disable it.

PR #157, opened 2020-06-04
Address most of the current warnings from cppcheck static analysis

jrood-nrel:


amr-wind/incflo_advance.cpp
```diff
@@ -41,8 +41,8 @@ void incflo::Advance()
     ApplyPredictor();
 
     if (!m_use_godunov) {
-        auto& vel = icns().fields().field;
-        vel.state(amr_wind::FieldState::New).fillpatch(m_time.current_time());
+        auto& vel2 = icns().fields().field;
```
sayerhs, line 44: This is not needed... you can use `vel` from the outer scope.

PR #159, opened 2020-06-08
Fix CTV for GPU

marchdf:


marchdf:
I tested this on an Eagle GPU node and seems to runs just fine.

amr-wind/physics/ConvectingTaylorVortex.cpp
```diff
@@ -161,8 +162,10 @@ amrex::Real ConvectingTaylorVortex::compute_error(
 
 void ConvectingTaylorVortex::output_error()
 {
-    const amrex::Real u_err = compute_error(0, m_velocity, u_exact);
-    const amrex::Real v_err = compute_error(1, m_velocity, v_exact);
+    UExact u_exact;
+    VExact v_exact;
```
sayerhs, line 166: you don't need these. You can remove the first and third argument from `compute_error` and just do 

```C++
compute_error<UExact>(m_velocity);
```

marchdf: let me try that!

marchdf: Ok I fixed that. Thanks for seeing that

PR #166, opened 2020-07-07
General-purpose body force

gdeskos:
General-purpose body force which supports

- A constant force e.g. F=(1,0,0)
- An oscillatory force e.g. F=(1,0,0)*sin(time)

### DIFF
--- a/amr-wind/equation_systems/icns/source_terms/BodyForce.H
+++ b/amr-wind/equation_systems/icns/source_terms/BodyForce.H
@@ -1,0 +1,47 @@
+#ifndef BODYFORCE_H
+#define BODYFORCE_H
+
+#include "amr-wind/equation_systems/icns/MomentumSource.H"
+#include "amr-wind/core/SimTime.H"
+
+namespace amr_wind {
+namespace pde {
+namespace icns {
+
+class BodyForce : public MomentumSource::Register<BodyForce>
+{
+public:
+    static const std::string identifier() { return "BodyForce"; }
+
+    BodyForce(const CFDSim&);
+
+    virtual ~BodyForce();
+
+    virtual void operator()(
+        const int lev,
+        const amrex::MFIter& mfi,
+        const amrex::Box& bx,
+        const FieldState fstate,
+        const amrex::Array4<amrex::Real>& src_term) const override;
+
+private:
+    const Field& m_density;
+   
+    //! Forcing source term (pressure gradient)
+    amrex::Vector<amrex::Real> m_body_force{{0.0, 0.0, 0.0}};
+
+    //! Body Force Type
+    std::string m_type;
+    
+    //! Angular frequency used in the oscillatory forcing
+    amrex::Real m_omega;
+
+    //! Time
+    const SimTime& m_time;
+};
+
+} // namespace icns
+} // namespace pde
+} // namespace amr_wind
+
+#endif /* BODYFORCE_H */
--- a/amr-wind/equation_systems/icns/source_terms/BodyForce.H
+++ b/amr-wind/equation_systems/icns/source_terms/BodyForce.H
@@ -28,9 +28,6 @@ private:
     //! Time
     const SimTime& m_time;
 
-    //! Density
-    const Field& m_density;
-
     //! Forcing source term (pressure gradient)
     amrex::Vector<amrex::Real> m_body_force{{0.0, 0.0, 0.0}};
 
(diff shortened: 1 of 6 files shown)


amr-wind/equation_systems/icns/source_terms/BodyForce.cpp
```diff
@@ -0,0 +1,60 @@
...
+namespace pde {
+namespace icns {
+
+/** Body Force
+ */
+BodyForce::BodyForce(const CFDSim& sim)
+    : m_time(sim.time()), m_density(sim.repo().get_field("density"))
+{
+
+    {
```
sayerhs, line 18: Don't need this additional scope

amr-wind/equation_systems/icns/source_terms/BodyForce.H
```diff
@@ -0,0 +1,47 @@
...
+        const amrex::Array4<amrex::Real>& src_term) const override;
+
+private:
+    const Field& m_density;
+
+    //! Forcing source term (pressure gradient)
+    amrex::Vector<amrex::Real> m_body_force{{0.0, 0.0, 0.0}};
+
+    //! Body Force Type
+    std::string m_type;
```
sayerhs, line 34:
```C++
std::string m_type{"constant"};
```

Since you're are using `pp.query` and not forcing it to be mandatory.

amr-wind/equation_systems/icns/source_terms/BodyForce.cpp
```diff
@@ -0,0 +1,60 @@
...
+        // Read the geostrophic wind speed vector (in m/s)
+        amrex::ParmParse pp("BodyForce");
+        pp.query("type", m_type);
+        pp.queryarr("magnitude", m_body_force);
+        if (m_type == "Oscillatory" || m_type == "oscillatory") {
```
sayerhs, line 23: I would recommend doing this instead:

```C++
pp.query("type", "m_type");
m_type = amrex::toLower(m_type);

if (m_type == "oscillatory") pp.query("angular_frequency", m_omega);
``` 

so that user can also do `OsCiLlaToRy` and still have it work

amr-wind/equation_systems/icns/source_terms/BodyForce.cpp
```diff
@@ -0,0 +1,60 @@
...
+            pp.query("AngularFrequency", m_omega);
```
sayerhs, line 24: We have been using camel case to indicate runtime selection of sub-classes and snake case for input variables. Your choice if you want to use camel case for variable names.

amr-wind/equation_systems/icns/source_terms/BodyForce.H
```diff
@@ -0,0 +1,47 @@
...
+
+    //! Angular frequency used in the oscillatory forcing
+    amrex::Real m_omega;
```
sayerhs, line 37:
```C++
amrex::Real m_omega{0.0};
```

Initialize to proper value for use with `pp.query`

amr-wind/equation_systems/icns/source_terms/BodyForce.cpp
```diff
@@ -0,0 +1,60 @@
...
+    amrex::GpuArray<amrex::Real, AMREX_SPACEDIM> forcing{
+        {m_body_force[0], m_body_force[1], m_body_force[2]}};
+
+    amrex::Real coeff;
+
+    if (m_type == "constant" || m_type == "Constant") {
+        coeff = 1.;
+    } else if (m_type == "Oscillatory" || m_type == "oscillatory") {
+        coeff = std::sin(m_omega * time);
+    }
```
sayerhs, line 49: converting using `amrex::toLower` after read will simplify this check. 

```C++
amrex::Real coeff = (m_type == "oscillatory") ? std::sin(m_omega * time) : 1.0;
```

amr-wind/equation_systems/icns/source_terms/BodyForce.cpp
```diff
@@ -0,0 +1,60 @@
...
+    const auto& rho = m_density.state(fstate)(lev).const_array(mfi);
+    amrex::ParallelFor(bx, [=] AMREX_GPU_DEVICE(int i, int j, int k) noexcept {
+        src_term(i, j, k, 0) += rho(i, j, k) * coeff * forcing[0];
```
michaeljbrazell, line 52: I don't think density should be here, what do you think @sayerhs?

gdeskos: Yes, actually that's right... you can assume that rho is already incorporated in the forcing

PR #170, opened 2020-07-10
Update source to build with Intel DPC++

sayerhs:
Make changes to sources so that they will build properly with Intel OneAPI DPC++ compiler and introduce GitHub CI workflow to test this on the development branch.

- Add workaround for compiler bug in beta07. Will revert this when this is fixed in future updates
- Convert math functions to use `amrex::Math::` versions instead of `std::` versions


sayerhs:
I am going to post this here for future reference: https://github.com/rscohn2/oneapi-ci

PR #173, opened 2020-07-13
Update CMake installation rules for hybrid-solver coupling strategy

sayerhs:
- Create a shared library for external linking

- Add necessary CMake install rules to allow linking either via external AMReX
  or internal AMReX build

- Allow AMR-Wind to find its own dependencies in Config mode


cmake/set_compile_flags.cmake
```diff
@@ -18,12 +18,12 @@ endif()
 # Add our extra flags according to language
 separate_arguments(AMR_WIND_CXX_FLAGS)
 target_compile_options(
-  ${amr_wind_lib_name} PUBLIC
+  ${amr_wind_lib_name} PRIVATE
```
jrood-nrel, line 21: Just want to double check this is intentional.

sayerhs: @jrood-nrel Yes. Currently this is all warning flags and it is leaking those flags to downstream. I suppose we should have a way to separate out `warning` flags from other flags.

PR #177, opened 2020-07-21
Enable fortran as a cmake language again to fix GPU builds

jrood-nrel:
I was able to build with CUDA again by just adding fortran as a language. Doesn't seem that the commit that removed it will need to be reverted.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 ############################ BASE ######################################
 
 cmake_minimum_required (VERSION 3.14 FATAL_ERROR)
-project(AMR-Wind CXX C Fortran)
+project(AMR-Wind CXX C)
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake")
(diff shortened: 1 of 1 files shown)


jrood-nrel:
I guess it becomes a problem for DPC++.

CMakeLists.txt
```diff
@@ -37,6 +37,10 @@ set(CMAKE_CXX_STANDARD 14)
 set(CMAKE_CXX_EXTENSIONS OFF)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 
+if(NOT AMR_WIND_ENABLE_DPCPP)
+  enable_language(Fortran)
+endif()
+
 if(AMR_WIND_ENABLE_CUDA)
   enable_language(CUDA)
```
sayerhs, line 45:
```suggestion
if(AMR_WIND_ENABLE_CUDA)
  enable_language(CUDA)
  # Fix issues with GPU builds
  enable_language(Fortran)
```

jrood-nrel: Ah thanks!

PR #192, opened 2020-08-20
Implement vorticity function in fvm operators

gantech:


sayerhs:
@gantech for AMR we will need vorticity as a vector. Can you rename this as VorticityMag so that we can use vorticity for the vector when we get to implementing that? Although you have all the parts to implement both

gantech:
So are we good then?

sayerhs:
missing `clang-format` ;)

PR #203, opened 2020-09-07
Fix bug in CFL computation and clean up CFL/time output in log file

sayerhs:
This PR fixes a bug in the combined CFL computation logic in amr-wind https://github.com/Exawind/amr-wind/blob/4c033b9323e9a72dae9ab9ab56b265cf0847b648/amr-wind/incflo_compute_dt.cpp#L119-L122

There is an extra factor of 2.0 that shouldn't be there. With this bug-fix, we recover the exact CFL (and timestep) in the initial time when there is only advection term and there are no contributions from diffusion or source terms. 

In addition to the bug fix, the following changes clean up output of time/cfl details into log file

- CFL components are always printed (previously required `verbose > 2`)
- CFL components are scaled by the current `deltaT` instead of printed at unit time. 
- Time step output now clearly specifies the old/new time 

Will need re-bless of adaptive time-stepping and explicit diffusion tests.

### DIFF
--- a/amr-wind/core/SimTime.H
+++ b/amr-wind/core/SimTime.H
@@ -49,11 +49,13 @@ public:
      */
     bool write_last_checkpoint();
 
-    /** Set current CFL and update timestep based on CFL
+    /** Set current CFL and update timestep based on CFL components
      *
-     *  @param cfl_unit_time `(U / dx)`, i.e., CFL assuming unit dt
      */
-    void set_current_cfl(amrex::Real cfl_unit_time);
+    void set_current_cfl(
+        const amrex::Real conv_cfl,
+        const amrex::Real diff_cfl,
+        const amrex::Real src_cfl);
 
     /** Set start time and index based on checkpoint/restart file
      *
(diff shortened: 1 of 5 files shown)


sayerhs:
New output formats 

*Intro*
```
Begin simulation:
  Run until 22000 sec. or 10 timesteps
  Fixed timestepping with dt = 0.5; max. CFL from inputs = 0.95
```

*Fixed timestepping*
```
Step: 4 dt: 0.5 Time: 1.5 to 2
CFL: 0.171305 (conv: 0.170274 diff: 0 src: 0.00035306 )
```

*Adaptive timestepping*
```
Step: 10 dt: 0.004072284545 Time: 0.0234524 to 0.0275247
CFL: 0.133092 (conv: 0.129947 diff: 0 src: 0.102782 )
```

*Explicit diffusion*
```
Step: 10 dt: 0.7237723895 Time: 4.16822 to 4.89199
CFL: 0.250057 (conv: 0.245945 diff: 0.00411261389377 src: 0 )
```

michaeljbrazell (approved):
looks good to me, it does break the abl_godunov_explicit reg test but that is expected since it is the only non fixed dt case @jrood-nrel

PR #208, opened 2020-09-17
Preliminary overset implementation for AMR-Wind 

sayerhs:
This PR introduces necessary changes in the codebase to facilitate overset simulations with Nalu-Wind and AMR-Wind using TIOGA library. 

- Refactor time advance methods to interject overset connectivity and solution exchange steps at appropriate times from the driver
- Update MAC projection to provide a *pressure-like* solution in presence of overset constraints
- Change RHS computation to skip overset fringe/hole cells during update 
- Update linear solver interfaces to register overset mask and use it during linear solves.

### DIFF
--- a/amr-wind/incflo_advance.cpp
+++ b/amr-wind/incflo_advance.cpp
@@ -58,15 +58,7 @@ void incflo::advance()
 
     ApplyPredictor();
 
-    if (!m_use_godunov) {
-        vel.state(amr_wind::FieldState::New).fillpatch(m_time.current_time());
-        for (auto& eqn: scalar_eqns()) {
-            auto& field = eqn->fields().field;
-            field.state(amr_wind::FieldState::New).fillpatch(m_time.current_time());
-        }
-
-        ApplyCorrector();
-    }
+    if (!m_use_godunov) ApplyCorrector();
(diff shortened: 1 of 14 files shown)


amr-wind/incflo_advance.cpp
```diff
@@ -52,27 +36,37 @@ void incflo::advance()
...
+ * \callgraph
+ */
+void incflo::advance()
+{
+    BL_PROFILE("amr-wind::incflo::Advance");
     ApplyPredictor();
 
-    if (!m_use_godunov) {
-        vel.state(amr_wind::FieldState::New).fillpatch(m_time.current_time());
-        for (auto& eqn: scalar_eqns()) {
```
michaeljbrazell, line 63: should we keep this around?

michaeljbrazell: regression tests pass so maybe this was not necessary?

michaeljbrazell (approved):

amr-wind/incflo_advance.cpp
sayerhs: Yeah, the logic is now handled in post solve ops.

PR #213, opened 2020-10-05
Ekman spiral

michaeljbrazell:
- added Ekman spiral initial condition
- currently only valid for wind direction in x-dir only (could be probably be generalized if needed)
- only valid for latitude=90 degrees due to geostrophic forcing
- added example test file, do we want this as a regression test?


amr-wind/physics/EkmanSpiral.H
```diff
@@ -0,0 +1,78 @@
...
+};
+
+struct VExact
+{
+    AMREX_GPU_DEVICE AMREX_FORCE_INLINE amrex::Real
+    operator()(const amrex::Real, const amrex::Real, const amrex::Real) const;
+    const int m_comp{1};
+};
+
+} // namespace
```
sayerhs, line 26: Move this to the `.cpp` file, no need to expose them in the header file.

lawrenceccheung (approved):

sayerhs:
@michaeljbrazell To answer your question, I think it would be good to add this as a regression test. Just to make sure that the EkmanSpiral physics class doesn't rot.

PR #247, opened 2020-11-04
added QCriterionRefinment

michaeljbrazell:
I still need to add to the docs

### DIFF
--- a/amr-wind/utilities/tagging/QCriterionRefinement.cpp
+++ b/amr-wind/utilities/tagging/QCriterionRefinement.cpp
@@ -43,6 +43,8 @@ void QCriterionRefinement::operator()(
 {
     const bool tag_field = level <= m_max_lev_field;
 
+    if (!tag_field) return;
+
     m_vel->fillpatch(level, time, (*m_vel)(level), 1);
 
     const auto& mfab = (*m_vel)(level);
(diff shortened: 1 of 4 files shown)


amr-wind/utilities/tagging/QCriterionRefinement.cpp
```diff
@@ -0,0 +1,154 @@
...
+        size_t fcount = std::min(qc_value.size(), m_qc_value.size());
+        for (size_t i = 0; i < fcount; ++i) m_qc_value[i] = qc_value[i];
+        m_max_lev_field = fcount - 1;
+    }
+}
+
+void QCriterionRefinement::operator()(
+    int level, amrex::TagBoxArray& tags, amrex::Real time, int)
+{
+    const bool tag_field = level <= m_max_lev_field;
```
sayerhs, line 44:
```suggestion
    const bool tag_field = level <= m_max_lev_field;
    if (!tag-field) return;
```

Do an early return here and don't check it inside the MFIter loop. This will also avoid doing a needless fillpatch operation

amr-wind/utilities/tagging/QCriterionRefinement.cpp
```diff
@@ -0,0 +1,154 @@
...
+#ifdef _OPENMP
+#pragma omp parallel if (Gpu::notInLaunchRegion())
+#endif
+    for (amrex::MFIter mfi(mfab, amrex::TilingIfNotGPU()); mfi.isValid();
+         ++mfi) {
+        const auto& bx = mfi.tilebox();
+        const auto& tag = tags.array(mfi);
+        const auto& vel = mfab.const_array(mfi);
+
+        if (tag_field) {
```
sayerhs, line 60: This is not needed if we do an early return.

amr-wind/utilities/tagging/QCriterionRefinement.cpp
michaeljbrazell: good catch, I fixed it

PR #255, opened 2020-11-10
Changes to WENOJS

nataraj2:
Making the hybrid WENO/PPM scheme consistent with https://arxiv.org/pdf/1902.06665.pdf (Equation 12 and 16)


sayerhs:
@nataraj2 It would have been better if this was added as an option instead of changing the existing scheme.

gdeskos (approved):

gdeskos:
> @nataraj2 It would have been better if this was added as an option instead of changing the existing scheme. Or was there a discussion with @gdeskos and it was decided to switch to WENO-JS scheme permanently?

That was a small correction to the WENO-JS scheme not a switch to the WENO-Z that we discussed last week. This is to make WENO-JS consistent with the PELEc WENO/PPM scheme

PR #262, opened 2020-11-26
Custom interior and inflow BC initialization

sayerhs:
This PR generalizes the Dirichlet inflow interface to allow custom user-defined functions (UDFs) instead of the constant values that were possible before. It also generalizes the FreeStream class to allow initialization of different fields and use UDFs for field initialization also.


gantech (approved):
Thanks! I tested this and works great!

michaeljbrazell (approved):

PR #264, opened 2020-11-29
Fix build errors in single precision mode

sayerhs:
- [x] Fix errors from `amrex::min` and `amrex::max` template instantiations
- [x] Wait for amrex-codes/amrex#1577 
- [x] Fix `amrex::ReduceRealSum` in `SamplingContainer.cpp` (wait for #263)
- [x] Update NetCDF interface to handle float/double I/O
- [x] Fix inconsistent usage of `double`/`amrex::Real` in `ABLStats.cpp`
- [ ] Streamline `tol`/`eps` usage in code and unit tests


jrood-nrel (approved):

jrood-nrel:
If we need single precision in the future, I will pick this work back up.

PR #268, opened 2020-12-04
Move documentation from RTD to GitHub pages

jrood-nrel:
While I'm at it I thought I would move AMR-Wind to Github pages and I will do Nalu-Wind next. I'm still using CMake to generate docs but I modified it a bit. I put the Doxygen output in the Sphinx `html` directory and then deploy that. I will need to update the URLs to have all the links work if/when this would get merged. I noticed the `gh-pages` branch history with the Doxygen graphs has bloated the repo so this should fix that by enforcing a single commit in that branch.


jrood-nrel:
Closing this to remove cmake dependency and resubmit.

PR #270, opened 2020-12-04
added CTV dpdx to error calculation/output

michaeljbrazell:


amr-wind/physics/ConvectingTaylorVortex.H
```diff
@@ -33,6 +33,54 @@ struct VExact
...
+        const amrex::Real,
+        const amrex::Real,
+        const amrex::Real,
+        const amrex::Real,
+        const amrex::Real,
+        const amrex::Real) const;
+    const int m_comp{2};
+};
+
+struct gpxExact
```
sayerhs, line 48: Can we switch these names to be consistent with how we define struct/class in AMR-Wind? 

```suggestion
struct GpxExact
```

michaeljbrazell: thanks, fixed it

sayerhs (approved):

PR #271, opened 2020-12-05
Move documentation from RTD to Github pages

jrood-nrel:
I reduced the dependencies to cmake, sphinx, and optionally doxygen. I'm open to discussion on whether or not cmake should be a dependency here, but cmake helps simplify the github action in my opinion. This feels minimal enough that a user could generate things manually fairly easily. It was easy for me to port what I already had to this configuration. I just made the docs a separate cmake project with no compiler dependencies.

Making this a draft at first because I didn't test it on my fork before making this PR.


sayerhs:
So I tried expanding the `api-docs.yml` we already have in the repo to also handle `sphinx` docs (see #273). It didn't seem like CMake was really needed. Can you elaborate on why CMake is needed for the Github CI workflow? 

Personally, I feel that the CMake-based workflow is not conducive to incremental development/docs generation:

- `add_custom_target` is always out-of-date, therefore, if the user just wants to update user manual... it will still trigger compilation of API docs... this can be time-consuming especially if the user is fixing minor typos. We can break the dependency by removing the `add_dependencies` 
- In the current setup, Sphinx is mandatory but Doxygen is not. So if doxygen is not found there is no warning about this. 
- Moving `Doxyfile` to `Doxyfile.in` 
  - We also have to update the [documenting source code section](https://amr-wind.readthedocs.io/en/latest/developer/documentation.html#documenting-source-code).
  - If user relies on CMake to generate `Doxyfile` then they must have `sphinx` installed even if all they care about is generating API docs.

jrood-nrel:
#273 is fine with me. I guess my thought was that the user could still do whatever they want. Having the recipe for building the documentation encoded into cmake seemed more convenient to me. Otherwise, I have to look up all the steps to type into the command line to recreate it as it's deployed here. The docs are simple enough that I don't have a strong enough preference to refine  this PR any further and the yml file can just be the reference as to how to build them. If we don't use cmake, I would prefer to get rid of all the extra stuff in the docs folder after #273 is merged though.

jrood-nrel:
All of a sudden the docs workflow won't run. I'm going to try to reopen this.

PR #274, opened 2020-12-07
Move documentation from readthedocs to github pages

jrood-nrel:


.github/workflows/docs.yml
```diff
@@ -0,0 +1,51 @@
+name: AMR-Wind Docs
+
+ on:
```
sayerhs, line 3: @jrood-nrel so there is a syntax error in the file... this node and subsequent nodes are offset by 1 space. So it appears as a sub-dictionary of `name` and that's why I think the workflow didn't run when you submitted the PR.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,51 @@
...
+       - name: Dependencies
+         run: |
+           export DEBIAN_FRONTEND=noninteractive
+           sudo apt-get update
+           sudo apt-get install -y --no-install-recommends \
+             doxygen graphviz
+           python3 -m pip install --upgrade pip
+           pip3 install sphinx sphinx_rtd_theme
+       - name: Build
+         working-directory: ${{github.workspace}}
```
sayerhs, line 37: Not necessary

.github/workflows/docs.yml
```diff
@@ -0,0 +1,51 @@
...
+         run: |
+           sphinx-build -M html ./docs/sphinx ./build_docs/manual
+           doxygen ./docs/doxygen/Doxyfile
+           mv ./build_docs/manual/html ./documentation
+           mv ./build_docs/doxygen/html ./documentation/api_docs
+           touch ./documentation/.nojekyll
+       - name: Deploy
+         if: github.event_name == 'push' && github.ref == 'refs/heads/development'
```
sayerhs, line 45:
```suggestion
         if: github.event_name == 'push'
```

Can be simplified as the workflow won't run on any other branches anyway due to `on:` rules.

.github/workflows/docs.yml
jrood-nrel: Yes not necessary, but as someone who would reference this to build the docs myself, I think it's helpful to see the working directory explicitly.

.github/workflows/docs.yml
jrood-nrel: Ah, damn. Copy and paste foiled me.

.github/workflows/docs.yml
sayerhs: Sure... just requires that person to know whether `github.workspace` refers to the parent directory into which the project is cloned by `actions/checkout@v2` or the root directory of the git repo. 

May be a comment to this effect?

```
# execute from top amr-wind directory
```

jrood-nrel: Heh yeah...

sayerhs (approved):

PR #276, opened 2020-12-09
Formatting check

jrood-nrel:
This adds a check that `clang-format` has been applied to all source files. It also adds an action to each workflow which cancels previous workflow ones when new pushes originate. I also formatted the source files to pass the format test.

One note about this, is that I wish I could have other workflows depend on the formatting check so it's done first before the heavyweight actions start. This only currently seems possible with lots of archaic code, or by putting all workflows into a single file with dependencies. So my hope is that the cancelling actions will help save resources when the formatting fails quickly and someone pushes the formatted files.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,14 +51,18 @@ jobs:
             build_type: Debug
 
     steps:
+    - name: Cancel previous runs
+      uses: styfle/cancel-workflow-action@0.6.0
+      with:
+        access_token: ${{github.token}}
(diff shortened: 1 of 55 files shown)


jrood-nrel:
My test PR here https://github.com/jrood-nrel/amr-wind/pull/2 seems to show that the cancelling works as I would expect.

jrood-nrel:
I just noticed it really gives you a lot of emails when it cancels previous runs.

sayerhs (approved):

PR #278, opened 2020-12-11
Add more compiler warning output and a few more things

jrood-nrel:
1. Adds more compiler warning output.
2. Attempts to fix newly seen warnings.
3. Adds Werror option in CMake.
4. Enables Werror option in CI.
5. Removes hardcoded CPU core count in CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,7 +61,7 @@ jobs:
     - name: Configure
       run: |
         cmake \
-          -B${{runner.workspace}}/build-ci-${{matrix.build_type}}
+          -B${{runner.workspace}}/build-ci-${{matrix.build_type}} \
           -DCMAKE_INSTALL_PREFIX:PATH=${{runner.workspace}}/install-${{matrix.build_type}} \
           -DCMAKE_BUILD_TYPE:STRING=${{matrix.build_type}} \
           -DAMR_WIND_ENABLE_MPI:BOOL=ON \
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,12 +22,6 @@ on:
       - 'CMakeLists.txt'
       - '.github/workflows/ci.yml'
 
-env:
-  CXX_COMPILER: mpicxx
-  C_COMPILER: mpicc
-  FORTRAN_COMPILER: mpifort
-  NUM_PROCS: 16
-
 jobs:
   build:
     runs-on: ${{matrix.os}}
@@ -39,9 +33,11 @@ jobs:
           - os: macos-latest
             install_deps: brew install open-mpi
             comp: llvm
+            procs: $(sysctl -n hw.ncpu)
           - os: ubuntu-latest
             install_deps: sudo apt-get install mpich libmpich-dev
             comp: gnu
+            procs: $(nproc)
           - build_type: RelWithDebInfo
             ctest_args: -LE no_ci
           - build_type: Debug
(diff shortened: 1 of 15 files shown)


jrood-nrel:
Actually the additional warning flags might need some logic since they're not all supported by older compilers.

amr-wind/physics/ConvectingTaylorVortex.cpp
```diff
@@ -103,8 +103,8 @@ ConvectingTaylorVortex::ConvectingTaylorVortex(const CFDSim& sim)
     pp.query("error_log_file", m_output_fname);
     {
         amrex::Real nu;
-        amrex::ParmParse pp("transport");
-        pp.query("viscosity", nu);
+        amrex::ParmParse pp2("transport");
+        pp2.query("viscosity", nu);
```
sayerhs, line 107: In future we should just introduce a scope in the previous block. 

```c++
{
    amrex::ParmParse pp("CTV");
    pp.query("density", m_rho);
    pp.query("u0", m_u0);
    pp.query("v0", m_v0);
    pp.query("activate_pressure", m_activate_pressure);
    pp.query("error_log_file", m_output_fname);
}
```

cmake/set_compile_flags.cmake
```diff
@@ -1,20 +1,29 @@
...
   endif()
-  if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU" AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 7.0)
-    # Avoid notes about -faligned-new with GCC > 7
-    list(APPEND AMR_WIND_CXX_FLAGS "-faligned-new")
+  if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU" OR
+     "${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang" OR
+     "${CMAKE_CXX_COMPILER_ID}" STREQUAL "AppleClang")
+     if(CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 7.0)
+       list(APPEND AMR_WIND_CXX_FLAGS "-faligned-new"
+                                      "-Wno-sign-compare"
```
sayerhs, line 13: Is disabling sign-compare warnings really necessary? I would prefer we kept that.

cmake/set_compile_flags.cmake
```diff
@@ -1,20 +1,29 @@
...
+  if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Intel")
     # Intel always reports some diagnostics we don't necessarily care about
     list(APPEND AMR_WIND_CXX_FLAGS "-diag-disable:11074,11076,15335")
   endif()
-  if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU" AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 7.0)
-    # Avoid notes about -faligned-new with GCC > 7
-    list(APPEND AMR_WIND_CXX_FLAGS "-faligned-new")
+  if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU" OR
+     "${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang" OR
+     "${CMAKE_CXX_COMPILER_ID}" STREQUAL "AppleClang")
```
sayerhs, line 10: Would this syntax work?
```cmake
if ("${CMAKE_CXX_COMPILER_ID}" IN_LIST "GNU;Clang;AppleClang") 
```

cmake/set_compile_flags.cmake
```diff
@@ -1,20 +1,29 @@
...
+                                      "-Wnull-dereference"
+                                      "-Wfloat-conversion"
+                                      "-Wshadow"
+                                      "-Woverloaded-virtual")
+    endif()
   endif()
 endif()
 
+if(AMR_WIND_ENABLE_WERROR)
+  list(APPEND AMR_WIND_CXX_FLAGS "-Werror")
```
sayerhs, line 24: While we are at, we should also add `-fno-operator-names` here.

sayerhs (approved):

.github/workflows/ci.yml
```diff
@@ -61,19 +57,22 @@ jobs:
     - name: Dependencies
       run: ${{matrix.install_deps}}
     - name: Setup
-      run: cmake -E make_directory ${{runner.workspace}}/build-ci-${{matrix.build_type}}
+      run: |
+        echo "NUM_PROCS=${{matrix.procs}}" >> $GITHUB_ENV
```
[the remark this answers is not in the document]
jrood-nrel: This way shows the numerical value in the output which I prefer.

cmake/set_compile_flags.cmake
jrood-nrel: I will try with that enabled. I really just copied AMReX's warning flags here.

cmake/set_compile_flags.cmake
jrood-nrel: Heh, I was going to look into a generator for this, but I agree, it felt redundant. I will try that.

cmake/set_compile_flags.cmake
jrood-nrel: Ok.

amr-wind/physics/ConvectingTaylorVortex.cpp
jrood-nrel: Yeah this makes more sense looking at it closer.

cmake/set_compile_flags.cmake
jrood-nrel: Actually this didn't seem to work. I found some code where Kitware themselves do this and it looks like this `if(NOT CMAKE_CXX_COMPILER_ID MATCHES "^(Cray|PGI|XL|XLClang)$")` . So I implemented that.

PR #279, opened 2020-12-14
Add linting to CI

jrood-nrel:
Let me know if you think this is worthwhile to add. The check takes less than 5 minutes. If it's worth having, I can fix the warnings it's reporting. I got PeleC down to 0 with not too much work.

I think it's at least worth reporting at a certain severity level if not all, but I kind of like having it report everything. It can also generate an HTML report we could deploy to `gh-pages` but I think the text output is more readable. Currently this check doesn't error and just reports. It's using `macos-latest` because ubuntu has a far too old version of `cppcheck` with `apt-get`.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -50,7 +50,7 @@ jobs:
         working-directory: build-lint
         run: |
           # Sanitize AMREX_GIT_VERSION define with 3 backslashes and quotes which cppcheck can't parse
-          sed -i.bak 's/\\\\\\"//g' compile_commands.json
+          sed -i '' -e 's/\\\\\\"//g' compile_commands.json
           cppcheck --inline-suppr \
             --suppress=unmatchedSuppression --suppress=syntaxError --suppress=internalAstError \
             --std=c++14 --language=c++ --enable=all --project=compile_commands.json \
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,69 @@
+name: AMR-Wind Lint
+
+on:
+  push:
+    branches: [development]
+    paths:
+      - 'cmake/**'
+      - 'amr-wind/**'
+      - 'submods/**'
+      - 'test/**'
+      - 'unit_tests/**'
+      - 'CMakeLists.txt'
+      - '.github/workflows/lint.yml'
+  pull_request:
+    branches: [development]
+    paths:
+      - 'cmake/**'
+      - 'amr-wind/**'
+      - 'submods/**'
+      - 'test/**'
+      - 'unit_tests/**'
+      - 'CMakeLists.txt'
+      - '.github/workflows/lint.yml'
+
+jobs:
+  Lint:
+    runs-on: macos-latest
+    steps:
+      - name: Cancel previous runs
+        uses: styfle/cancel-workflow-action@0.6.0
+        with:
+          access_token: ${{github.token}}
+      - uses: actions/checkout@v2
+        with:
+          submodules: true
+      - name: Dependencies
+        run: brew install cppcheck
+      - name: Configure
+        run: |
+          echo "NPROCS=$(sysctl -n hw.ncpu)" >> $GITHUB_ENV
+          cmake \
+          -Bbuild-lint \
+          -DAMR_WIND_ENABLE_MPI:BOOL=OFF \
+          -DAMR_WIND_ENABLE_TESTS:BOOL=ON \
+          -DAMR_WIND_TEST_WITH_FCOMPARE:BOOL=OFF \
+          -DAMR_WIND_ENABLE_MASA:BOOL=OFF \
+          -DCMAKE_EXPORT_COMPILE_COMMANDS:BOOL=ON \
+          ${{github.workspace}}
+      - name: Check
+        working-directory: build-lint
+        run: |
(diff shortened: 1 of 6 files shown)


.github/workflows/lint.yml
```diff
@@ -0,0 +1,68 @@
...
+          -DAMR_WIND_ENABLE_TESTS:BOOL=ON \
+          -DAMR_WIND_TEST_WITH_FCOMPARE:BOOL=OFF \
+          -DAMR_WIND_ENABLE_MASA:BOOL=OFF \
+          -DCMAKE_EXPORT_COMPILE_COMMANDS:BOOL=ON \
+          ${{github.workspace}}
+      - name: Check
+        working-directory: build-lint
+        run: |
+          # Sanitize AMREX_GIT_VERSION define with 3 backslashes and quotes which cppcheck can't parse
+          sed -i.bak 's/\\\\\\"//g' compile_commands.json
```
sayerhs, line 53: Why do we need a backup file here? Is this to workaround BSD sed's `-i` option?
```suggestion
          sed -i'' 's/\\\\\\"//g' compile_commands.json
```

sayerhs (approved):

.github/workflows/lint.yml
jrood-nrel: I've never been able to do in place on on MacOS.
```
% sed -i'' 's/\\\\\\"//g' compile_commands.json
sed: 1: "compile_commands.json": command c expects \ followed by text
```

sayerhs:
I agree that a lint check is a good idea. Instead of posting to `gh-pages`, my preference would be to look at github actions that would automatically add annotations to PRs so that it is easier to review and address (something like [this](https://github.com/marketplace/actions/annotations-action)). For now instead of failing, may be there can just be a one line annotation with number of warnings.

.github/workflows/lint.yml
sayerhs:
```
❯ uname -s
Darwin
❯ echo '-DAMREX_GIT_VERSION=\\\\\\"20.12-9-g5fb87ea35fa5\\\\\\"' > test.txt
❯ cat test.txt
-DAMREX_GIT_VERSION=\\\"20.12-9-g5fb87ea35fa5\\\"
❯ sed -i '' -e 's/\\\\\\"//g' test.txt
❯ cat test.txt
-DAMREX_GIT_VERSION=20.12-9-g5fb87ea35fa5
```

sayerhs: You forgot `-e` before the replacement string

```
sed -i'' 's/\\\\\\"//g' compile_commands.json
```

jrood-nrel: Do you have a `test.txt-e` file now though?

jrood-nrel: Oh, wait, I see it now. There also needs to be a space between `-i ''`. Cool, this saves me from using `find . -name *.bak | xargs rm` on MacOS in the future.

sayerhs: No, there is no `test.txt-e` file created. That happens because `-i` in BSD sed expects an argument (unlike GNU sed). So if you do 

```
sed -i -e 's/abc/def/g' test.txt
```

`-i` gobbles `-e` and uses it as an extension. So you need an empty place holder `''` to suppress this.

jrood-nrel:
Annotations look perfect for this. I will look into having it report the number of warnings in the PR before merging. Since it can output xml with files and line numbers, it seems possible to annotate files by making that json, which would be quite nice. I will look into that.

Yeah, I will need to enable more of the dependencies in the future for this for more than just the core code.

sayerhs:
Speaking of annotations, I've been meaning to use something like this [run-cmake](https://github.com/lukka/run-cmake) for annotating warnings instead of `-Werror`.

jrood-nrel:
I did some research and I have some things to try. I'm going to close this while I iterate on them in my fork and then reopen this.

jrood-nrel:
@sayerhs I think this might be easier than I thought and maybe you already know, but I found out Github supports problem matchers that can parse GCC warning templates. Since `cppcheck` outputs in a GCC warning template, we can use this to annotate both GCC warnings and `cppcheck` issues. After more research, I think there is a maximum of 10 annotations able to be shown in a PR though which is quite a bummer.

jrood-nrel:
Looks like it works on the files changed tab. I also added an extra semicolon and removed `Werror` to see the warnings work.

sayerhs:
@jrood-nrel Yes, I was just looking at the diffs in the files changed tab. That is looking quite good. After fixing our initial cppcheck warnings, I am not too worried about the limitation of 10 warnings... if there are more warnings than that in a PR, it should most likely fail. So may be the thing to do is check for number of warnings (might need `jq` for that) and then fail the CI action based on that?

sayerhs:
Also looking at the annotations, I see more than 10 annotations.

jrood-nrel:
Ah you're right, I didn't even count. Looks like there is some kind of limit per file, but as you say, once we fix them, 10 is probably workable. Yeah we might need to iterate on it to figure out the right way to fail checks that have lots of warnings.

There is also a Sphinx problem matcher so let me add that.

jrood-nrel:
How would you feel about me merging this? I think I will start working on cleaning up `cppcheck` warnings in a separate PR.

PR #280, opened 2020-12-14
switched to constant nodal projection and updated amrex submodule

michaeljbrazell:


amr-wind/projection/incflo_apply_nodal_projection.cpp
```diff
@@ -104,6 +104,11 @@ void incflo::ApplyProjection(
         }
     }
 
+    bool variable_density =
+        (!m_constant_density || m_sim.physics_manager().contains("MultiPhase"))
+            ? true
+            : false;
+
```
sayerhs, line 111:
```suggestion
    bool variable_density =
        (!m_constant_density || m_sim.physics_manager().contains("MultiPhase"));
           
```

PR #284, opened 2020-12-15
Geostrophic Forcing latitude check

michaeljbrazell:
added check for Geostrophic Forcing to make sure that latitude is set to 90 degrees if Coriolis is also used

### DIFF
--- a/amr-wind/equation_systems/icns/source_terms/GeostrophicForcing.cpp
+++ b/amr-wind/equation_systems/icns/source_terms/GeostrophicForcing.cpp
@@ -39,7 +39,7 @@ GeostrophicForcing::GeostrophicForcing(const CFDSim&)
         pp.query("latitude", latitude);
         AMREX_ALWAYS_ASSERT(
             amrex::Math::abs(latitude - 90.0) <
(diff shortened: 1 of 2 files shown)


amr-wind/equation_systems/icns/source_terms/GeostrophicForcing.cpp
```diff
@@ -31,6 +31,10 @@ GeostrophicForcing::GeostrophicForcing(const CFDSim&)
         amrex::Real rot_time_period = 86400.0;
         pp.query("rotational_time_period", rot_time_period);
         coriolis_factor = 2.0 * utils::two_pi() / rot_time_period;
+
+        amrex::Real latitude = 90.0;
+        pp.query("latitude", latitude);
+        AMREX_ALWAYS_ASSERT(amrex::Math::abs(latitude - 90.0) < 1.0e-14);
```
sayerhs, line 37: While we are at it can we also add an information message that prints out the `f` 

```c++
amrex::Print() << "Geostrophic forcing: Coriolis factor = " << coriolis_factor << std::endl;
```

sayerhs (approved):

PR #286, opened 2020-12-16
Fix cppcheck warnings

sayerhs:


jrood-nrel (approved):
Looks good to me. I found problems in my cppcheck configuration last night though. It's currently not finding any headers because I learned it ignores `-isystem`, so I'm going to submit a PR to fix this (sed the compile commands file further). Surprisingly it reduces the number of issues reported rather than increase it, so I'm a little confused. I'm still working out some kinks it seems, but I think the changes so far have been good regardless.

sayerhs:
@jrood-nrel OK. That explains why it is emitting spurious warnings about certain arguments can be const.

PR #289, opened 2020-12-16
Change order of calculating averages for ABL sims

gantech:


sayerhs:
@gantech what is the magnitude of diffs introduced by this change?

michaeljbrazell:
only abl_godunov_static_refinement caused a diff for me. largest one was velocityx 1.133837433e-07

gantech:
Will this diff be ok?

michaeljbrazell:
I can't track down why this happens with only static refinement...

PR #292, opened 2020-12-16
Fix or suppress last cppcheck issues

jrood-nrel:
Let me rephrase my initial comment. I would prefer to somehow solve these, but I think we might just have to suppress them. This should get to 0 warnings though.

### DIFF
--- a/amr-wind/core/vs/tensorI.H
+++ b/amr-wind/core/vs/tensorI.H
@@ -153,7 +153,8 @@ operator&&(const TensorT<T>& t1, const TensorT<T>& t2)
 template <typename T>
 AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE T mag_sqr(const TensorT<T>& t)
 {
-    return (t);
+    // cppcheck-suppress duplicateExpression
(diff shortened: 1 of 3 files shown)


amr-wind/core/vs/tensorI.H
```diff
@@ -153,7 +153,7 @@ operator&&(const TensorT<T>& t1, const TensorT<T>& t2)
 template <typename T>
 AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE T mag_sqr(const TensorT<T>& t)
 {
-    return (t && t);
+    return (t);
```
sayerhs, line 156: This is not correct.

sayerhs (changes requested):
Making sure this doesn't get accidentally merged

amr-wind/core/vs/tensorI.H
jrood-nrel: Interesting. I just looked this up. I didn't know it was a C++11 thing.

sayerhs (approved):

PR #294, opened 2020-12-18
Add clang-tidy to CMake and CI

jrood-nrel:
@sayerhs what do you think of this? We could try using both `cppcheck` and `clang-tidy` for a bit. We definitely need to iterate on `.clang-tidy`. There are too many checks for me to understand at the moment, so I started with the default for now. I think `-isystem` seems to take care of filtering AMReX includes, but I tried to add filtering in `.clang-tidy` for them just in case.

This is bad form, but I also crammed in the removal of all the Fortran stuff along with this. We don't seem to need it anymore. I remember it breaking linking in CUDA or something, which was why I kept it, but I think that's all been fixed over time. I will fix it if it breaks the CUDA nightly tests. It built for me on Eagle. If you would prefer a separate PR for it, I can do that.


.clang-tidy
```diff
@@ -0,0 +1,47 @@
+---
+#Checks:          '*,-cppcoreguidelines-avoid-magic-numbers,-readability-magic-numbers,-readability-braces-around-statements,-fuchsia-default-arguments,-fuchsia-default-arguments-calls,-fuchsia-overloaded-operator,-fuchsia-statically-constructed-objects,-google-runtime-references,-hicpp-no-assembler,-llvm-include-order,-llvmlibc-callee-namespace,-llvmlibc-implementation-in-namespace,-llvmlibc-restrict-system-libc-headers'
+Checks:          'clang-diagnostic-*,clang-analyzer-*'
```
sayerhs, line 3: Since we are test driving this, I propose we add a few more to this list:

```suggestion
Checks:          'clang-diagnostic-*,clang-analyzer-*,corecppguidelines-*,modernize-*,readability-*'
```

.github/workflows/lint.yml
```diff
@@ -61,15 +61,54 @@ jobs:
...
+        run: |
+          echo "::add-matcher::.github/problem-matchers/gcc.json"
+          # Crude warning number approximation
+          WARNINGS_TMP=$(wc -l < clang-tidy-warnings.txt | xargs echo -n)
           WARNINGS=$(bc <<< "$WARNINGS_TMP/3")
-          printf "%s warnings\n" "${WARNINGS}" >> amr-wind-static-analysis.txt
-          cat amr-wind-static-analysis.txt
-          # Just report for now and don't fail
-          #exit ${WARNINGS}
+          printf "Approximately %s warnings\n" "${WARNINGS}" >> clang-tidy-warnings.txt
```
sayerhs, line 113: Looking at the outputs in the Checks tab, it looks like `clang-tidy` outputs a lot more than 3 lines per warning. It also shows the code path it took to arrive at that warning.

.clang-tidy
```diff
@@ -0,0 +1,47 @@
...
+    value:           reasonable
+  - key:             modernize-loop-convert.NamingStyle
+    value:           CamelCase
+  - key:             modernize-pass-by-value.IncludeStyle
+    value:           llvm
+  - key:             modernize-replace-auto-ptr.IncludeStyle
+    value:           llvm
+  - key:             modernize-use-nullptr.NullMacros
+    value:           'NULL'
+...
```
sayerhs, line 47:
```suggestion
  - { key: readability-identifier-naming.NamespaceCase,       value: lower_case }
  - { key: readability-identifier-naming.ClassCase,           value: CamelCase  }
  - { key: readability-identifier-naming.PrivateMemberPrefix, value: m_         }
  - { key: readability-identifier-naming.ProtectedMemberPrefix, value: m_         }
  - { key: readability-identifier-naming.StructCase,          value: CamelCase  }
  - { key: readability-identifier-naming.FunctionCase,        value: lower_case }
  - { key: readability-identifier-naming.VariableCase,        value: lower_case }
  - { key: readability-identifier-naming.GlobalConstantCase,  value: UPPER_CASE }
...
```

CMakeLists.txt
```diff
@@ -128,7 +123,9 @@ endif()
 
 if (AMR_WIND_ENABLE_UNIT_TESTS OR AMR_WIND_ENABLE_TESTS)
   add_executable(${amr_wind_unit_test_exe_name})
+  disable_clang_tidy()
   add_subdirectory("submods/googletest")
+  enable_clang_tidy()
```
sayerhs, line 128: Instead of the toggle, is it possible to use `set_target_properties` to only add the `CXX_CLANG_TIDY` property to AMR-Wind targets?

```cmake
set_target_properties(${amr_wind_lib_name} PROPERTIES CXX_CLANG_TIDY ${CLANG_TIDY_EXE})
```

amr-wind/CMakeLists.txt
```diff
@@ -37,8 +37,11 @@ if(AMR_WIND_ENABLE_MASA)
 endif()
 
 # Generate AMReX build information file
+include(amr-wind-utils)
```
sayerhs, line 40: I thought that the `include` from top-level would be inherited here... is that not the case?

sayerhs (approved):

.github/workflows/lint.yml
jrood-nrel: Yeah, I could try to parse for `warning:`. I mostly just wanted something to show as 0 warnings in the report once we get there, but I can try to make something actually count the warnings.

CMakeLists.txt
jrood-nrel: Oh that's a good idea. I was trying to think of how to do this originally, but I kept thinking of it as having to turn off `${CLANG_TIDY_EXE}` for all other targets.

amr-wind/CMakeLists.txt
jrood-nrel: No, I tried, but it didn't seem to work. I will double check again though. I expected it to.

jrood-nrel: Something must have been in flux that made this not work initially because it does now that I tried it again.

PR #296, opened 2020-12-21
Reduce atomic pressure in plane averages with sequential loop

maxpkatz:
For a 256x256x64 (single-box) case on an NVIDIA V100, the third moment averaging of velocity takes ~70 ms which is quite expensive for this domain size. The expense is mainly due to many threads performing atomic operations on the same output locations. We can bring down the cost by having fewer threads simultaneously writing to a given output location. A way to achieve this is, instead of doing the whole 3D box in parallel at once, to break down the box into 2D planes and then loop sequentially over the 2D planes. For example, when the axis is the z-axis, loop over the x-y planes from the bottom of the box to the top. This is paired with a ParallelFor that launches only as many threads as needed to span the x-y plane. This brings down the cost to ~20 ms for the case mentioned above.

### DIFF
--- a/amr-wind/utilities/DirectionSelector.H
+++ b/amr-wind/utilities/DirectionSelector.H
@@ -43,8 +43,8 @@ using ZDir = DirectionSelector<2>;
 // For example, if we're using ZDir, return a box covering the x-y plane.
 // The IntVect is used to set the constant index in the parallel direction.
 template <typename IndexSelector>
-AMREX_GPU_HOST_DEVICE
-amrex::Box PerpendicularBox(const amrex::Box& bx, const amrex::IntVect& iv)
+AMREX_GPU_HOST_DEVICE amrex::Box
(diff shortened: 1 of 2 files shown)


sayerhs (approved):

sayerhs:
@gantech @michaeljbrazell It seems that while optimizing the moment averaging code for performance, @maxpkatz might have stumbled upon a bug in our third moment averaging code. Can one of you quickly look at the code and also confirm? If you do that @maxpkatz can add the bugfix to this PR before we merge it. 

The bug in question is in line 191-192, I think the last array index `mfab_arr3` should be `p` instead of `n`

https://github.com/Exawind/amr-wind/blob/04453b8461c0906979485e6a94245fbc6a964a15/amr-wind/utilities/ThirdMomentAveraging.cpp#L182-L200

michaeljbrazell:
yep, that looks like a bug, good catch!

maxpkatz:
Updated with that bug fix.

PR #297, opened 2020-12-21
Sequential loop over axis index for all plane averages

maxpkatz:
Extends the approach from #296 to the remainder of the plane averages. A typical speedup for the updated kernels is 2.5x for a large box on V100.


sayerhs (approved):

michaeljbrazell:
I think there is a typo in the PerpendicularBox line 55, it should be: plane_lo = {bx.smallEnd(0), iv[1], bx.smallEnd(2)};

that should fix the unit tests

maxpkatz:
Nice catch, @michaeljbrazell. That appears to have been the issue.

PR #300, opened 2020-12-23
Move cppcheck to cmake target

jrood-nrel:
I would like to make it easier for users to run these new checks locally. Once we settle on a method, I will write some documentation. I will also start addressing `clang-tidy` warnings with some PRs and automatic fixes.

- Move cppcheck to a cmake target
- Update CI accordingly
- Add cppcheck-specific problem matcher
- Reduce a few clang-tidy checks while I address issues
- Generate `compile_commands.json` by default

### DIFF
--- a/cmake/amr-wind-utils.cmake
+++ b/cmake/amr-wind-utils.cmake
@@ -84,8 +84,8 @@ macro(init_code_checks)
           COMMAND ${CMAKE_COMMAND} -E echo "Running cppcheck on project using ${NP} cores..."
           COMMAND ${CMAKE_COMMAND} -E make_directory cppcheck/cppcheck-wd
           # cppcheck ignores -isystem directories, so we change them to regular -I include directories (with no spaces either)
-          COMMAND sed "s/isystem /I/g" compile_commands.json > cppcheck_compile_commands.json
-          COMMAND ${CPPCHECK_EXE} --template=gcc --inline-suppr --suppress=internalAstError --suppress=unusedFunction --std=c++14 --language=c++ --enable=all --project=cppcheck_compile_commands.json --cppcheck-build-dir=cppcheck/cppcheck-wd -i ${CMAKE_SOURCE_DIR}/submods/amrex/Src -i ${CMAKE_SOURCE_DIR}/submods/googletest --output-file=cppcheck/cppcheck-full-report.txt -j ${NP}
+          COMMAND sed "s/isystem /I/g" compile_commands.json > cppcheck/cppcheck_compile_commands.json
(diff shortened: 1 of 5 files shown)


jrood-nrel:
For some reason it seems my stripped down reports don't get picked up correctly by the problem-matcher. I need to fix that.

cmake/amr-wind-utils.cmake
```diff
@@ -70,4 +70,34 @@ macro(init_clang_tidy)
...
+          COMMAND ${CPPCHECK_EXE} --template=gcc --inline-suppr --suppress=internalAstError --suppress=unusedFunction --std=c++14 --language=c++ --enable=all --project=cppcheck/cppcheck_compile_commands.json --cppcheck-build-dir=cppcheck/cppcheck-wd -i ${CMAKE_SOURCE_DIR}/submods/amrex/Src -i ${CMAKE_SOURCE_DIR}/submods/googletest --output-file=cppcheck/cppcheck-full-report.txt -j ${NP}
+          # Filter out submodule source files after analysis
+          COMMAND awk -v nlines=2 "/submods/ {for (i=0; i<nlines; i++) {getline}; next} 1" < cppcheck/cppcheck-full-report.txt > cppcheck/cppcheck-short-report.txt
+          COMMAND cat cppcheck/cppcheck-short-report.txt | egrep "information:|error:|performance:|portability:|style:|warning:" | sort > cppcheck-warnings.txt
+          COMMAND printf "Warnings: " >> cppcheck-warnings.txt
+          COMMAND cat cppcheck-warnings.txt | awk "END{print NR-1}" >> cppcheck-warnings.txt
+          COMMENT "Run cppcheck on project compile_commands.json"
+          BYPRODUCTS cppcheck-warnings.txt
+          WORKING_DIRECTORY ${CMAKE_BINARY_DIR}
+          VERBATIM USES_TERMINAL
```
sayerhs, line 97: This target mixes what I would expect to be a simple `make cppcheck` target that I would want to use during local development with all the processing and reporting concerns for CI workflows. This code should be broken down into modular pieces that separates these tasks. I would recommend that `make cppcheck` just runs cppcheck and then may be another dependent target `cppcheck-ci` that generates a CI report.

jrood-nrel: Makes sense.

jrood-nrel:
I fixed the problem matchers too. I was doing a `uniq -c` and messing up the warning line format.

sayerhs (approved):

jrood-nrel:
Ok, I think I'm going to merge this. One thing bugging me is that `cppcheck` on MacOS seems to find much less issues than on Linux. Looking at the nightly tests, these are all missing on MacOS https://my.cdash.org/build/1940306/notes . I have this same experience in PeleC. I will switch the `cppcheck` VM here to ubuntu once they update `ubuntu-latest` so I can get a newer `cppcheck` than currently available with `apt-get`.

sayerhs:
@jrood-nrel We should update `ubuntu-latest` to explicitly be `ubuntu-20.04`

jrood-nrel:
Yeah. I remember seeing notes somewhere about them updating `ubuntu-latest` to that very soon, but maybe we just do it expilcitly. I will look into it next week.

PR #301, opened 2020-12-24
Fix segfaults with fcompare on nightly tests

sayerhs:
https://my.cdash.org/test/19431410

### DIFF
--- a/test/CTestList.cmake
+++ b/test/CTestList.cmake
@@ -37,19 +37,21 @@ function(add_test_r TEST_NAME)
     if(AMR_WIND_ENABLE_CUDA)
       set(FCOMPARE_TOLERANCE "-r 1e-10 --abs_tol 1.0e-12")
     endif()
-    # Use fcompare to test diffs in plots against gold files
-    if(AMR_WIND_TEST_WITH_FCOMPARE)
-      set(FCOMPARE_COMMAND "&& ${FCOMPARE_EXE} ${FCOMPARE_TOLERANCE} ${PLOT_GOLD} ${PLOT_TEST}")
-    endif()
     if(AMR_WIND_ENABLE_MPI)
       set(NP 4)
       set(MPI_COMMANDS "${MPIEXEC_EXECUTABLE} ${MPIEXEC_NUMPROC_FLAG} ${NP} ${MPIEXEC_PREFLAGS}")
+      set(FCOMPARE_COMMAND "${MPIEXEC_EXECUTABLE} ${MPIEXEC_NUMPROC_FLAG} 1 ${MPIEXEC_PREFLAGS} ${FCOMPARE_EXE} ${FCOMPARE_TOLERANCE} ${PLOT_GOLD} ${PLOT_TEST}")
     else()
       set(NP 1)
       unset(MPI_COMMANDS)
+      set(FCOMPARE_COMMAND "${FCOMPARE_EXE} ${FCOMPARE_TOLERANCE} ${PLOT_GOLD} ${PLOT_TEST}")
     endif()
-    # Add test and actual test commands to CTest database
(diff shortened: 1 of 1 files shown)


test/CTestList.cmake
```diff
@@ -37,19 +37,21 @@ function(add_test_r TEST_NAME)
...
       set(MPI_COMMANDS "${MPIEXEC_EXECUTABLE} ${MPIEXEC_NUMPROC_FLAG} ${NP} ${MPIEXEC_PREFLAGS}")
+      set(FCOMPARE_COMMAND "${MPIEXEC_EXECUTABLE} ${MPIEXEC_NUMPROC_FLAG} 1 ${MPIEXEC_PREFLAGS} ${FCOMPARE_EXE} ${FCOMPARE_TOLERANCE} ${PLOT_GOLD} ${PLOT_TEST}")
     else()
       set(NP 1)
       unset(MPI_COMMANDS)
+      set(FCOMPARE_COMMAND "${FCOMPARE_EXE} ${FCOMPARE_TOLERANCE} ${PLOT_GOLD} ${PLOT_TEST}")
+    endif()
+    if (AMR_WIND_TEST_WITH_FCOMPARE)
+      # Add test and actual test commands to CTest database
+      add_test(${TEST_NAME} sh -c "${MPI_COMMANDS} ${CMAKE_BINARY_DIR}/${amr_wind_exe_name} ${MPIEXEC_POSTFLAGS} ${CURRENT_TEST_BINARY_DIR}/${TEST_NAME}.i ${RUNTIME_OPTIONS} > ${TEST_NAME}.log && ${FCOMPARE_COMMAND}")
```
jrood-nrel, line 51: So I gather `fcompare` needs an `mpirun`, but I don't think I like the duplication of this line. Seems like it should go in the lines above. Is it done to not obscure the `&&`?

test/CTestList.cmake
```diff
@@ -37,19 +37,21 @@ function(add_test_r TEST_NAME)
...
       set(FCOMPARE_TOLERANCE "-r 1e-10 --abs_tol 1.0e-12")
     endif()
-    # Use fcompare to test diffs in plots against gold files
-    if(AMR_WIND_TEST_WITH_FCOMPARE)
-      set(FCOMPARE_COMMAND "&& ${FCOMPARE_EXE} ${FCOMPARE_TOLERANCE} ${PLOT_GOLD} ${PLOT_TEST}")
-    endif()
     if(AMR_WIND_ENABLE_MPI)
       set(NP 4)
       set(MPI_COMMANDS "${MPIEXEC_EXECUTABLE} ${MPIEXEC_NUMPROC_FLAG} ${NP} ${MPIEXEC_PREFLAGS}")
+      set(FCOMPARE_COMMAND "${MPIEXEC_EXECUTABLE} ${MPIEXEC_NUMPROC_FLAG} 1 ${MPIEXEC_PREFLAGS} ${FCOMPARE_EXE} ${FCOMPARE_TOLERANCE} ${PLOT_GOLD} ${PLOT_TEST}")
```
jrood-nrel, line 43: I don't understand `NP=1` here. Why use `mpirun`?

jrood-nrel: Actually, maybe the question here is why does it segfault without an `mpirun`? I thought if `np=1`, then `mpirun` shouldn't be necessary.

test/CTestList.cmake
sayerhs: Yeah, I don't like the duplication. This function is already kind of a mess. I didn't want to do nested if conditions, so this was the best way to introduce fewest lines of change. If you see a better way, go for it.

test/CTestList.cmake
sayerhs: You can take this up with NREL Eagle. :)

jrood-nrel: Actually, I'm thinking we just let `fcompare` run with `NP` and use the already existing `MPI_COMMANDS`?

sayerhs: That's fine with me too.

jrood-nrel:
What you think about #302 I just made? I haven't tested it, but just what I'm thinking.

sayerhs:
Closing in favor of #302

PR #317, opened 2021-01-15
Update GoogleTest submodule and suppress remaining unread value warnings

jrood-nrel:
I think the unread value warnings are false positives. Updating GoogleTest avoids the CMake deprecation warning about CMake compatibility < 2.18.2 being removed in future releases.

### DIFF
--- a/amr-wind/convection/incflo_godunov_advection.cpp
+++ b/amr-wind/convection/incflo_godunov_advection.cpp
@@ -74,7 +74,6 @@ void godunov::compute_advection(
     Array4<Real> xyzlo = makeArray4(p, bxg1, ncomp);
     p += xyzlo.size();
     Array4<Real> xyzhi = makeArray4(p, bxg1, ncomp);
-    p += xyzhi.size(); // NOLINT: Value not read warning
 
     // Use PPM to generate Im and Ip */
     switch (godunov_scheme) {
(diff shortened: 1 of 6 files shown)


jrood-nrel:
There are 3 warnings left about `warning: Call to virtual function during construction` in which I don't know the correct solution.

amr-wind/convection/incflo_godunov_advection.cpp
```diff
@@ -74,7 +74,7 @@ void godunov::compute_advection(
     Array4<Real> xyzlo = makeArray4(p, bxg1, ncomp);
     p += xyzlo.size();
     Array4<Real> xyzhi = makeArray4(p, bxg1, ncomp);
-    p += xyzhi.size();
+    p += xyzhi.size(); // NOLINT: Value not read warning
```
sayerhs, line 77: I was thinking we should remove this line.

sayerhs (approved):

sayerhs:
@jrood-nrel regarding the call to virtual, I think we can clean it up... it is not much code to fix that.

amr-wind/convection/incflo_godunov_advection.cpp
jrood-nrel: I wasn't certain if `p` was required elsewhere, but I can remove it.

PR #318, opened 2021-01-17
Eulerian Implicit scheme for VOF + new multiphase physics sub-classes

gdeskos:
- Implemented and tested the split-advection Eulerian scheme of Weymouth & Yue (2010)
- Added the raindrop and sloshing tank examples
- For Zalesak disk and Vortex Patch density is initialized within the sub-classes' "initialize_fields" so AMR operates at t=0
- Density is computed at n+1/2 and n+1 within PDE's PostOp for consistency. Changes may affect TKE PostOp 
- Added the surface tension forcing term to nodal projection (surface tension forcing is set to zero at the moment)


amr-wind/equation_systems/vof/SplitAdvection.cpp
```diff
@@ -40,126 +42,273 @@ void multiphase::split_advection(
...
-                multiphase::lagrangian_advection(
-                    i, j, k, 1, dtdx, uL, uR, volfrac, fluxL, fluxC, fluxR);
-            });
-        amrex::ParallelFor(
-            bx, [=] AMREX_GPU_DEVICE(int i, int j, int k) noexcept {
-                multiphase::balance_fluxes(
-                    i, j, k, 1, volfrac, fluxL, fluxC, fluxR, pbc, domlo.x,
-                    domhi.x);
-            });
+    if (is_lagrangian) {
```
sayerhs, line 45: Please refactor these nested if conditions to reduce code duplication.

gdeskos: See latest commit

sayerhs (approved):

michaeljbrazell (approved):

PR #320, opened 2021-01-18
Resolve pass by const reference warnings

jrood-nrel:
Also remove some unread variables.


sayerhs:
Was this tested with netcdf enabled?

jrood-nrel:
Yes this is from the checks in the nightly tests and NetCDF is enabled. I see `-DAMR_WIND_USE_NETCDF` in the `compile_commands.json` that cppcheck is using as well.

sayerhs (approved):

PR #340, opened 2021-02-11
Allow AMR-Wind to change max_grid_size on restarts

sayerhs:
Modify checkpoint file read behavior to allow the user to change
``max_grid_size`` on restart. The use-case is to restart from a checkpoint file
that was run on CPUs (``max_grid_size = 32``) on GPUs where we want 128.

### DIFF
--- a/amr-wind/utilities/IOManager.H
+++ b/amr-wind/utilities/IOManager.H
@@ -6,6 +6,8 @@
 #include <set>
 
 #include "AMReX_Vector.H"
+#include "AMReX_BoxArray.H"
+#include "AMReX_DistributionMapping.H"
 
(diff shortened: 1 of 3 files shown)


jrood-nrel (approved):
Pretty good functionality to have. Should we have a restart test that uses this?

sayerhs:
Yeah, we might be able to adapt `abl_godunov_restart` to test this also. I need to add one more capability before this can be merged to handle `ChopGrids`

sayerhs:
@WeiqunZhang can you please review this? Particularly my usage of `ChopGrids`.

michaeljbrazell (approved):

PR #359, opened 2021-02-21
Adjust MAC/Nodal projection settings for channel flow in GPU builds

sayerhs:
Relax the absolute tolerance for elliptic solves so that MLMG does not fail on GPU runs. Also enable semicoarsening for MAC solves.

Requires #358


sayerhs:
[CDash link](https://my.cdash.org/viewTest.php?onlyfailed&buildid=1970513) to failing nighty tests.

michaeljbrazell (approved):

sayerhs:
The changes in this PR are not robust. We use this as a temporary placeholder until we can find robust settings for RANS simulations (probably using hypre as bottom solver).

PR #365, opened 2021-02-27
Add ability to start Re averaging at user-specified time

sayerhs:


tonyinme:
@sayerhs, this looks great!

Is the value of the 'averaging_end_time' the maximum allowed float value in the code?
Will this be ok in terms of precision when computing the time averaging?

sayerhs:
@tonyinme The default value for `averaging_end_time` would be the maximum allowable `amrex::Real` for a given system (`amrex::Real` is almost always set to double precision in AMR-Wind builds). Currently, there are some build errors if `amrex::Real` is set to `float` (see PR #264) for the changes required. See [C++ reference page](https://en.cppreference.com/w/cpp/types/numeric_limits) for more details on the default value. 

I am not sure I understand what you mean by _will this be ok in terms of precision when computing time averaging_. As implemented, this value does not enter into the calculations. It is only used to check if the averaging functions be called or not.

tonyinme:
ok, perfect.
I thought that averaging_end_time was entering the calculation of `filter` and `factor` but I can see that it is not. 
This looks good!

tonyinme (approved):

PR #368, opened 2021-02-28
Fix LinearProfile UDF to work on both scalars and vectors

sayerhs:


gantech (approved):
This fixes the bug for LinearProfile on scalars

PR #376, opened 2021-03-19
multilevel flux bug

michaeljbrazell:
- added bugfix from incflo that averages fluxes before using them which will maintain conservation
- this is only done for godunov
- @jrood-nrel we will need to rebless the multilevel godunov regression tests: abl_godunov_static_refinement and rayleigh_taylor_godunov

### DIFF
--- a/amr-wind/convection/incflo_godunov_advection.cpp
+++ b/amr-wind/convection/incflo_godunov_advection.cpp
@@ -7,11 +7,13 @@
 
 using namespace amrex;
 
-void godunov::compute_advection(
+void godunov::compute_fluxes(
     int lev,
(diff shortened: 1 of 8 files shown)


amr-wind/convection/incflo_godunov_advection.cpp
```diff
@@ -565,26 +576,48 @@ void godunov::compute_advection(
                        ? 0.5 * (stl + sth)
                        : temp;
             qz(i, j, k, n) = temp;
```
michaeljbrazell, line 578: should I remove the qx, qy, qz storage?

sayerhs (approved):

PR #389, opened 2021-04-15
remove extrapolation option

michaeljbrazell:
- removed extrapolation option, we never used it anyways and it's risky due to a possible negative viscosity
- removed density from gravity term
- should we remove the old wall model implementation?


sayerhs (approved):

michaeljbrazell:
@gdeskos @shashankNREL what do you think? should I delete the old implementation of the wall model BCs or keep it around?

amr-wind/diffusion/incflo_diffusion.cpp
```diff
@@ -133,16 +130,6 @@ void wall_model_bc(
...
-    // copies cell center to face
-    Real c0 = 1.0;
-    Real c1 = 0.0;
-
-    // linear extrapolate onto face
-    if (extrapolate) {
-        c0 = 1.5;
-        c1 = -0.5;
-    }
-
```
gdeskos, line 145: Don't we want to keep the "copy cell center to face" case ? This doesn't run the risk of negative viscosity. But I am happy to remove it all together if you don't think we are going to need it in the future

shashankNREL:
> @gdeskos @shashankNREL what do you think? should I delete the old implementation of the wall model BCs or keep it around?

Please remove the old implementation of the wall model BCs.

amr-wind/diffusion/incflo_diffusion.cpp
michaeljbrazell: that's what I did, the PR only does the copy from cell center to face. Unless I made a typo...

PR #393, opened 2021-04-19
Failing CDash tests: remove y-component output for GPU tests

sayerhs:
Adjust `sloshing_tank` and `dam_break_godunov` to skip the `y-component` of
velocity and `grad(p)` terms in nightly regression tests.


gdeskos (commented):
Looks good at the moment. I should get to the bottom of what's wrong with the periodic Bus here!

gdeskos (approved):

PR #399, opened 2021-04-24
Fix small bugs in Synthetic Turbulence

gantech:
1. Fix bug in conversion of wind direction to transformation matrix for local coordinate system
2. Fix bug in interpolation of wind perturbation from netcdf file

### DIFF
--- a/amr-wind/physics/SyntheticTurbulence.cpp
+++ b/amr-wind/physics/SyntheticTurbulence.cpp
@@ -279,7 +279,6 @@ AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE bool find_point_in_box(
     // Get y and z w.r.t. the lower corner of the grid
     const amrex::Real yy = pt[1] + t_grid.box_len[1] * 0.5;
     const amrex::Real zz = pt[2] + t_grid.box_len[2] * 0.5;
-
     bool inBox =
         ((yy >= 0.0) && (yy <= t_grid.box_len[1]) && (zz >= 0.0) &&
          (zz <= t_grid.box_len[2]));
@@ -302,7 +301,7 @@ AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE void interp_perturb_vel(
     // clang-format off
     // Indices of the 2-D cell that contains the sampling point
     int qidx[4]{wt.jl * nz + wt.kl, wt.jr * nz + wt.kl, wt.jr * nz + wt.kr,
-                wt.jl * nz + wt.kl};
+                wt.jl * nz + wt.kr};
     // clang-format on
 
     vs::Vector vel_l, vel_r;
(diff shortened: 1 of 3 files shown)


amr-wind/physics/SyntheticTurbulence.cpp
```diff
@@ -449,15 +448,16 @@ SyntheticTurbulence::SyntheticTurbulence(const CFDSim& sim)
     //
     // x-direction points to flow direction (convert from compass direction to
     // vector)
-    m_turb_grid.tr_mat.xx() = -std::sin(wind_direction);
-    m_turb_grid.tr_mat.xy() = -std::cos(wind_direction);
+    m_turb_grid.tr_mat.xx() = std::cos(theta_t);
```
sayerhs, line 451: What's the bug here?

gantech: For some reason, the second row of `tr_mat` came out to zero. This seemed to fix that problem.

sayerhs: Seems like there bug is in there cross product and not here.

amr-wind/physics/SyntheticTurbulence.cpp
```diff
@@ -449,15 +448,16 @@ SyntheticTurbulence::SyntheticTurbulence(const CFDSim& sim)
...
     m_turb_grid.tr_mat.xz() = 0.0;
+    m_turb_grid.tr_mat.yx() = -std::sin(theta_t);
+    m_turb_grid.tr_mat.yy() = std::cos(theta_t);
+    m_turb_grid.tr_mat.yz() = 0.0;
     // z always points upwards (for now...)
     m_turb_grid.tr_mat.zx() = 0.0;
     m_turb_grid.tr_mat.zy() = 0.0;
     m_turb_grid.tr_mat.zz() = 1.0;
-    // y = z .cross. x
-    m_turb_grid.tr_mat.y() = m_turb_grid.tr_mat.z() ^ m_turb_grid.tr_mat.x();
```
sayerhs, line 460: @gantech See #400 for a fix to catch this error in future.

amr-wind/physics/SyntheticTurbulence.cpp
sayerhs: Also you can simplify this section by writing 

```C++
const amrex::Real wdir = 270.0 - wind_direction; 
m_turb_grid.tr_mat = Tensor::zrot(wdir);
```

amr-wind/physics/SyntheticTurbulence.cpp
gantech: Thanks Shreyas.

amr-wind/physics/SyntheticTurbulence.cpp
gantech: Thanks Shreyas. Fixed in https://github.com/Exawind/amr-wind/pull/399/commits/57de3963a2bf9632042441ceed12f74cde912cf5

sayerhs (approved):

PR #403, opened 2021-04-26
Fix ABL IO boundary planes on GPU

### LINKED ISSUE #396: Boundary plane input on GPUs
I am trying to run the abl_bndry_input regression test case on GPUs (Eagle and Summit). I first ran the abl_bndry_output test case to generate all the relevant files to run the abl_bndry_input regression test case. The case works without any problem on CPUs on Eagle. The same case with GPU build on eagle (and Summit) gives the wrong initial residual for Mac projection and crashes in the MAC_projection step. 
This issue can be reproduced using the abl_bndry_input regression test on GPU builds. 
@marchdf : Can you verify that this case worked on GPUs in the past. Is that something that you could look into? 
@jrood-nrel : Can we set up a suite of regression case testing using GPUs on Eagle.

marchdf:
Just needed to remove `AMREX_GPU_DEVICE` in a `LoopOnCpu`.

Closes #396

### DIFF
--- a/amr-wind/wind_energy/ABLBoundaryPlane.cpp
+++ b/amr-wind/wind_energy/ABLBoundaryPlane.cpp
@@ -48,7 +48,7 @@ AMREX_FORCE_INLINE amrex::IntVect offset(const int face_dir, const int normal)
     return offset;
 }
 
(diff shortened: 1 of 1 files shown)


michaeljbrazell (approved):
looks good, confirmed it runs on eagle with GPUs

PR #407, opened 2021-05-03
Cleaning up warnings May 2021 round 1

jrood-nrel:
I'm working on cleaning up warnings again. This first round is mostly to add `explicit` to constructors with one argument, and adding some `const`, but there are a few other fixes as well.


michaeljbrazell (approved):
looks good!

PR #409, opened 2021-05-05
Averaging documentation

tonyinme:
Here is some documentation for the Averaging options within AMR-Wind.


michaeljbrazell:
@tonyinme can you add a sentence that you need to have the label in the incflo.post_processing = averaging. and/or add it in the post_processing example section. Otherwise looks good. Are you going to add documentation for actuator lines in a separate PR?

tonyinme:
hey @michaeljbrazell, just added it!
Yes, the plan is to have another PR for the ALM documentation.

PR #410, opened 2021-05-07
Address more warnings round 3

jrood-nrel:


michaeljbrazell (approved):
compiles and passed reg tests on my end (mac clang)

PR #412, opened 2021-05-13
Fix warnings round 4

jrood-nrel:
I really just fix these while I'm waiting for other things I'm working on.


jrood-nrel:
HIP is expected to fail for amrex.

PR #413, opened 2021-05-20
Actuator disk first pass

psakievich:
First pass at the actuator disk implementation.  

Still working through the concepts a bit.


sayerhs:
Why are OpenFAST files tracked within the main source code? This a really bad idea. I also think replicating the nalu set-up for OpenFAST files is not the way to go. We should be taking OpenFAST files directly from the OpenFAST r-test repo.

psakievich:
That is for convenience while I work through the tests.  I already had plans to make a common infrastructure for both codes wrt OpenFAST files but I don't have time to implement it right now.

michaeljbrazell:
@psakievich can you add a regression test to this PR?

michaeljbrazell:
> @psakievich can you add a regression test to this PR?

Thanks for the quick chat, let's hold off on adding the regression test. When Jon gets back we can discuss how to incorporate openfast.

michaeljbrazell (approved):

PR #418, opened 2021-06-02
Add component view for hub

psakievich:
This fixes a memory leak because I forgot to deal with all the hub stuff and added it hastily at the end of #413.

I've added a component view for this rather than just using index 0.  This will allow for turbine data structures beyond openfast in the future and keeps the framework the same for hub when compared with the blades and tower.


michaeljbrazell (approved):

tonyinme (approved):
Looks good.

PR #422, opened 2021-06-09
Add a Constant Ct Disk

psakievich:
Implement a constant Ct actuator disk.  Adding as a draft to make development transparent. Will update to PR when all the pieces are in place.


amr-wind/wind_energy/actuator/disk/constant_ct_ops.H
```diff
@@ -0,0 +1,116 @@
...
+
+        const auto& cc = ddata.center;
+        const amrex::Real xLen = (unit.x()&n)*nl + (unit.x()&planeVec)*dl;
+        const amrex::Real yLen = (unit.y()&n)*nl + (unit.y()&planeVec)*dl;
+        const amrex::Real zLen = (unit.z()&n)*nl + (unit.z()&planeVec)*dl;
+
+        info.bound_box = amrex::RealBox(
+          cc.x()-xLen, cc.y()-yLen, cc.z()-zLen, 
+          cc.x()+xLen, cc.y()+yLen, cc.z()+zLen
+        );
```
sayerhs, line 56: This doesn't look correct to me. Let us say

```
n = (1 0 0)
unit.z() = (0 0 1)
planeVec = n ^ z
``` 

Then zLen works out to be zero right?

amr-wind/wind_energy/actuator/disk/constant_ct_ops.H
```diff
@@ -0,0 +1,116 @@
...
+        const vs::Vector& n = ddata.normal;
+        // get a vector in the plane of the disk
+        vs::Vector planeVec = vs::Vector::one();
+        const auto unit = vs::Tensor::I();
+
+        AMREX_ALWAYS_ASSERT_WITH_MESSAGE(
+            std::abs((unit.z() & n) - vs::mag_sqr(n)) >
+                vs::DTraits<amrex::Real>::eps(),
+            "A constant Ct disk normal is too close to vertical. Vertical axis "
+            "wind turbines are not supported with this model.");
```
sayerhs, line 39: I don't think the actuator disk would be normal to wind direction in VAWTs.

amr-wind/wind_energy/actuator/disk/ActSrcDiskOp.H
```diff
@@ -0,0 +1,133 @@
...
+    amrex::Gpu::copy(
+        amrex::Gpu::hostToDevice, grid.force.begin(), grid.force.end(),
+        m_force.begin());
+}
+
+template <typename ActTrait>
+void ActSrcOp<
+    ActTrait,
+    ActSrcDisk,
+    typename std::enable_if<std::is_base_of<DiskType, ActTrait>::value>::type>::
```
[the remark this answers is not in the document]
psakievich: @sayerhs yeah I'm still working through this section of the design. When I put in the OpenFAST version of the actuator disk I realized that there are two main variants of the disk we are supporting.  Those that are derivative of a turbine type (have tower and hub) and those that are just a flat disk.  From my understanding having a tower and nacelle aren't a high priority for the TCF project.  Right now this is design to break the ops along those lines. I'm open to another suggestion for how to do this if you have one.

amr-wind/wind_energy/actuator/disk/constant_ct_ops.H
```diff
@@ -0,0 +1,116 @@
...
+        pp.get("disk_normal", ddata.normal);
+        pp.get("epsilon", ddata.epsilon);
+        pp.get("diameter", ddata.diameter);
+        // TODO think about center of disk
+        const amrex::Real radius = ddata.diameter * 0.5;
+        ddata.dr = radius / ddata.num_points;
+
+        // ensure normal is normalized
+        ddata.normal = ddata.normal.normalize();
+        const vs::Vector& n = ddata.normal;
```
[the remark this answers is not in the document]
psakievich: ah thanks! this is good to know

amr-wind/wind_energy/actuator/disk/ActSrcDiskOp.H
```diff
@@ -0,0 +1,133 @@
...
+    initialize()
+{
+    const auto& grid = m_data.grid();
+    m_pos.resize(grid.pos.size());
+    m_force.resize(grid.force.size());
+}
```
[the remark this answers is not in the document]
psakievich: Not really.  I wrote this first and just copied the logic from `ActSrcLineOp`.  I'm working through the Ops one by one and will probably rework this section once I finish stitching the aero model and other things together.

amr-wind/wind_energy/actuator/disk/constant_ct_ops.H
```diff
@@ -0,0 +1,116 @@
...
+        pp.get("disk_center", ddata.center);
+        pp.get("disk_normal", ddata.normal);
+        pp.get("epsilon", ddata.epsilon);
+        pp.get("diameter", ddata.diameter);
+        // TODO think about center of disk
+        const amrex::Real radius = ddata.diameter * 0.5;
+        ddata.dr = radius / ddata.num_points;
+
+        // ensure normal is normalized
+        ddata.normal = ddata.normal.normalize();
```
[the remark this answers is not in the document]
psakievich: :+1:

amr-wind/wind_energy/actuator/disk/ActSrcDiskOp.H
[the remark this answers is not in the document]
sayerhs: OK. That makes sense.

amr-wind/wind_energy/actuator/disk/constant_ct_ops.H
sayerhs: Try this 

```C++
const auto dvec = n * nl + planeVec * dl + vs::Vector::khat() * dl;
const auto p1 = cc - dvec;  // front
const auto p2 = cc + dvec;  // back
info.bound_box = amrex::RealBox(
    amrex::min(p1.x(), p2.x()),
    amrex::min(p1.y(), p2.y()),
    amrex::min(p1.z(), p2.z()),
    amrex::max(p1.x(), p2.x()),
    amrex::max(p1.y(), p2.y()),
    amrex::max(p1.z(), p2.z())
);
```

amr-wind/wind_energy/actuator/disk/constant_ct_ops.H
```diff
@@ -0,0 +1,185 @@
...
+    {
+        auto& ddata = data.meta();
+        auto& info = data.info();
+        pp.get("num_force_points", ddata.num_force_pts);
+        // eventually we will change velocity sampling location
+        // pp.get("num_vel_points_r", ddata.num_vel_points_r);
+        // pp.get("num_vel_points_t", ddata.num_vel_points_t);
+        // pp.get("diameters_to_sample_velocity", ddata.diameters_to_sample);
+        pp.get("disk_center", ddata.center);
+        pp.get("disk_normal", ddata.normal_vec);
```
lawrenceccheung, line 25: Hi @psakievich, do you know if it would be possible to include provide turbine `yaw` as an input?  It could be in addition to `disk_normal`, with `yaw` overriding `disk_normal` in case both are there.  I'm thinking that most cases will assume turbine will be on a flat-ish terrain, and be dealing with turbine yaw from the farm layout inputs.

amr-wind/wind_energy/actuator/disk/ActSrcDiskOp.H
```diff
@@ -0,0 +1,133 @@
...
+            const auto dist_on_disk =
+                utils::delta_pnts_cyl(m_origin, m_normal, cc, pos[ip]);
+            const auto& pforce = force[ip];
+
+            const amrex::Real weight_R =
+                utils::linear_basis_1d(dist_on_disk.x(), dR);
+            const amrex::Real weight_T =
+                1.0 / (::amr_wind::utils::two_pi() * R);
+            const amrex::Real weight_N =
+                utils::gaussian1d(dist_on_disk.z(), epsilon);
```
lawrenceccheung, line 116: It just occurred to be that there's a possbility for the gaussian in the z direction could interfere with the upstream blockage measurements, especially if the gaussian has a long tail.  This is something of a hypothetical at this point, and we'll probably need to study the effect of the epsilon.  But we may need to use compact functions, or gaussians with cut-off tails if that is a problem.

lawrenceccheung (commented):
Looking good so far.

amr-wind/wind_energy/actuator/disk/ActSrcDiskOp.H
psakievich: @lawrenceccheung yes that is a good point. My goal is to make this modular enough to select the spreading function in each principle direction in the input deck. So we could add a more compact function if we'd like. Really with this Gaussian it just needs to be selected to match the mesh requirements so it can still be relatively compact.

amr-wind/wind_energy/actuator/disk/constant_ct_ops.H
psakievich: yeah this sounds good

psakievich: @lawrenceccheung do you think we should also add a pitch option as well?

lawrenceccheung: Good thought.  If you're referring to turbine tilt, yes, we should probably add that in (I'd probably call it `tilt` or `shaft_tilt`, basically how much the rotor shaft deviates from horizontal, or how much the rotor disk deviates from vertical).  I don't think it's widely used in various modeling approaches yet, but since it can be a few degrees and noticeable on large turbines, might as well include it now.

amr-wind/wind_energy/actuator/disk/constant_ct_ops.H
```diff
@@ -0,0 +1,185 @@
...
+    void operator()(ConstantCt::DataType& data, const utils::ActParser& pp)
+    {
+        auto& ddata = data.meta();
+        auto& info = data.info();
+        pp.get("num_force_points", ddata.num_force_pts);
+        // eventually we will change velocity sampling location
+        // pp.get("num_vel_points_r", ddata.num_vel_points_r);
+        // pp.get("num_vel_points_t", ddata.num_vel_points_t);
+        // pp.get("diameters_to_sample_velocity", ddata.diameters_to_sample);
+        pp.get("disk_center", ddata.center);
```
lawrenceccheung, line 24: My apologies, I should have mentioned this earlier.  Is it easy to make a change to use `base_position` and `hub_height` instead, so it follows [turbine_utils.cpp](https://github.com/Exawind/amr-wind/blob/6af41101c73b89b1c89053ffab7ef09efcf37ac7/amr-wind/wind_energy/actuator/turbine/turbine_utils.cpp#L29-L31)?  (`disk_center` will then be `base_position` + `[hub_height, 0.0, 0.0]`).  `disk_center` can still be an optional input, but my thought is to standardize the inputs so that it's easier to switch turbine models with the fewest changes possible.

psakievich: yeah that's not a problem.

amr-wind/wind_energy/actuator/disk/constant_ct_ops.cpp
```diff
@@ -0,0 +1,128 @@
...
+           << std::endl;
+}
+
+void required_parameters(ConstantCt::MetaType& meta, const utils::ActParser& pp)
+{
+    pp.get("num_force_points", meta.num_force_pts);
+    pp.get("disk_center", meta.center);
+    pp.get("epsilon", meta.epsilon);
+    pp.get("rotor_diameter", meta.diameter);
+    pp.get("thrust_coeff", meta.thrust_coeff);
```
tonyinme, line 35: I am wondering if it would be best to have a table with velocity and ct such that the thrust coefficient can change according to velocity. If only one ct is given then it would default to constant ct.

test/test_files/constant_ct_disk/constant_ct_disk.i
```diff
@@ -0,0 +1,126 @@
...
+nodal_proj.mg_atol = 1.0e-12
+mac_proj.mg_rtol = 1.0e-6
+mac_proj.mg_atol = 1.0e-12
+
+#¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨#
+#              Sampling                 #
+#.......................................#
+incflo.post_processing = sampling line_sampling averaging
+# Frequency of output for the data
+sampling.output_frequency = 1000
```
tonyinme, line 89: Do you want to have this sampling here even though it is not used?

tonyinme (commented):
hey @psakievich, this is looking good. 
I am wondering if you were thinking of having the option to provide tabulated ct values for different inflow velocities?

test/test_files/constant_ct_disk/constant_ct_disk.i
psakievich: No not really.  Thanks! I'll delete it

amr-wind/wind_energy/actuator/disk/constant_ct_ops.cpp
psakievich: @tonyinme sure that is a good idea.

amr-wind/wind_energy/actuator/disk/constant_ct_ops.H
```diff
@@ -0,0 +1,167 @@
...
+// this will then be spread uniformly over that area in the
+// projection step
+template <>
+struct ComputeForceOp<ConstantCt, ActSrcDisk>
+{
+    void operator()(ConstantCt::DataType& data)
+    {
+        auto& grid = data.grid();
+        auto& ddata = data.meta();
+        const amrex::Real uInf = ddata.reference_velocity & ddata.normal_vec;
```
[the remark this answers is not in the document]
psakievich: @tonyinme no it can be set to be sampled upstream by setting the parameter `diameters_to_sample`.  Basically we sample a projected disk upstream (or downstream if for some reason you set this to a negative value) and average over all the points to get the reference velocity. Then `uInf` is taken as the disk normal component of this value.  

You can also sample in a direction other than the disk normal. So if the turbine is yawed you can sample without yaw to get the "free stream" and then this projection will make sure that only the disk normal component is applied when computing the force. I've update the regression test to show how this works.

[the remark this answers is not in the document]
psakievich: I should add that if you don't add the `diameters_to_sample` flag then it's value will default to 0 and sampling will be at the disk.

amr-wind/wind_energy/actuator/disk/constant_ct_ops.cpp
psakievich: @tonyinme I'm going to add this in a follow on PR.  This one has gotten pretty big.

psakievich:
I'd like to merge at this point. There are some things that still need to be done, but I feel like this is a good break point in terms of PR's.

psakievich:
Converting back to draft. Found a couple more issues in testing.

amr-wind/wind_energy/actuator/disk/uniform_ct_ops.H
```diff
@@ -0,0 +1,201 @@
...
+struct UpdateVelOp<UniformCt, ActSrcDisk>
+{
+    void operator()(UniformCt::DataType& data)
+    {
+        auto& meta = data.meta();
+        const auto& grid = data.grid();
+        auto& refVel = meta.reference_velocity;
+        auto& diskVel = meta.disk_velocity;
+        const int np = meta.num_vel_pts / 2;
+        for (int i = 0; i < np; i++) {
```
lawrenceccheung, line 113: Should `refVel` and `diskVel` be initialized to zero before entering the loop?  I'm getting some strange output velocities when running the disk, but not sure if that's the cause of this.

psakievich: :+1: update coming :)

lawrenceccheung (approved):
I tested out the code, and it's behaving as expected.  Later on, we might add a unit test with a Ct=0 turbine to double check the velocity sampling returns the expected values, but that's not necessary now.

psakievich:
@lawrenceccheung thats a good idea for a test. I'll add it in the next round

michaeljbrazell (approved):
looks good, thank you!

michaeljbrazell:
@jrood-nrel this PR adds a new regression test

Issue #425, opened 2021-06-11
Inflow/outflow mesh refinement


tonyinme:
I am working on creating an example for inflow/outflow simulations with an openfast actuator line turbine and grid refinement. We currently need some workarounds/hacks in order to get this working. 

This workaround was done by following the recommendations from @gantech.

The steps to make it work is as follows:

0. Modify the code in [ABLBoundaryPlane.cpp ](https://github.com/Exawind/amr-wind/blob/main/amr-wind/wind_energy/ABLBoundaryPlane.cpp) by setting `nlevels=1`. If this is not done, the boundary data cannot work if there are more than one refinement levels inside the domain.

1. **Simulation 1:** Run ABL precursor simulation with boundary data sampling. (This works without a problem)

2. **Simulation 2:** Create the refined mesh by running a simulation starting from Simulation 1 for as many time-steps as refinement levels needed and setting `time.regrid_interval = 1` in the input file. For example, if we want 2 levels of refinement for the turbine simulation, the simulation 2 will be run for 2 time-steps in order to create the new refined mesh.

3. **Simulation 3:** Run a new simulation with the inflow/outflow starting from the last time-step of simulation 2 which contains the refined mesh.

Steps 0 and 2 are hacks that we have in order to make it work. Ideally, we would not need those steps.
I am looking for your feedback on how to improve this?

One option is to hardcode `nlevels=1` in [ABLBoundaryPlane.cpp ](https://github.com/Exawind/amr-wind/blob/main/amr-wind/wind_energy/ABLBoundaryPlane.cpp).
Ideally, Simulation 2 would not be needed and simulation 3 would automatically refine the mesh in the first time step.

Please let me know what ideas you have to improve this.

@michaeljbrazell @sayerhs @gantech @marchdf @gdeskos @shashankNREL

sayerhs:
@tonyinme 

- We should not have to hard code `nlevels` in ABLBoundaryPlane. The code should be refactored to only apply refinements for levels that are available in the saved precursor file. However, one question we have to agree on is whether we will have refinement levels touching the precursor inlet boundaries. Handling all use cases is going to be a bit tricky. 
- Simulation 2 should not be required, that was a workaround for getting the IEA simulations done (*horror*). Instead what is needed is a `regrid_on_restart` setup during initialization. Unfortunately, this wasn't implemented when AMR-wind refinement infrastructure was implemented. It is a couple of lines of code in the initialization logic to make this happen.

[...]

PR #426, opened 2021-06-13
Regrid on restart

sayerhs:
Add ability to refine the AMR-Wind mesh upon restart to a user-specified number
of levels. The use-case is performing a turbine simulation from a restart file
containing the ABL precursor solution. In this case, typically the user wants to
only perform regrid once at the beginning of the restarted simulation and then
maintain the nested grids for the duration of the turbine run. The usual process
of restart will not work in this case, as AMReX only adds one level of
refinement for each `regrid` call.

In the proposed implementation, we check if the user has requested more levels
than the finest level available in the restart solution. If true, then we call
regrid for all the new levels after reading the checkpoint file.

The current implementation assumes that if the user has set `max_level` to a
value greater than the number of levels in the checkpoint file, refinement is
intended during initialization. It might be necessary to provide
`SimTime::regrid_on_restart()` flag to control this behavior.

### DIFF
--- a/amr-wind/incflo.cpp
+++ b/amr-wind/incflo.cpp
@@ -65,6 +65,9 @@ void incflo::init_mesh()
     } else {
         // Read starting configuration from chk file.
         ReadCheckpointFile();
+        for (int lev = finestLevel(); lev <= maxLevel(); ++lev) {
+            regrid(lev, m_time.current_time());
+        }
         if (ParallelDescriptor::IOProcessor()) {
             amrex::Print() << "Grid summary: " << std::endl;
             printGridSummary(amrex::OutStream(), 0, finest_level);


sayerhs:
@tonyinme @gantech @michaeljbrazell @marchdf 

Here is a possible solution to fix the _regrid on restart_ issue mentioned in #425.

marchdf:
Hi all, I tried this on Eagle and got something that looks reasonable: 
Screen Shot 2021-06-14 at 8 10 35 AM

This is the `abl_bndry_input.i` file with 
```
tagging.labels = "static"
tagging.static.type = "CartBoxRefinement"
tagging.static.static_refinement_def = "static_box.txt"
```

And the `static_box.txt`:
```
2
1
300.0 300.0 40.0 700.0 700.0 400.0
1
450.0 450.0 100.0 550.0 550.0 200.0
```

marchdf (approved):

amr-wind/incflo.cpp
```diff
@@ -65,6 +65,9 @@ void incflo::init_mesh()
     } else {
         // Read starting configuration from chk file.
         ReadCheckpointFile();
+        for (int lev = finestLevel(); lev <= maxLevel(); ++lev) {
+            regrid(lev, m_time.current_time());
```
marchdf, line 69: Do we want to add some output here, e.g. "Regriding on restart" or something to that effect so that the user isn't confused by the checkpoint file and resulting possible mismatch in grid summary?

marchdf (approved):

amr-wind/incflo.cpp
sayerhs: Yeah, we can do that. I just figured that they had to manually change `max_level` in the input file. ;)

marchdf: Yes but maybe they want confirmation that it did actually regrid? I could go either way on this one.

tonyinme: With the current version, the code does output the refinement levels right after reading the mesh.
Here is the output of the example I ran. 

`Restarting from checkpoint ../amrwind_abl/chk40000
Grid summary:
  Level 0   1024 grids  33554432 cells  100 % of domain
            smallest grid: 32 x 32 x 32  biggest grid: 32 x 32 x 32
  Level 1   800 grids  819200 cells  0.3051757812 % of domain
            smallest grid: 16 x 8 x 8  biggest grid: 16 x 8 x 8
  Level 2   648 grids  4810752 cells  0.2240180969 % of domain
            smallest grid: 24 x 16 x 16  biggest grid: 32 x 16 x 16
`

tonyinme:
This looks good. Thanks for adding the code @sayerhs . Tested it on the inflow/outflow simulation on Eagle with 2 levels of refinement and it works.
Here is a visual with the case I ran.

Screen Shot 2021-06-14 at 9 15 28 AM

tonyinme (approved):

amr-wind/incflo.cpp
sayerhs: @marchdf I'll merge this capability as it removes one of the impediments mentioned in #425. Based on user feedback, we can decide what additional informational messages we should add. Hope that is ok.

PR #428, opened 2021-06-14
More robust ABL IO

### LINKED ISSUE #414: Boundary input file feature on GPU crashing
I am trying to run the regression test case `abl_bndry_input` on GPUs on Eagle. One key difference is that I tried adding 2 levels of refinement far from the boundary through static refinement. As soon as the code initiates a refinement the code crashes. 
I am running the `main` branch. 

```
 AMR-Wind version :: 319ad956
  
 AMR-Wind Git SHA :: 319ad956f0f85fbe2c41fbec97b69f7b63a190c9

  AMReX version    :: 21.05-20-gfb0c16e34b93
```
Following is the error message that I see. Somehow the error is with NetCDF complaining `No group found` which is after running a few time steps. 

```
Regrid mesh ... time elapsed = 0.003602950135
Grid summary:
  Level 0   8 grids  110592 cells  100 % of domain
            smallest grid: 16 x 16 x 16  biggest grid: 32 x 32 x 32
  Level 1   8 grids  64000 cells  7.233796296 % of domain
            smallest grid: 16 x 16 x 16  biggest grid: 24 x 24 x 24
  Level 2   8 grids  64000 cells  0.904224537 % of domain
            smallest grid: 16 x 16 x 16  biggest grid: 24 x 24 x 24

For godunov_type select between plm, ppm, ppm_nolim, weno_js, and weno_z: it defaults to ppm
For godunov_type select between plm, ppm, ppm_nolim, weno_js, and weno_z: it defaults to ppm
Step: 7 dt: 0.4 Time: 2.9 to 3.3
CFL: 0.768155 (conv: 0.768008 diff: 0 src: 0.0106265 )

NetCDF: No group found.

  what():  Encountered NetCDF error; aborting
MPT ERROR: Rank 0(g:0) received signal SIGABRT/SIGIOT(6).
	Process ID: 8057, Host: r104u37, Program: /lustre/eaglefs/scratch/syellapa/Wind/WRF/BplaneTest/amr-wind/build/amr_wind
	MPT Version: HPE MPT 2.22  03/31/20 16:12:29

MPT: --------stack traceback-------
MPT: Attaching to program: /proc/8057/exe, process 8057
MPT: [New LWP 8102]
MPT: [New LWP 8078]
MPT: [New LWP 8077]
MPT: [Thread debugging using libthread_db enabled]
```

On Summit the error shows up as `what():  GPU last error detected in file /gpfs/alpine/cfd142/scratch/syellapa/WRF/amr-wind/submods/amrex/Src/Base/AMReX_GpuLaunchFunctsG.H line 1000: misaligned address
[f12n08:40078] *** Process received signal ***`

@marchdf @sayerhs @jrood-nrel  : Have you seen this kind of error before? Can you help me fix this issue? 

Thanks

### LINKED ISSUE #425: Inflow/outflow mesh refinement

marchdf:
Improvements:
- Output will only happen for levels that touch the output plane (previously it would create groups for levels even if they didn't intersect with the output plane)
- Loops involving levels now loop over the level groups in each plane (previously they looped over the number of levels in the domain, which is not relevant if the levels aren't at an input plane)
-  added a regression test for input with AMR
- removed `and`/`or`/`not`

Closes #414 
Closes #425


marchdf:
This PR let's me do inflow from the right simulation (no AMR) into the left simulation (AMR inside the domain):

Screen Shot 2021-06-14 at 11 20 57 AM

marchdf:
I think this is the use case we are aiming for... @tonyinme can you confirm that this works for your case?

amr-wind/wind_energy/ABLBoundaryPlane.cpp
```diff
@@ -543,8 +556,9 @@ void ABLBoundaryPlane::populate_data(
 
     for (amrex::OrientationIter oit; oit; ++oit) {
         auto ori = oit();
-        if ((not m_in_data.is_populated(ori)) or
-            (fld.bc_type()[ori] != BC::mass_inflow))
+        if ((!m_in_data.is_populated(ori)) ||
+            (fld.bc_type()[ori] != BC::mass_inflow) ||
+            (lev >= m_in_data.nlevels(ori)))
```
sayerhs, line 561: @marchdf If I understand the logic correctly, the code will silently do nothing if there is a refined level that touches the inlet boundary that was not part of the original _precursor simlation_. This would mean that the inflow boundary has wrong values for the finer levels. I don't think a subsequent `fillpatch` will fix this issue either as for inflow it expects the user code to set the boundary values. It might make sense to check for intersection of the finer level box with inflow and throw here?

marchdf: You understood correctly. And yes it probably makes more sense to do check like I do for outflow where I check the intersection of the fine level with the boundary and then throw.

marchdf: Ok I tried to address... but it's tricky. Lmk what you think.

amr-wind/wind_energy/ABLBoundaryPlane.cpp
```diff
@@ -707,5 +728,23 @@ void ABLBoundaryPlane::impl_buffer_field(
                                              k - v_offset[2], n));
         });
 }
+
+// True if box intersects the boundary
```
sayerhs, line 732:
```suggestion
//! True if box intersects the boundary
```

Make doxygen friendly

sayerhs (approved):

PR #443, opened 2021-06-29
Immersed boundary implementation using the direct BC imposition/forcing approach

gdeskos:
Proof-of-concept immersed boundary method for AMR-Wind

- Adding an immersed boundary model
- Available immersed/bluff-body geometries include the cylinder, sphere and box
- Capability for computing the forces on bluff-bodies is coded in but apparently does not work properly yet
- Changes in pressure projection to allow for the masking of the RHS of the poisson solve to obtain a smooth pressure field
- Model has gone through basic Verification & Validation


amr-wind/incflo_advance.cpp
```diff
@@ -258,16 +260,32 @@ void incflo::ApplyPredictor(bool incremental_projection)
         seqn->compute_advection_term(amr_wind::FieldState::Old);
     }
 
+    // Create a cell-centred field by interpolation -- Hard-coded, needs
+    // refactor
+    /****************************************************************************/
+    /*
+    auto& velocity_nph = velocity_new.state(amr_wind::FieldState::NPH);
+    auto& umac = m_sim.repo().get_field("u_mac");
+    auto& vmac = m_sim.repo().get_field("v_mac");
```
michaeljbrazell, line 269: remove this?

gdeskos: done !

PR #446, opened 2021-06-30
ABL boundary plane orientation

michaeljbrazell:
now we can output other orientations with abl boundary plane, still need to test the reading though

### DIFF
--- a/amr-wind/wind_energy/ABLBoundaryPlane.cpp
+++ b/amr-wind/wind_energy/ABLBoundaryPlane.cpp
@@ -484,7 +484,7 @@ void ABLBoundaryPlane::read_header()
             // Create the data structures for the input data
             amrex::IntVect plo(lo);
             amrex::IntVect phi(hi);
-            plo[normal] = ori.isHigh() ? lo[normal] + 1 : -1;
+            plo[normal] = ori.isHigh() ? hi[normal] + 1 : -1;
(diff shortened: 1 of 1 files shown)


amr-wind/wind_energy/ABLBoundaryPlane.cpp
```diff
@@ -232,9 +232,10 @@ void ABLBoundaryPlane::initialize_data()
 
         if ((std::find(valid_planes.begin(), valid_planes.end(), plane) ==
              valid_planes.end())) {
-            throw std::runtime_error(
-                "Requested plane (" + plane +
-                ") does not exist. Pick one of [xlo, ylo].");
+	     amrex::Print() << "WARNING: only xlo and ylo thoroughly tested" << std::endl;
+            //throw std::runtime_error(
```
marchdf, line 236: I think you can remove these now

marchdf (approved):

lawrenceccheung:
Thanks @michaeljbrazell.  I tested this on a small ABL test case with the `xhi/yhi` planes, and it seems to be working well so far.  You can see the case and comparison here: https://github.com/lawrenceccheung/IEATask29_precursor/tree/main/REF1_ABLIO/baseprecursor_WDir050_bcio1.  

I may try to test with a longer run or with a refinement region case, but so far so good!

lawrenceccheung (approved):

PR #449, opened 2021-07-01
Fix typo in type value for AMR tagging docs

diederikb:
Docs said to use the value `VorticityRefinement` instead of `VorticityMagRefinement` for AMR tagging type.


michaeljbrazell (approved):
good catch! thank you for updating the docs

PR #455, opened 2021-07-13
Ascent integration

nicholasbl:
This PR adds initial experimental support for the Ascent in-situ visualization infrastructure.

To use Ascent, an `ascent_actions.yaml` file can be placed in the simulation directory containing the `input.i` file. The infrastructure will then follow the actions and render scenes as described in the yaml. 

Current Issues:
- As identified in the prototype patch, OpenMP support should be consistent across the build. If either Ascent or amr_wind is built with OpenMP and the other is not, there might be runtime issues.
- This requires an as-of-yet unreleased development build of AMReX with important fixes to the Conduit system. These fixes should be released in a version sometime around August 1, but it's probably a good idea to get things an initial review and comments before then.


CMakeLists.txt
```diff
@@ -136,6 +137,17 @@ if (AMR_WIND_ENABLE_OPENFAST)
...
 
+if(AMR_WIND_ENABLE_ASCENT)
+  set(CMAKE_PREFIX_PATH ${ASCENT_DIR} ${CMAKE_PREFIX_PATH})
+  find_package(Ascent QUIET REQUIRED)
+  if (Ascent_FOUND)
+    message(STATUS "Found Ascent = ${ASCENT_INCLUDE_DIRS}")
+  endif()
+  target_include_directories(${amr_wind_lib_name} PUBLIC ${ASCENT_INCLUDE_DIRS})
+  target_link_libraries(${amr_wind_lib_name} PUBLIC ${ASCENT_LIBRARIES})
+  target_compile_definitions(${amr_wind_lib_name} PRIVATE AMR_WIND_USE_ASCENT)
```
sayerhs, line 148: Does ascent provide the modern CMake library definitions?

amr-wind/utilities/IOManager.cpp
```diff
@@ -101,6 +107,64 @@ void IOManager::initialize_io()
...
+    for (int lev = 0; lev < nlevels; ++lev) {
+        int icomp = 0;
+        auto& mf = (*outfield)(lev);
+
+        for (auto* fld : plt_fields) {
+            amrex::MultiFab::Copy(
+                mf, (*fld)(lev), 0, icomp, fld->num_comp(), 0);
+            icomp += fld->num_comp();
+        }
+    }
```
sayerhs, line 136: This makes a third copy of all the output fields which feels unnecessary. Why not just directly pass the `outfield` MultiFab from `write_plot_file` to this function?

amr-wind/utilities/IOManager.cpp
```diff
@@ -101,6 +107,64 @@ void IOManager::initialize_io()
     }
 }
 
+#ifdef AMR_WIND_USE_ASCENT
+static void ascent_pass(
```
sayerhs, line 111: What is the reason for this `static` keyword here?

amr-wind/utilities/IOManager.cpp
```diff
@@ -142,6 +206,10 @@ void IOManager::write_plot_file()
         mesh.Geom(), m_sim.time().new_time(), istep, mesh.refRatio());
 
     write_info_file(plt_filename);
+
+#ifdef AMR_WIND_USE_ASCENT
+    ascent_pass(m_sim, m_plt_num_comp, m_plt_fields, m_plt_var_names);
+#endif
```
sayerhs, line 212: This ties ascent output frequency to plot file write frequency. I would imagine that if in-situ visualization works we would most likely want a higher frequency of output for in-situ processed data. Instead of bolting on in-situ viz logic onto IOManager, Ascent capability should be a separate PostProcessing instance that can be activated independently of plot file output.

CMakeLists.txt
nicholasbl: Looks like they do; I'll update this.

amr-wind/utilities/IOManager.cpp
nicholasbl: Good point, I'll explore that approach.

amr-wind/utilities/IOManager.cpp
nicholasbl: This function has internal linkage.

amr-wind/utilities/IOManager.cpp
nicholasbl: Sounds good. Do you have a pointer to the best example to follow in creating a PostProcessing pass?

sayerhs: The base class for post-processing is here: 

https://github.com/Exawind/amr-wind/blob/de3fd119f2e53b9360c6c9b6f9412521e4643497/amr-wind/utilities/PostProcessing.H#L20-L29

you can look at the `sampling` example which uses probes and planes to extract subset of data in NetCDF or text formats: 

https://github.com/Exawind/amr-wind/blob/de3fd119f2e53b9360c6c9b6f9412521e4643497/amr-wind/utilities/sampling/Sampling.H#L27-L40

An example of the input file:

https://github.com/Exawind/wind-energy/blob/28e8a9805653aaf3945a1d6ff379b51580545e8b/Pedersen_N02/AmrWind/abl.2.inp#L92-L107

nicholasbl: Excellent, thanks very much for your feedback!

sayerhs (approved):

jrood-nrel:
We will merge this when the necessary AMReX version is released.

jrood-nrel (approved):

michaeljbrazell (approved):

nicholasbl:
Updated patch to use the latest amrex release, which contains required fixes.

nicholasbl:
Looks like this passed all the checks. Is there anything else I should do?

PR #456, opened 2021-07-14
Add ct table option for ct disk

psakievich:
Add table input for ct vs wind speed to the input deck


psakievich:
I didn't update the regression test for this one.  I wanted to confirm it is working. I've tested the table input as well.  I can either update the current regression test, or add a second one.  Thoughts?

tonyinme (approved):
Looks good!

michaeljbrazell (commented):
I got a NaN when running regression tests. So I vote update the regression test input file

psakievich:
@michaeljbrazell yeah sorry that's a bug I need to fix. I noticed it after I pushed this. I'll update soon.

amr-wind/wind_energy/actuator/disk/uniform_ct_ops.cpp
```diff
@@ -192,7 +192,7 @@ void optional_parameters(UniformCt::MetaType& meta, const utils::ActParser& pp)
     if (pp.contains("yaw")) {
         amrex::Real yaw;
         pp.get("yaw", yaw);
-        normalRotOp = normalRotOp & vs::zrot(yaw);
+        normalRotOp = normalRotOp & vs::zrot(-yaw);
```
lawrenceccheung, line 195: Oh, this isn't a critical point, but I would add comments here that indicate the `-yaw` is so that the rotation is applied in a clockwise direction starting from north = zero degrees.  Mostly so that future me can remember how the conventions worked and prevent additional confusion.

psakievich: yeah that's a good point. I'll add that

PR #462, opened 2021-07-28
Corrected S2 and W2 definitions for Q criterion

diederikb:
[Hussain, F. (1995) On the identification of a vortex. JFM 285, 69-94](https://www.researchgate.net/publication/231865270_Hussain_F_On_the_identification_of_a_vortex_JFM_285_69-94), [Haller, G. (2005). An objective definition of a vortex. Journal of Fluid Mechanics, 525, 1-26](http://www.georgehaller.com/reprints/vortex.pdf), and others use the following definition for the Q-criterion:

with  and  the symmetric and anti-symmetric parts, respectively, of the gradient of velocity tensor.

With these definitions, the tensors are

and

.

Therefore, the square of their Frobenius norm are

and

.

### DIFF
--- a/amr-wind/fvm/qcriterion.H
+++ b/amr-wind/fvm/qcriterion.H
@@ -77,9 +77,9 @@ struct Qcriterion
                      idx[2];
 
                 const amrex::Real S2 =
-                    std::pow(ux, 2) + std::pow(vy, 2) +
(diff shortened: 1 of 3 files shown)


michaeljbrazell (approved):
nice catch!

PR #468, opened 2021-08-05
Changed KE and enstrophy sampling formatting

diederikb:
Changed formatting from std::fixed to std::scientific and increased printed column widths by four characters to accommodate for 'e+xx' part.


michaeljbrazell (approved):
looks good

PR #473, opened 2021-08-23
added bdryHi to wall model

michaeljbrazell:


michaeljbrazell:
@neilmatula is this working? should we merge?

neilmatula:
Yes, it works.  We can merge.

michaeljbrazell:
there might be a buh, I think we should check the inputs on the hi side... I'll look into this

PR #475, opened 2021-08-25
Fat-cored vortex ring

diederikb:
This PR introduces a new physics type: the fat-cored vortex ring. This case is from [Yu et al. (2020)](https://arxiv.org/pdf/2010.13213.pdf) and can be used for a grid convergence study with dynamic mesh refinement, similar to the paper. The fat-cored vortex ring is specified in terms of its vorticity field and the velocity field is initialized by first solving the Poisson equation for the vector potential and then taking the discrete curl of the vector potential to obtain the velocity.


amr-wind/physics/VortexRing.H
```diff
@@ -0,0 +1,90 @@
...
+
+    void post_regrid_actions() override {}
+
+    void pre_advance_work() override {}
+
+    void post_advance_work() override {}
+
+private:
+    template <typename VortexRingType>
+    void initialize_velocity(const VortexRingType& core);
```
michaeljbrazell, line 34: move this up to public

amr-wind/physics/VortexRing.H
```diff
@@ -0,0 +1,90 @@
...
+ */
+class VortexRing : public Physics::Register<VortexRing>
+{
+public:
+    static const std::string identifier() { return "VortexRing"; }
+
+    explicit VortexRing(const CFDSim& sim);
+
+    virtual ~VortexRing() = default;
+    //! Initialize the temperature and velocity fields for BoussinesqBubble
```
michaeljbrazell, line 20: delete this comment

michaeljbrazell:
Looking good! We can move some of the linear solver stuff out later

PR #476, opened 2021-08-30
Fix UniformCtDisk uInfSqr calculation and force direction

lawrenceccheung:
The UniformCtDisk has a bug where it calculates zero forces for certain wind directions and turbine yaw's (basically wind from 315 or 135, and turbine pointed directly into the wind).  This PR should fix that calculation.

Lawrence


psakievich (approved):
Thanks for fixing this

PR #477, opened 2021-09-03
Error check utility

diederikb:
This PR adds a tool that calculates the relative L2 and the L infinity norm of the error between a multilevel plotfile and a reference plotfile with a single level that is finer than the finest level of the multilevel plotfile.


michaeljbrazell:
looks good! thank you. We'll add cmake to it later

PR #478, opened 2021-09-08
multicomponent solve

michaeljbrazell:
- switched to multicomponent version of scalar solve in hopes to speed up velocity solve
- warning this still does not treat the off diagonal terms correctly for variable viscosity flows

@jrood-nrel this will cause a very small diff in abl_godunov_scalar_velocity_solve, please rebless


michaeljbrazell:
I'm going to wait before merging this since the overset interface (in amrex) needs updating before switching to this

gdeskos (approved):

PR #479, opened 2021-09-08
Fix install errors with ninja

psakievich:
amr-wind fails to install correctly when using `ninja` as the build generator.  This is because `ninja` is looking for fortran files to install. Since we don't actually build with fortran we should not enable it as a compiler language.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,7 +47,6 @@ set(CMAKE_CXX_STANDARD_REQUIRED ON)
 if(AMR_WIND_ENABLE_CUDA)
   enable_language(CUDA)
   # Fix issues with GPU builds
-  enable_language(Fortran)
   if(CMAKE_CUDA_COMPILER_VERSION VERSION_LESS "10.0")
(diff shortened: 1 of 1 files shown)


jrood-nrel:
I'm fine with this, but does OpenFAST work without it?

CMakeLists.txt
```diff
@@ -47,7 +47,6 @@ set(CMAKE_CXX_STANDARD_REQUIRED ON)
 if(AMR_WIND_ENABLE_CUDA)
   enable_language(CUDA)
   # Fix issues with GPU builds
```
jrood-nrel, line 49: Let's remove this comment too then.

psakievich:
@jrood-nrel  I built with OpenFAST and it worked fine

CMakeLists.txt
psakievich: I haven't tested this with a GPU build+ninja.  Should we try that first?

jrood-nrel: No. If we don't need these, then they should be removed regardless.

jrood-nrel (approved):

PR #481, opened 2021-09-09
replicate feature

michaeljbrazell:
- useful for weak scaling studies
- can be used to initialize a larger ABL run from a smaller restart file

if `amr.n_cell = 48 48 48` and `geometry.prob_hi=1000 1000 1000` to replicate in the x direction use a restart file and set
`amr.n_cell = 96 48 48` and `geometry.prob_hi=2000 1000 1000`

can also shift the domain at the same time:

original inputs:
`amr.n_cell = 48 48 48
geometry.prob_lo = 0 0 0
geometry.prob_hi = 1000 1000 1000`

new inputs:
`amr.n_cell = 96 144 48
geometry.prob_lo = -1000 0 0
geometry.prob_hi = 1000 3000 1000`

This replicated twice in x-dir and three times in y-dir, also translated domain in x by -1000.0

### DIFF
--- a/amr-wind/utilities/io.cpp
+++ b/amr-wind/utilities/io.cpp
@@ -21,13 +21,6 @@ void incflo::ReadCheckpointFile()
 {
     BL_PROFILE("amr-wind::incflo::ReadCheckpointFile()");
 
-    //    amrex::ParmParse pp("io");
-    //    amrex::Vector<int> repvec{{1, 1, 1}};
-    //    pp.queryarr("replicate", repvec);
-    //    IntVect rep(repvec[0], repvec[1], repvec[2]);
-    //    IntVect rep(1,1,1);
-    //    bool replicate = (rep == IntVect::TheUnitVector()) ? false : true;
-
     const std::string& restart_file = m_sim.io_manager().restart_file();
     amrex::Print() << "Restarting from checkpoint " << restart_file
                    << std::endl;
--- a/amr-wind/utilities/io.cpp
+++ b/amr-wind/utilities/io.cpp
@@ -115,8 +115,20 @@ void incflo::ReadCheckpointFile()
     IntVect rep(1, 1, 1);
     for (int d = 0; d < AMREX_SPACEDIM; d++) {
         AMREX_ALWAYS_ASSERT(prob_hi[d] - prob_lo[d] > 0.0);
-        rep[d] = static_cast<int>(
-            (prob_hi_input[d] - prob_lo_input[d]) / (prob_hi[d] - prob_lo[d]));
+
+        const amrex::Real domain_ratio =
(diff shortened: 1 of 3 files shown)


amr-wind/utilities/io.cpp
```diff
@@ -21,6 +21,13 @@ void incflo::ReadCheckpointFile()
 {
     BL_PROFILE("amr-wind::incflo::ReadCheckpointFile()");
 
+    amrex::ParmParse pp("amr");
```
sayerhs, line 24: Why not have this in the IO namespace?

amr-wind/utilities/io.cpp
```diff
@@ -21,6 +21,13 @@ void incflo::ReadCheckpointFile()
...
+    amrex::Vector<int> rep{{1, 1, 1}};
+    pp.queryarr("replicate", rep);
+    IntVect Nrep(rep[0], rep[1], rep[2]);
```
sayerhs, line 27:
```suggestion
    IntVect nrep(rep[0], rep[1], rep[2]);
```

amr-wind/utilities/io.cpp
```diff
@@ -96,6 +103,13 @@ void incflo::ReadCheckpointFile()
         }
     }
 
+    if (replicate) {
+        for (int d = 0; d < BL_SPACEDIM; d++) {
+            prob_lo[d] = Nrep[d] * prob_lo[d];
```
sayerhs, line 108: This seems like it can lead to confusion. Since restart files have the old prob lo/hi would it make more sense to compare user provided lo/hi and compute nrep?

sayerhs: Also this only works as advertised if problo is (0,0,0). Otherwise it seems that your domain will be 2x times larger and cells coarser

michaeljbrazell: yeah it is confusing, I'll change it up. Thanks!

michaeljbrazell:
Screen Shot 2021-09-09 at 4 49 56 PM

amr-wind/utilities/io.cpp
```diff
@@ -108,25 +149,77 @@ void incflo::ReadCheckpointFile()
...
+
     for (int lev = 0; lev <= finest_level; ++lev) {
         // read in level 'lev' BoxArray from Header
         ba_inp[lev].readFrom(is);
         GotoNextLine(is);
 
+        // always use level 0 to check domain size
+        Box orig_domain(ba_inp[0].minimalBox());
+
+        if (replicate && lev == 0) {
```
sayerhs, line 161: You could move this outside the for loop

sayerhs (approved):

PR #487, opened 2021-09-24
rectangular domain bug

michaeljbrazell:
fixed bug in compute zi and second moment averaging when using rectangular domains


lawrenceccheung (commented):
I tested out the changes to the second moment avg and `compute_zi()` on ABL's with both `Ny>Nx` and `Nx>Ny`.  Seems to be working so far, no more core dumps with memory errors.  I'll also double check the third moment changes.

Lawrence

PR #488, opened 2021-10-01
fix time indices for output and checkpoint file 

ashesh2512:
When restarting, currently the time step indices do not take into account the time index of the restart file. As a result the checkpoint and output files are plotted at intervals inconsistent with the requested interval.

### DIFF
--- a/amr-wind/core/SimTime.cpp
+++ b/amr-wind/core/SimTime.cpp
@@ -169,12 +169,14 @@ bool SimTime::do_regrid()
 
 bool SimTime::write_plot_file()
 {
-    return ((m_plt_interval > 0) && (m_time_index % m_plt_interval == 0));
+    return ((m_plt_interval > 0) &&
+            ((m_time_index - m_start_time_index) % m_plt_interval == 0));
 }
 
 bool SimTime::write_checkpoint()
 {
-    return ((m_chkpt_interval > 0) && (m_time_index % m_chkpt_interval == 0));
+    return ((m_chkpt_interval > 0) &&
+            ((m_time_index - m_start_time_index) % m_chkpt_interval == 0));
 }
 
 bool SimTime::write_last_plot_file()


ashesh2512:
@gantech and @shashankNREL could you please review this ?

michaeljbrazell:
should we do the same for write_last_plot_file and write_last_checkpoint? what about regrid?

ashesh2512:
> should we do the same for write_last_plot_file and write_last_checkpoint? what about regrid?

@michaeljbrazell Maybe for regrading, but I don't suppose it makes a difference for the last plot file and last checkpoint file because those still hold valid, i.e. the last time step will print and output file and a checkpoint file.

michaeljbrazell:
it won't print the last file in some situations... I suggest adding it there too

ashesh2512:
@michaeljbrazell I have modified the other functions too in 15b4751

sayerhs:
@ashesh2512 Can you explain the objective behind this PR? I don't quite follow your original message, and especially the logic behind changing `plot` and `checkpoint` intervals. Can you give an example? 

This is how I expect the behavior. Let's say that I intent to output plot files every 10 timesteps but write restarts after every 25 timesteps. Now if the restart was saved in the 75th timestep, and we restart from there, based on the modifications the restart will start writing outputs `[75, 85, ...]` where as the original outputs were `[0, 10, ... 70]` and I would expect it to continue outputs at `[80, 90, ...]`.

ashesh2512:
@sayerhs I see your point. 

The use case behind the PR is - For instance, I run a precursor ABL run in AMR-Wind for 40,200 time steps. For blade-resolved simulations we couple AMR-Wind with a near-body solver like Nalu-Wind, and we want to output checkpoint files every 1440 time steps (which is 1 rotation at 0.25 degree time step). In such a case, the first checkpoint file output by AMR-Wind would be at 40,320 time steps based on the current logic instead of 40,200 + 1440 time steps. As a result the restart files that are output would be inconsistent between the two solvers. 

Alternatively, I could change the logic to - 
`    return ((m_chkpt_interval > 0) &&
            ((m_time_index % m_chkpt_interval == 0) ||
             (m_time_index - m_start_time_index) % m_chkpt_interval == 0));`

gantech (approved):

sayerhs:
@ashesh2512 Ah yes, the annoying restart from precursor issue. Your suggestion will output additional restart files and that might not be great for disk usage. My approach would be to provide additional user-configurable parameters (something along the lines of `chkpt_start_index` and/or `chkpt_interval_from_restart`) that allows you to control the behavior. Defaulting the `plt` and `chkpt` start indices to 0 would retain the original behavior for standalone amr-wind runs. The idea would mimic the `output_start` and `restart_start` options in nalu-wind. The downside is that it would make the logic in `SimTime` methods more involved, but it would be configuration and provide a lot of control for different use cases in future.

ashesh2512:
@sayerhs I have updated the logic in 5bc9e39

New parameters have been introduced to allow flexibility based on user defined start index.

ashesh2512:
@gantech I have set the default values of the restart indices in e118f4b 

Could you please review this PR again ?

PR #494, opened 2021-10-26
ActDisk: Multiple Spread functions

psakievich:
This commit allows users to specify multiple spreading models for the UniformCtDisk model.
This ~~requires~~ allows you to include the `spreading` key in the input file.  This is a prototype to keep actuator types from becoming too verbose.

For example a classic ADM would be specified as:
```
Actuator.type = UniformCtDisk
Actuator.spreading = UniformGaussian
```

and the new one we've been testing out for TCF would be specified as:
```
Actuator.type = UniformCtDisk
Actuator.spreading = LinearBasis
```

The rational for adding this is to give the _defacto_ standard of a uniform Gaussian sampled at points as an option.  The `LinearBasis` model we first put in will likely be faster, but there may be some corner cases where it would be preferable to switch.  This also allows us to run direct comparisons of the two models in the same code.  

~~I still need to test that the uniform Gaussian is working correct for this case, and I need to add some parsing checks so users get intelligible errors if they forget the new `spreading` parameter. I'm thinking I'll just add a default but I need to work through it.~~

Update:  I made it so it defaults to the first model we put in `LinearBasis`

Eventually, I'm envisioning getting to where the spreading function is fully configurable in all 3 principle directions for all models. In this case our `LinearBasis` model could be specified as:
```
Actuator.spreading = Linear Linear Gaussian
```
and you could also do something like:

```
Actuator.spreading = Gaussian Linear Gaussian
```
if you only wanted the linear basis function in the theta direction.


amr-wind/wind_energy/actuator/disk/disk_spreading.H
```diff
@@ -0,0 +1,170 @@
...
+        const amrex::Geometry& geom)
+    {
+        (this->*m_function)(actObj, lev, mfi, geom);
+    }
+
+    SpreadingFunction(const SpreadingFunction&) = delete;
+    void operator=(const SpreadingFunction&) = delete;
+
+    void (SpreadingFunction::*m_function)(
+        T* actObj, const int, const amrex::MFIter&, const amrex::Geometry&);
```
sayerhs, line 37: Is a virtual function call much slower than a call to via a function pointer?

psakievich: probably not

sayerhs: Thanks, so we probably could have done without this complexity?

psakievich: probably. if you have a suggestion or requested change let me know.  There are several ways to do a design. The actual interface for this is not really any more complicated than a virtual function call. It could be done either way.  Off the top of my head, with a virtual call I could could use a unique_ptr with some logic to select the right object type based on the string handle I'm using to select the function in `initialize()`.  I don't see how that is much different than this. Not sure if it is worth a re-write at this point.

If you are going to just ask me leading questions to criticize the code then I will remove you as a reviewer. If you have an actual suggestion to make the code simpler and cleaner I'm all ears.

amr-wind/wind_energy/actuator/disk/disk_spreading.H
```diff
@@ -0,0 +1,170 @@
...
+ * https://www.codeproject.com/Articles/11015/The-Impossibly-Fast-C-Delegates
+ *
+ * @tparam T
+ */
+template <typename T>
+class SpreadingFunction
+{
+public:
+    void operator()(
+        T* actObj,
```
sayerhs, line 25: Is it possible to pass a const reference instead of a pointer here? 

```C++
const T& actObj
```

amr-wind/wind_energy/actuator/disk/disk_spreading.H
```diff
@@ -0,0 +1,170 @@
...
+        const int lev,
+        const amrex::MFIter& mfi,
+        const amrex::Geometry& geom)
+    {
+        const auto& bx = mfi.tilebox();
+        const auto& sarr = actObj->m_act_src(lev).array(mfi);
+        const auto& problo = geom.ProbLoArray();
+        const auto& dx = geom.CellSizeArray();
+
+        auto& data = actObj->m_data.meta();
```
sayerhs, line 50:
```suggestion
        const auto& data = actObj->m_data.meta();
```

amr-wind/wind_energy/actuator/disk/uniform_ct_ops.H
```diff
@@ -5,6 +5,7 @@
 #include "amr-wind/wind_energy/actuator/actuator_ops.H"
 #include "amr-wind/utilities/tensor_ops.H"
 #include "amr-wind/utilities/linear_interpolation.H"
+#include "amr-wind/utilities/io_utils.H"
```
sayerhs, line 8: Is this left over from refactor? Would the code compile if this was moved back to the `.cpp` file?

amr-wind/wind_energy/actuator/disk/disk_spreading.H
```diff
@@ -0,0 +1,170 @@
...
+    SpreadingFunction() : m_function(&SpreadingFunction::linear_basis_spreading)
+    {}
+    void initialize(std::string key)
+    {
+        if (key == "UniformGaussian") {
+            m_function = &SpreadingFunction::uniform_gaussian_spreading;
+        } else if (key == "LinearBasis") {
+            m_function = &SpreadingFunction::linear_basis_spreading;
+        } else {
+            throw std::runtime_error("Invalide spreading type");
```
sayerhs, line 163:
```suggestion
            amrex::Abort("Invalid spreading type");
```

amr-wind/wind_energy/actuator/disk/disk_spreading.H
```diff
@@ -0,0 +1,170 @@
...
+
+                sarr(i, j, k, 0) += src_force[0];
+                sarr(i, j, k, 1) += src_force[1];
+                sarr(i, j, k, 2) += src_force[2];
+            });
+    }
+
+    SpreadingFunction() : m_function(&SpreadingFunction::linear_basis_spreading)
+    {}
+    void initialize(std::string key)
```
sayerhs, line 156:
```suggestion
    void initialize(const std::string& key)
```

amr-wind/wind_energy/actuator/disk/disk_spreading.H
```diff
@@ -0,0 +1,170 @@
+#ifndef DISK_SPREADING_H_
+#define DISK_SPREADING_H_
+
+#include "amr-wind/wind_energy/actuator/actuator_utils.H"
+#include "amr-wind/core/FieldRepo.H"
+#include <stdexcept>
```
sayerhs, line 6: Can remove this once switched to `amrex::Abort`

lawrenceccheung (approved):

amr-wind/wind_energy/actuator/disk/disk_spreading.H
psakievich: yeah it is. thanks good catch.

amr-wind/wind_energy/actuator/disk/disk_spreading.H
```diff
@@ -0,0 +1,172 @@
...
+        const auto& bx = mfi.tilebox();
+        const auto& sarr = actObj.m_act_src(lev).array(mfi);
+        const auto& problo = geom.ProbLoArray();
+        const auto& dx = geom.CellSizeArray();
+
+        const auto& data = actObj.m_data.meta();
+
+        const vs::Vector epsilon = vs::Vector::one() * data.epsilon;
+        const vs::Vector m_normal(data.normal_vec);
+        const vs::Vector m_origin(data.center);
```
michaeljbrazell, line 56: unused variable? m_origin and m_coplanar

psakievich: ah nice! yeah I changed the algorithm so we don't need those. I'll remove them now

michaeljbrazell (approved):
looks good

PR #496, opened 2021-10-27
Added a test filter (Delta=2dx) plus a Favre filtering operation for vof interface smoothing

gdeskos:
- Test filter that resembles the Gaussian filter at 2*dx
- Introducing interface smoothing by adding a Favre filter operation in the multiphase class
- Small change in the sloshing_tank regression test file


michaeljbrazell (approved):
looks good, I ran reg tests on my mac and only the sloshing tank had reg test diffs (which is expected)

PR #502, opened 2021-11-04
Virtual lidar

tonyinme:
This adds the feature of a virtual lidar to the code.
The LidarSampler class is derived from the LineSampler class.
It takes a time history of azimuth and elevation angles and creates a line with uniformly distributed points at every time and samples the fields requested.


amr-wind/utilities/sampling/LidarSampler.cpp
```diff
@@ -0,0 +1,91 @@
...
+    vs::Tensor r1 = vs::yrot(m_current_elevation) & vs::zrot(m_current_azimuth);
+
+    // Perform the vector rotation
+    beam_vector = r1 & beam_vector;
+
+    // Add the origin location to the beam vector
+    for (int d = 0; d < AMREX_SPACEDIM; ++d) {
+        beam_vector[d] += m_origin[d];
+        m_end[d] = beam_vector[d];
+    }
```
sayerhs, line 82:
```suggestion
    const vs::Vector beam_vector = {m_length, 0., 0.};
    // The rotation matrix (takes in angles in degrees)
    vs::Tensor r1 = vs::yrot(current_elevation) & vs::zrot(current_azimuth);
    
    m_end = m_start + (r1 & beam_vector);
```

amr-wind/utilities/sampling/Sampling.cpp
```diff
@@ -65,25 +65,47 @@ void Sampling::initialize()
...
 {
+
     BL_PROFILE("amr-wind::Sampling::post_advance_work");
     const auto& time = m_sim.time();
     const int tidx = time.time_index();
     // Skip processing if it is not an output timestep
     if (!(tidx % m_out_freq == 0)) return;
 
+    pre_advance_work();
+
```
sayerhs, line 108: It is misleading to call `pre_advance_work` within `post_advance_work`. Please fix this. Looking at the logic, it appears that we don't need to add `pre_advance_work` but instead rename the function `update_sampling_locations` in `Sampling` class too.

amr-wind/utilities/sampling/SamplerBase.H
```diff
@@ -45,6 +45,9 @@ public:
     //! Populate the vector with coordinates of the sampling locations
     virtual void sampling_locations(SampleLocType&) const = 0;
 
+    //! Update the sampling locations
+    virtual void update_sampling_locations() {}
+
```
sayerhs, line 50:
```suggestion
```

Not needed based on the actual final implementation of `Sampling.cpp`

amr-wind/utilities/sampling/LidarSampler.H
```diff
@@ -0,0 +1,51 @@
...
+namespace amr_wind {
+namespace sampling {
+
+/** Sample data along a line
+ *  \ingroup sampling
+ *
+ *  Defines probes along a line defined by `start` and `end` coordinates and
+ *  divides it into equal segements as determined by `num_points`. The length of
+ *  each segment is given by \f[ (end - start) / (num_points - 1) \f]
+ */
```
sayerhs, line 17: Also update the documentation to add Lidar specific details?

amr-wind/utilities/sampling/LidarSampler.H
```diff
@@ -0,0 +1,51 @@
...
+    void initialize(const std::string& key) override;
+
+    /** Update the position of the sampling locations in time
+     *
+     *
+     */
+    void update_sampling_locations() override;
+
+private:
+    amrex::Vector<amrex::Real> m_origin;
```
sayerhs, line 38:
```suggestion
    vs::Vector m_origin;
```

amr-wind/utilities/sampling/LidarSampler.H
```diff
@@ -0,0 +1,51 @@
...
+    amrex::Vector<amrex::Real> m_time_table;
+    amrex::Vector<amrex::Real> m_azimuth_table;
+    amrex::Vector<amrex::Real> m_elevation_table;
+    amrex::Vector<amrex::Real> m_distribution;
```
sayerhs, line 42:
```suggestion
```

Looks like an unused variable.

amr-wind/utilities/sampling/LidarSampler.H
```diff
@@ -0,0 +1,51 @@
...
+    amrex::Real m_current_azimuth{0};
+    amrex::Real m_current_elevation{0};
```
sayerhs, line 44:
```suggestion
```

These need not be instance variables

amr-wind/utilities/sampling/LidarSampler.cpp
```diff
@@ -0,0 +1,91 @@
...
+{
+
+    amrex::Real time = m_sim.time().current_time();
+
+    // The current azimuth angle
+    m_current_azimuth =
+        ::amr_wind::interp::linear(m_time_table, m_azimuth_table, time);
+
+    m_current_elevation =
+        ::amr_wind::interp::linear(m_time_table, m_elevation_table, time);
```
sayerhs, line 62:
```suggestion
    // The current azimuth angle
    const amrex::Real current_azimuth =
        ::amr_wind::interp::linear(m_time_table, m_azimuth_table, time);

    const amrex::Real current_elevation =
        ::amr_wind::interp::linear(m_time_table, m_elevation_table, time);
```

amr-wind/utilities/sampling/LidarSampler.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+    // The number of points
+    pp.get("num_points", m_npts);
+
+    // The length of the beam [m]
+    pp.get("length", m_length);
+
+    // The time table [s]
+    pp.getarr("time_table", m_time_table);
+    // Azimuth angle (this is in the N, E, S, W direction) [degrees]
```
sayerhs, line 29: Is this compass direction values or angle with reference to the inertial reference frame axes?

amr-wind/utilities/sampling/Sampling.cpp
```diff
@@ -65,25 +65,47 @@ void Sampling::initialize()
...
     m_scontainer->Redistribute();
     m_scontainer->num_sampling_particles() = m_total_particles;
+}
 
-    if (m_out_fmt == "netcdf") prepare_netcdf_file();
+void Sampling::pre_advance_work()
+{
+    BL_PROFILE("amr-wind::Sampling::pre_advance_work");
+
+    SamplerBase::SampleLocType locs;
```
sayerhs, line 90:
```suggestion
```

sayerhs (commented):
Are you planning to also update the input file documentation with `LidarSampler` inputs?

sayerhs:
I forgot to mention this earlier. I seems to me that the `LidarSampler` outputs would show up as `LineSampler` in the NetCDF output. I suspect we want to fix that. Might also want to output the position time-histories so as to avoid additional computations in the post-processing code.

amr-wind/utilities/sampling/LidarSampler.H
```diff
@@ -0,0 +1,54 @@
...
+
+    amrex::Vector<amrex::Real> m_origin;
+    amrex::Vector<amrex::Real> m_time_table;
+    amrex::Vector<amrex::Real> m_azimuth_table;
+    amrex::Vector<amrex::Real> m_elevation_table;
+    amrex::Vector<amrex::Real> m_distribution;
+	amrex::Real m_current_azimuth{0};
+	amrex::Real m_current_elevation{0};
+	amrex::Real m_length{0};
+
```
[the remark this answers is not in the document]
tonyinme: Not sure what this is. The clang-formatter fixes this. I changed the editor to only use spaces.

amr-wind/utilities/sampling/LidarSampler.cpp
```diff
@@ -0,0 +1,94 @@
...
+{
+	// Read in new inputs specific to this class
+    amrex::ParmParse pp(key);
+
+	// This is the origin of the scan (x, y, z) [m]
+	pp.getarr("origin", m_origin);
+
+	// The number of points
+    pp.get("num_points", m_npts);
+
```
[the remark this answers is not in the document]
tonyinme: done

amr-wind/utilities/sampling/LidarSampler.cpp
```diff
@@ -0,0 +1,94 @@
...
+        beam_vector[d] += m_origin[d];
+        m_end[d] = beam_vector[d];
+	}
+	
+}
+
+
+} // namespace sampling
+
+// Register the derived class as a samplerBase
```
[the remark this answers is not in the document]
tonyinme: done

amr-wind/utilities/sampling/LidarSampler.cpp
tonyinme: Yes, this is an inertial frame. I specified the orientation in the documentation.

tonyinme:
> Are you planning to also update the input file documentation with `LidarSampler` inputs?

Yes, this is done now.



> I forgot to mention this earlier. I seems to me that the `LidarSampler` outputs would show up as `LineSampler` in the NetCDF output. I suspect we want to fix that. Might also want to output the position time-histories so as to avoid additional computations in the post-processing code.

I'm not sure how to do this. Where is the code that does this part? I checked the amrex particles and those write out the location.

amr-wind/utilities/sampling/LidarSampler.cpp
tonyinme: I could not get this to work. m_start and m_end are inherited and they are `amrex::Vector<amrex::Real>`.

amr-wind/utilities/sampling/SamplerBase.H
tonyinme: The implementation in `Sampling.cpp` calls this function. Should I be doing this differently?

sayerhs:
@tonyinme 

> I'm not sure how to do this. Where is the code that does this part? I checked the amrex particles and those write out the location.

So currently the Sampling code will only write out the coordinate locations at the beginning of the simulation. It won't add a time dimension to the coordinates. It would be nice to either add the coordinates, or the time/elevation/azimuth tables to the NetCDF output so that the data is all contained in one place for future post-processing.

amr-wind/utilities/sampling/Sampling.cpp
```diff
@@ -220,6 +228,19 @@ void Sampling::write_netcdf()
...
+    for (const auto& obj : m_samplers) {
+        count[1] = obj->num_points();
+        auto grp = ncf.group(obj->label());
+        obj->sampling_locations(locs);
+        auto xyz = grp.var("coordinates");
+        count3[1] = obj->num_points();
+        xyz.put(&locs[0][0], start3, count3);
+    }
+
```
sayerhs, line 243: This probably isn't doing what you intended. You still only have one set of locations in NetCDF file at any given time step. For stationary samplers, you're writing the same data over and over again every output timestep. Where as for non-stationary samplers, you'll end up overwriting previous locations with the latest positions and won't get a time series. 

What was probably intended is to allow LidarSampler to create additional time-series entries to allow output of the positions as a time series entry in NetCDF file.

michaeljbrazell: doesn't nt move the starting point in start 3? looks like line 236 should be deleted since 240 is what's used.

sayerhs: Sorry, I overlooked the change in line 199. 

This is ugly though.

tonyinme: @sayerhs , what would be a better way of doing this?

sayerhs: @tonyinme I was hoping for a more flexible way of doing this so that stationary samplers didn't have to dump the same data over and over again, and providing more flexibility to specific samplers (like Lidar) to provide additional output. 

Using the example of `PlaneSampler` we currently have two functions to customize output behavior: 

https://github.com/Exawind/amr-wind/blob/d67e6e2db9b4d73acc05d516ba23d5ab0f6b839f/amr-wind/utilities/sampling/PlaneSampler.cpp#L75
https://github.com/Exawind/amr-wind/blob/d67e6e2db9b4d73acc05d516ba23d5ab0f6b839f/amr-wind/utilities/sampling/PlaneSampler.cpp#L88

My proposal would be to add these method overrides to `LidarSampler` customize the data output. Add 

- In addition to start/stop, also add the tables of azimuth and elevation
- Delcare a new entry `position` that has the exact coordinates of sampling as a function of time. This is probably not needed as it can be recomputed from the azimuth/elevation tables, but is nice to output. 
- use `populate_netcdf_metadata` to output the tables once for the LidarSampler
- Create a new virtual method (say `output_netcdf_data`) - You will have to add this to SamplerBase also - and then customize this to output the sampling locations for the LidarSampler. Call this in `Sampling.cpp` after the field data is output. 

This retains the original behavior for stationary samplers, but also provides a flexible way to customize output behavior for lidars and other samplers in future.

tonyinme: Thanks for the suggested workflow @sayerhs. This has been incorporated to the latest version.

sayerhs (approved):

michaeljbrazell (approved):
looks good!

PR #505, opened 2021-11-09
fixing favre_filtering

gdeskos:
- Small fix for favre_filter. Using a growntilebox to compute momentum before applying the filtering operator.
- Also removing pre_mac_projection from the physics class (not needed anymore)


mbkuhn (approved):
Looks like a proper fix for the filter at the boundaries.

PR #506, opened 2021-11-17
Fixing the sloshing tank reg_test files so they don't fail on the nightly tests

gdeskos:


jrood-nrel (approved):

mbkuhn (approved):
These improvements are good, and the expected variation of the surface height remains within the new highest resolution region.

PR #508, opened 2021-12-02
Adding if/else case for Ascent link libraries in CMakeLists

yuya737:
If using amr-wind with Ascent, with the previous setup, i.e. just ` target_link_libraries(${amr_wind_lib_name} PUBLIC ascent::ascent)`, the build process complains in `AMR_WIND_ENABLE_MPI=ON` is set. 

Specifically, it gives a `nvlink error   : Multiple definition` in `libvtkh_filters_mpi.a:GhostStripper.cpp.o` and `libvtkh_filters.a:GhostStripper.cpp.o`

Proposed change resolves this nvlink error


CMakeLists.txt
```diff
@@ -143,7 +143,11 @@ if(AMR_WIND_ENABLE_ASCENT)
   if (Ascent_FOUND)
     message(STATUS "Found Ascent = ${ASCENT_INCLUDE_DIRS}")
   endif()
-  target_link_libraries(${amr_wind_lib_name} PUBLIC ascent::ascent)
+  if (AMR_WIND_ENABLE_MPI)
+    target_link_libraries(${amr_wind_lib_name} PUBLIC ascent::ascent_mpi)
```
sayerhs, line 147: Does `Ascent` provide `COMPONENTS` so that the `find_package` in line 142 can check for `MPI` component?

PR #510, opened 2021-12-07
Liquid surface location sampling

mbkuhn:
Locating the liquid surface using a mesh-based approach, inspired by Shreyas, instead of a particle-based approach.

### DIFF
--- a/amr-wind/utilities/sampling/CMakeLists.txt
+++ b/amr-wind/utilities/sampling/CMakeLists.txt
@@ -10,4 +10,5 @@ target_sources(${amr_wind_lib_name}
   FieldNorms.cpp
   KineticEnergy.cpp
   Enstrophy.cpp
+  FreeSurface.cpp
   )
--- a/amr-wind/utilities/sampling/FreeSurface.H
+++ b/amr-wind/utilities/sampling/FreeSurface.H
@@ -58,7 +58,7 @@ public:
 
 protected:
     //! Output data based on user-defined format
-    void process_output();
+    virtual void process_output();
 
     //! Prepare NetCDF metadata
     void prepare_netcdf_file();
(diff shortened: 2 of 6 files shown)


sayerhs:
@WeiqunZhang This is the PR I had mentioned to you. It would be great if you can take a look at this.

WeiqunZhang:
My thought is this could be done with atomics, which should be much faster than a few thousand GPU kernels.

```
// I am going to assume there is only one MPI process, and worry about MPI                                                                                                                                         
// communication later.                                                                                                                                                                                            
//                                                                                                                                                                                                                 
// I am also going to assume there is only one level.                                                                                                                                                              
//                                                                                                                                                                                                                 
//  I am also assuming the xyplane here is the same as the level 0 xy-plane.                                                                                                                                       

Box xyplane = Geom(0).Domain();
xyplane.setRange(2, 0);

FArrayBox zint_fab(xyplane, 6); // This stores interface heights. Up to 6 interafaces.                                                                                                                             
BaseFab<unsigned int> nint_fab(xyplane); // This stores the number of interfaces.                                                                                                                                  
nint_fab.setVal(0);

Array4<Real> zint = zint_fab.array();
Array4<unsigned int> nint = nint_fab.array();

MultiFab mf(...);
auto const& ma = mf.arrays();

ParallelFor(mf, [=] AMREX_GPU_DEVICE (int box_no, int i, int j, int k) noexcept
{   
    if (there_is_an_interface_in_cell_ijk) {
        // again, this is not necessarily (i,j,0) if the xyplane is not the same as level 0 xy-plane                                                                                                               
        auto idx = Gpu::Atomic::Inc(nint.ptr(i,j,0), 100u); // 100u is just a big enough number                                                                                                                    
        zint(i,j,0,idx) = ... the height ...;
    }
});
```

gdeskos (changes requested):

gdeskos (approved):

Issue #500, opened 2021-11-01
Create a new user tutorial with up-to-date examples 


lawrenceccheung:
Starting an issue to track the creation of a new user tutorial with up-to-date examples.  

The few tasks I see under this issue are:
- Creating a tutorial section in the sphinx documentation
- Choosing which cases to include in the tutorial
- Writing documentation for the cases (obviously)
  - I'm thinking of end-to-end documentation, covering case setup, running, and postprocessing
- Figuring out a way to run the examples periodically, either scaled down as a part of regression test suite or any other method.

Ideally this will be completed before many more AMR-Wind users come in from other projects.

Lawrence

PR #512, opened 2021-12-09
Examples

tonyinme:
This is a draft PR meant to introduce examples. These examples are meant to run production-type simulations. I have been working with @neilmatula @gantech and @lawrenceccheung on this. This PR responds to https://github.com/Exawind/amr-wind/issues/500

Some of the key things to address:

- [x] Add documentation @neilmatula 
- [ ] Establish a way to test the examples regularly (ie, monthly)
- [ ] Add examples of different use cases


tonyinme:
hey @jrood-nrel , is there a way to setup the examples to be tested on Eagle every month? Do you think this would be a good idea?

jrood-nrel:
I think we talked about this. If you want to make them regression tests, that would be best.

PR #534, opened 2021-12-29
Reintroduce and fix more clang-tidy checks

jrood-nrel:
Add in `readability-redundant-member-init` and `modernize-use-bool-literals` checks and fix.


jrood-nrel:
I don't think the original integers into `contains_nan()` where necessary, but this preserves the original result. `CheckForNans` doesn't appear to be used anyway.

PR #547, opened 2022-01-05
Fix bugprone check warnings

jrood-nrel:


jrood-nrel:
Probably should look over these changes carefully. I get nervous mixing unsigned types if they're doing any arithmetic.

amr-wind/overset/TiogaInterface.cpp
```diff
@@ -225,7 +225,7 @@ void TiogaInterface::amr_to_tioga_mesh()
     BL_PROFILE("amr-wind::TiogaInterface::amr_to_tioga_mesh");
     auto& mesh = m_sim.mesh();
     const int nlevels = mesh.finestLevel() + 1;
-    const int iproc = amrex::ParallelDescriptor::MyProc();
+    const long iproc = amrex::ParallelDescriptor::MyProc();
```
michaeljbrazell, line 228: I think MyProc returns an int and dmap uses ints too

amr-wind/overset/TiogaInterface.cpp
```diff
@@ -235,7 +235,7 @@ void TiogaInterface::amr_to_tioga_mesh()
         ngrids_global += mesh.boxArray(lev).size();
 
         const auto& dmap = mesh.DistributionMap(lev);
-        for (long d = 0; d < dmap.size(); ++d) {
+        for (size_t d = 0; d < static_cast<size_t>(dmap.size()); ++d) {
```
michaeljbrazell, line 238: maybe we don't need the static_cast<size_t> ... I think .size() returns that anyways?

amr-wind/utilities/tagging/CurvatureRefinement.H
```diff
@@ -40,7 +40,7 @@ private:
 
     amrex::Vector<amrex::Real> m_curv_value;
 
-    int m_max_lev_field;
+    size_t m_max_lev_field;
```
michaeljbrazell, line 43: I think this should be an int and      
`   size_t fcount = std::min(curv_value.size(), m_curv_value.size());
` 
in CurvatureRefinement.cpp should be fixed instead

amr-wind/utilities/tagging/CurvatureRefinement.cpp
```diff
@@ -43,7 +43,7 @@ void CurvatureRefinement::initialize(const std::string& key)
 void CurvatureRefinement::operator()(
     int level, amrex::TagBoxArray& tags, amrex::Real time, int /*ngrow*/)
 {
-    const bool tag_field = level <= m_max_lev_field;
+    const bool tag_field = static_cast<size_t>(level) <= m_max_lev_field;
```
michaeljbrazell, line 46: this can go away once if m_max_lev_field is an int

amr-wind/wind_energy/ABLStats.cpp
```diff
@@ -228,7 +228,9 @@ void ABLStats::compute_zi(const h1_dir& h1Sel, const h2_dir& h2Sel)
         for (size_t i = 0; i < m_ncells_h1 * m_ncells_h2; i++) {
             m_zi += gtemp_grad[i].max_grad_loc;
         }
-        m_zi /= (m_ncells_h1 * m_ncells_h2);
+        m_zi /=
+            (static_cast<double>(m_ncells_h1) *
+             static_cast<double>(m_ncells_h2));
```
michaeljbrazell, line 233: amrex::Real?

amr-wind/utilities/tagging/FieldRefinement.H
```diff
@@ -44,8 +44,8 @@ private:
     amrex::Vector<amrex::Real> m_field_error;
     amrex::Vector<amrex::Real> m_grad_error;
 
-    int m_max_lev_field;
-    int m_max_lev_grad;
+    size_t m_max_lev_field;
```
michaeljbrazell, line 47: same as above, this just kept getting copied in all the refinement code :(

jrood-nrel:
Thanks for catching all these. I will look at them again and update.

amr-wind/wind_energy/ABLStats.cpp
jrood-nrel: I looked at that but `m_zi` is declared a `double`.

michaeljbrazell: ohhh because of netcdf. well double should be fine. Or we can cast the double to an amrex::Real at the netcdf call?

jrood-nrel: That would probably be better, but there are a lot of explicit `double` around so probably would tackle those later if possible.

amr-wind/overset/TiogaInterface.cpp
jrood-nrel: It actually returns a `long` from what the compiler is telling me. So if `d` stayed as a `long`, it would propagate narrowing conversions to the code within the loop.

amr-wind/wind_energy/ABLStats.cpp
sayerhs: @michaeljbrazell For things that write out to NetCDF interface, I think that we should be explicit about `float/double` instead of using `amrex::Real` This will be necessary to catch issues if an external AMReX was compiled with single precision. For often used NetCDF methods it might be useful to provide a templated version that can be specialized on `amrex::Real` at the call site.

amr-wind/overset/TiogaInterface.cpp
jrood-nrel: So I felt promoting the indexing to `size_t` from `long` was the safest choice?

jrood-nrel: Without these changes:
```
amr-wind/overset/TiogaInterface.cpp:239:22: warning: narrowing conversion from 'long' to signed type 'int' is implementation-defined [bugprone-narrowing-conversions]
            if (dmap[d] == iproc) {
                     ^
```

michaeljbrazell: this won't work? 
`for (size_t d = 0; d < dmap.size(); ++d) {`

jrood-nrel: No
```
amr-wind/overset/TiogaInterface.cpp:238:30: warning: comparison of integers of different signs: 'size_t' (aka 'unsigned long') and 'amrex::Long' (aka 'long') [clang-diagnostic-sign-compare]
        for (size_t d = 0; d < dmap.size(); ++d) {
                             ^
```

jrood-nrel: So if d is `long`, it has to be cast to an `int` in the array index, otherwise, promoting everything to `size_t` works. It's strange to me that `long` appears to not work as an array index, yet `int` and `size_t` (or `unsigned long`) seem to not cause warnings.

jrood-nrel: So right now the code implicitly casts `d` to an `int`. I could just explicitly do that instead of these changes.

jrood-nrel: I'm going to remove these changes from this PR to simplify it.

amr-wind/utilities/tagging/CurvatureRefinement.H
jrood-nrel: Is changing `fcount` to `int` what you meant?

michaeljbrazell: yep, the size of this array will be small so int should be sufficient
`int fcount = static_cast<int> std::min()`

michaeljbrazell (approved):
looks good, thank you for this

PR #548, opened 2022-01-06
Fix or suppress cppcheck warnings

jrood-nrel:


amr-wind/turbulence/LES/OneEqKsgs.cpp
```diff
@@ -12,6 +12,7 @@ namespace amr_wind {
 namespace turbulence {
 
 template <typename Transport>
+// cppcheck-suppress uninitMemberVar
```
sayerhs, line 15: @jrood-nrel  do you remember what member it was complaining about here?

amr-wind/wind_energy/actuator/disk/ActSrcDiskOp.H
```diff
@@ -37,6 +37,7 @@ protected:
     void copy_to_device();
 
 public:
+    // cppcheck-suppress uninitMemberVar
```
sayerhs, line 40: Same question here.

amr-wind/wind_energy/actuator/turbine/fast/FastIface.H
```diff
@@ -67,13 +67,13 @@ protected:
...
     double m_start_time{0.0};
-    double m_stop_time;
+    double m_stop_time{-1.0};
     SimMode m_sim_mode{SimMode::init};
 
     int m_num_sc_inputs{0};
     int m_num_sc_outputs{0};
 
-    char m_errmsg[fast_strlen()];
+    // char m_errmsg[fast_strlen()];
```
sayerhs, line 76: This line can be removed entirely. Superseded by https://github.com/Exawind/amr-wind/blob/dc887e210c9ad99e2fd7225a6d91b9826c350d72/amr-wind/wind_energy/actuator/turbine/fast/FastIface.cpp#L18

amr-wind/turbulence/LES/OneEqKsgs.cpp
jrood-nrel: Sure. Here is the list I was working on: https://github.com/Exawind/amr-wind/runs/4717856939?check_suite_focus=true

amr-wind/wind_energy/actuator/turbine/fast/FastIface.H
jrood-nrel: Thanks. Wasn't sure why it was there.

amr-wind/turbulence/LES/OneEqKsgs.cpp
sayerhs: Interesting, it complains about members that seem to be initialized:

Screen Shot 2022-01-06 at 1 42 19 PM

https://github.com/Exawind/amr-wind/blob/dc887e210c9ad99e2fd7225a6d91b9826c350d72/amr-wind/turbulence/LES/OneEqKsgs.cpp#L14-L26

jrood-nrel: Yeah it seems to get false positives when the initialization is complex.

amr-wind/physics/HybridRANSLESABL.cpp
```diff
@@ -7,6 +7,7 @@
 namespace amr_wind {
 namespace hybrid_rans_les_abl {
 
+// cppcheck-suppress uninitMemberVar
 HybridRANSLESABL::HybridRANSLESABL(const CFDSim& sim) : m_sim(sim) {}
```
michaeljbrazell, line 11: is it complaining that m_tke and m_sdr are not initialized? we could set them to nullptr in the header?

amr-wind/physics/HybridRANSLESABL.cpp
```diff
@@ -7,6 +7,7 @@
 namespace amr_wind {
 namespace hybrid_rans_les_abl {
 
+// cppcheck-suppress uninitMemberVar
```
michaeljbrazell, line 10: is it complaining that m_tke and m_sdr are not initialized? we could set them to nullptr in the header?

jrood-nrel: Yeah I'm pretty sure it was complaining about those. Anything we can do to not have to use suppressions is good.

sayerhs: Can you guys explain your thought process? I ask because it is also complaining for smagorinsky and that class doesn't have any pointer instance members. 

Of course, defaulting the pointers for tke/SDR in this class should be done regardless of the warnings.

sayerhs: While we are at it, we should also fix usages like this in Turbulence models: https://github.com/Exawind/amr-wind/blob/5525d75745b1f2a9a0a4da9504a818f8db662f77/amr-wind/turbulence/LES/Smagorinsky.cpp#L19

I think I am making a mistake here and it won't work as intended. It is better to be explicit here.

jrood-nrel: Yeah those need to be fixed for us to get to 0 clang-tidy warnings. That's all that's left.

PR #550, opened 2022-01-12
fixed a few more cppcheck and clang tidy warnings

michaeljbrazell:


amr-wind/core/Field.cpp
```diff
@@ -28,6 +28,7 @@ FieldInfo::FieldInfo(
 {
     for (int i = 0; i < AMREX_SPACEDIM * 2; ++i) {
         m_bc_type[i] = BC::undefined;
+        m_bc_values_d[i] = nullptr;
```
sayerhs, line 31: Would it be possible to do this in constructor initializer list?

sayerhs (approved):

jrood-nrel (approved):

amr-wind/core/Field.cpp
michaeljbrazell: like this?
`    , m_bc_values(AMREX_SPACEDIM * 2, amrex::Vector<amrex::Real>(ncomp, 0.0))
    , m_bc_values_d({})
    , m_bc_values_dview(ncomp * AMREX_SPACEDIM * 2)`

sayerhs:
```
, m_bc_values_d(nullptr)
``` 

Since this is an AMReX data structure. I was curious if it had this option.

michaeljbrazell: it doesn't have that option :(

sayerhs: Thanks for clearing that up.

amr-wind/turbulence/LES/OneEqKsgs.cpp
```diff
@@ -41,7 +41,7 @@ OneEqKsgsM84<Transport>::OneEqKsgsM84(CFDSim& sim)
     }
 
     {
-        const std::string coeffs_dict = this->model_name() + "_coeffs";
+        const std::string coeffs_dict = model_name() + "_coeffs";
```
sayerhs, line 44: I doubt this will fix the issue... as the method is still declared virtual.

michaeljbrazell: darn, I need to set up clang-tidy on my machine instead of depending on CI

jrood-nrel: You can use it on eagle with `module load llvm` or rhodes when eagle is down.

PR #553, opened 2022-01-19
First cut at examples read-the-docs page.  Adding inflow-outflow example.

neilmatula:


tonyinme:
hey @neilmatula , the documentation is good. I was thinking it might be worth adding towards the end the commands used to run the code (including the number of processors recommended).

neilmatula:
@tonyinme Sounds good.  I'll add that.

lawrenceccheung (commented):
Hi @neilmatula, this is great.  I really like the schematic showing the different stages of the inflow/outflow run.

I was wondering if it's possible to show snippets of the log file during a successful run.  Maybe something like the start of the log file, the first iteration and the end of the log file so people know what it looks like when amr-wind is running properly.  It doesn't have to be for this example case, could be for the other ones too.

Thanks,

Lawrence

neilmatula:
@lawrenceccheung 
Yep.  I'll add that.

tonyinme (approved):
This looks good.
Is anything missing?
Or can it be merged?

jrood-nrel:
This references and `examples` directory that doesn't exist.

neilmatula:
Pinging for activity.

marchdf:
@neilmatula what are your thoughts on pushing this through? Do you need a hand?

neilmatula:
@marchdf With a bit more work it can be pushed through.
Thanks, but I think I can do it myself soon.

One remaining concern is ABLMeanBoussinesq.
My current thinking is that it needs to be turned on for both precursor and inflow-outflow.
If it's just turned on for the latter, strange effects can be observed in both pressure and the mean velocities.
I'd like to confirm that having it on for both is what we should be doing.  Also, if that's the case, can we guard against someone accidentally not doing that?

lawrenceccheung:
In one of my last conversations with Mike B., we mentioned that the ultimate solution is to have a better outflow BC which doesn't rely on `ABLMeanBoussinesq` being turned on.  (Ideally we'd do this with other planned MMC coupling type work).  However, in the absence of that, we should turn on `ABLMeanBoussinesq` in precursors because people are likely to forget about it in their normal workflows.

Lawrence

PR #557, opened 2022-01-26
native bndry I/O

michaeljbrazell:
- This is a work in progress but should allow for some basic testing. Ashesh's large ABL case outputs planes now
- branch passes regression tests (abl_bndry_output and abl_bndry_input) when comparing to plotfiles generated using the original netcdf pathway
- fixme: abl_bndry_input_amr regression does not work currently, need to generalize lev > 0
- fixme: copying multifab to fabs on all procs is not clean as it relies on finding the first distributation map proc


ashesh2512 (approved):

michaeljbrazell:
please test this PR and provide feedback, it's definitely not ready to merge

jrood-nrel:
Let's make it a draft then.

amr-wind/wind_energy/ABLBoundaryPlane.H
```diff
@@ -176,16 +194,28 @@ private:
...
     //! Flag indicating if this capability is available
     bool m_is_initialized{false};
+
+    //! controls extents on native bndry output
+    const int m_in_rad = 1;
+    const int m_out_rad = 1;
+    const int m_extent_rad = 0;
+
+    //! output format for bndry output
+#ifdef AMR_WIND_USE_NETCDF
```
marchdf, line 214: so if we turn on netcdf, then by default we would use netcdf output But if we find that the native is much faster, then we should probably default to `native`, right? Even if netcdf is turned on, e.g. for other reasons

amr-wind/wind_energy/ABLBoundaryPlane.cpp
```diff
@@ -186,19 +271,32 @@ ABLBoundaryPlane::ABLBoundaryPlane(CFDSim& sim)
...
     pp.queryarr("bndry_planes", m_planes);
     pp.query("bndry_output_start_time", m_out_start_time);
     pp.queryarr("bndry_var_names", m_var_names);
     pp.get("bndry_file", m_filename);
+    pp.query("bndry_output_format", m_out_fmt);
 
+#ifndef AMR_WIND_USE_NETCDF
+    if (m_out_fmt == "netcdf") {
+        amrex::Print()
+            << "Error: boundary output format using netcdf must link netcdf "
```
marchdf, line 284: I would expect an "error" to quit the program. Can we change to "warning"?

amr-wind/wind_energy/ABLBoundaryPlane.cpp
```diff
@@ -186,19 +271,32 @@ ABLBoundaryPlane::ABLBoundaryPlane(CFDSim& sim)
...
+               "library, changing output to native format"
+            << std::endl;
+        m_out_fmt = "native";
+    }
 #endif
+
+    if (!(m_out_fmt == "native" || m_out_fmt == "netcdf")) {
+        amrex::Print() << "Error: boundary output format not recognized, "
```
marchdf, line 292: same comment here.

amr-wind/wind_energy/ABLBoundaryPlane.cpp
michaeljbrazell: good call, I'll change it

amr-wind/wind_energy/ABLBoundaryPlane.H
michaeljbrazell: yep that works for me, I can change the current reg tests to use bndry_output_format=netcdf. I should probably add new ones too for native

michaeljbrazell:
- I tested GPUs and they match CPU runs
- cleaned up the multifab->fab copy using copyTo (Weiqun has a PR in amrex to make that more efficient, need to upgrade amrex)
- multilevel works now but only if lev>0 exist in the middle of the domain, refinement on inflows will not work

jrood-nrel (changes requested):
Can you fix the new clang-tidy errors here?

michaeljbrazell:
yikes I added a lot of clang tidy warnings! fixed 'em

jrood-nrel:
Can you rebase on `main`? Clang-tidy checks should pass for PRs now.

jrood-nrel (approved):

ashesh2512:
@michaeljbrazell i think we are missing `abl_bndry_input_amr_native.i`

PR #558, opened 2022-01-30
Fix FPE when CFLs are zero

marchdf:
It is possible to have CFLs computed to be zero, which leads to an FPE when computing `dt_new`. This fixes that potential FPE.

This FPE arose in several test cases, including BoussinesqBubble. The flow is initialized to zero and all CFLs are computed to be zero. This throws an FPE. 

Tests that throw an FPE and are resolved with this PR:
```
	  4 - boussinesq_bubble_godunov (Failed)
	 24 - boussinesq_bubble_mol (Failed)
	 30 - rayleigh_taylor_godunov (Failed)
	 31 - rayleigh_taylor_mol (Failed)
	 36 - dam_break_godunov (Failed)
```


jrood-nrel (approved):

michaeljbrazell (approved):
looks good, passes reg tests on my end

PR #560, opened 2022-01-30
abl_virtual_lidar test should only run with netcdf is enabled

marchdf:
This test uses netcdf output for sampling, therefore it should only
run when netcdf is enabled.


tonyinme (approved):
good catch!

jrood-nrel (approved):

PR #566, opened 2022-02-03
Linking to MOL in amrex-hydro

marchdf:
Draft for now, Jon and I are finding that it is fixing FPEs in intel.


michaeljbrazell (approved):

marchdf:
Yeah and I think there's a couple more things that we need to check here... I would like to replace the flux and div calcs with the amrex-hydro ones. Need to be very careful about the boxes...

jrood-nrel:
I just tested this again since I was looking at FPEs and this solves almost all of the MOL-related FPEs that Intel is showing otherwise. I think this fixes a bug in MOL.

jrood-nrel:
Would like to fix the nightly tests by merging this.

PR #570, opened 2022-02-10
Update OpenFAST interface to work with OF-3.0

ndevelder:
This makes two changes to allow AMR-Wind build with OpenFAST 3.0+:

1. Add _DX_ to data type names -> SC_InputType and SC_OutputType
2. Align "FAST_OpFM_Init" with definition in Fast_Library.h

CRITICAL: without any other changes, this will break builds with OF-2.6 and lower

To remedy this, does anyone have any good ideas about passing OF version to AMR-Wind to split the fast_func into version specific definitions? Or is there a more fancy/elegant solution that doesn't involve version specific definitions?

### DIFF
--- a/amr-wind/wind_energy/actuator/turbine/fast/FastIface.H
+++ b/amr-wind/wind_energy/actuator/turbine/fast/FastIface.H
@@ -72,6 +72,9 @@ protected:
 
     int m_num_sc_inputs{0};
     int m_num_sc_outputs{0};
+    int m_num_sc_inputs_glob{0};
+    float m_init_sc_inputs_glob{0.0}; 
+    float m_init_sc_inputs_turbine{0.0};
 
     bool m_is_initialized{false};
 };
--- a/amr-wind/wind_energy/actuator/turbine/fast/FastIface.cpp
+++ b/amr-wind/wind_energy/actuator/turbine/fast/FastIface.cpp
@@ -201,9 +201,10 @@ void FastIface::fast_init_turbine(FastTurbine& fi)
     char inp_file[fast_strlen()];
     copy_filename(fi.input_file, inp_file);
 
-    fast_func(
+    fast_func( 
         FAST_OpFM_Init, &fi.tid_local, &fi.stop_time, inp_file, &fi.tid_global,
-        &m_num_sc_inputs, &m_num_sc_outputs, &fi.num_pts_blade,
+        &m_num_sc_inputs_glob, &m_num_sc_inputs, &m_num_sc_outputs, 
+        &m_init_sc_inputs_glob, &m_init_sc_inputs_turbine, &fi.num_pts_blade,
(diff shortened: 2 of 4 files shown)


sayerhs:
I don't think there is an easy way to avoid version specific definitions, as the type signatures have changed. 

Instead of checking for versions, I would recommend using [check_include_file](https://cmake.org/cmake/help/latest/module/CheckIncludeFile.html) to check for the availability of `SCDataEx_Types.h` header file. After L140 in the CMake configuration phase: https://github.com/Exawind/amr-wind/blob/85a7be5493dd7cfcfdc42ebd8dc22424dc0117b8/CMakeLists.txt#L135-L144 

```CMake
include(CheckIncludeFile)
set(CMAKE_REQUIRED_INCLUDES "${OpenFAST_INCLUDE_DIRS}")
check_include_file("SCDataEx_Types.h" AMR_WIND_FAST_USE_SCDX)
```

And then use this variable to do your ifdefs.

ndevelder:
Shreyas' comment has been implemented. I had to use "add_definitions" in CMakeLists to get it to work. Will test with both openfast and non-openfast builds, then move out of draft status.

CMakeLists.txt
```diff
@@ -138,6 +138,13 @@ if (AMR_WIND_ENABLE_OPENFAST)
   if(OpenFAST_FOUND)
     message(STATUS "Found OpenFAST = ${OPENFAST_DIR}")
   endif()
+  include(CheckIncludeFileCXX)
+  set(CMAKE_REQUIRED_INCLUDES "${OpenFAST_INCLUDE_DIRS}")
+  check_include_file_cxx("SCDataEx_Types.h" USE_SCDX)
+  message(STATUS "AMR_WIND_FAST_USE_SCDX is: ${AMR_WIND_FAST_USE_SCDX}")
+  if(USE_SCDX)
+    add_definitions(-DAMR_WIND_FAST_USE_SCDX=1)
```
sayerhs, line 146:
```suggestion
    target_compile_definitions(${amr_wind_lib_name} PUBLIC AMR_WIND_FAST_USE_SCDX)
```

To be consistent with modern CMake usage (see L148). This will export the compile definitions as part of the AMR-Wind target if someone wants to use AMR-Wind as a library with other codes.

ndevelder:
No problem! Thanks for the suggestions, and it's all an opportunity to learn a bit more!

ndevelder:
I just built and ran cleanly with both OpenFAST 2.6 and OpenFAST 3.0. Moving this into review status. I used intel compilers, do we want someone to test this in any gcc environments too?

psakievich (approved):
@ndevelder I wouldn't worry about testing on gcc before merging. If it compiles and runs with intel it will compile with gcc, but compile times with GCC are very slow. Any issues beyond compiling may need to be taken up with the openfast team. Let's get this merged and then do more testing.

michaeljbrazell (approved):

PR #573, opened 2022-02-16
Mass-Momentum consistent transport for multiphase

mbkuhn:
implements a consistent approach between VOF transport and velocity, ICNS eq is based on momentum instead.
not expected to merge yet, but wanting to keep track of changes better.


amr-wind/equation_systems/SchemeTraits.H
```diff
@@ -17,11 +17,11 @@ struct Godunov
     //! Number of states in the field variable
     static constexpr int num_states = 2;
     //! Number of ghost in the state variable
-    static constexpr int nghost_state = 3;
+    static constexpr int nghost_state = 5;
```
gdeskos, line 20: Just do this in locally for VOF

gdeskos:
```
struct VOF : ScalarTransport
{
    using MLDiffOp = amrex::MLABecLaplacian;
    using SrcTerm = SourceTerm;

    static std::string pde_name() { return "VOF"; }
    static std::string var_name() { return "vof"; }

    static constexpr int ndim = 1;
    static constexpr bool multiply_rho = false;
    static constexpr bool has_diffusion = false;
    static constexpr bool need_nph_state = false;

    static constexpr int nghost_state = 5;

    static constexpr amrex::Real default_bc_value = 0.0;
};

```

amr-wind/equation_systems/vof/vof_advection.H
```diff
@@ -59,16 +62,59 @@ struct AdvectionOp<VOF, fvm::Godunov>
...
-                    u_mac(lev).const_array(mfi), v_mac(lev).const_array(mfi),
-                    w_mac(lev).const_array(mfi), aa_x(lev).array(mfi),
-                    aa_y(lev).array(mfi), aa_z(lev).array(mfi),
-                    dof_field.bcrec_device().data(), tmpfab.dataPtr(), geom, dt,
+
+                multiphase::cmask_loop(
+                    bx, dof_field(lev).array(mfi), (*fluxC)(lev).array(mfi),
                     m_use_lagrangian);
 
+                multiphase::split_advection_step(
```
gdeskos, line 72: I am sure this can be done in a more efficient (and prettier) way, but I am happy for time being. I guess we can refactor later on

gdeskos (approved):

PR #575, opened 2022-02-18
Update virtual lidar to use coordinate transformations and add period…

tonyinme:
This adds the ability to have a scan pattern that repeats itself in time.

The computation on the virtual lidar locations has been changed from vector transformations to a coordinate system transformation. This computation is more general, fixes a bug and simplifies the computation.

### DIFF
--- a/amr-wind/utilities/sampling/LidarSampler.H
+++ b/amr-wind/utilities/sampling/LidarSampler.H
@@ -47,7 +47,7 @@ protected:
     amrex::Vector<amrex::Real> m_elevation_table;
     amrex::Real m_length{0};
     bool m_periodic{true};
-    amrex::Real m_period{99999999};
+    amrex::Real m_period{std::numeric_limits<double>::max()};
 };
 
 } // namespace sampling
--- a/amr-wind/utilities/sampling/LidarSampler.H
+++ b/amr-wind/utilities/sampling/LidarSampler.H
@@ -46,6 +46,8 @@ protected:
     amrex::Vector<amrex::Real> m_azimuth_table;
     amrex::Vector<amrex::Real> m_elevation_table;
     amrex::Real m_length{0};
+    bool m_periodic{true};
+    amrex::Real m_period{99999999};
 };
 
 } // namespace sampling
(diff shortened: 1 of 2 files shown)


ndevelder:
Hi Tony! This generally looks good. I might be missing it, or misunderstanding how it works, but do lines 80-84 in LidarSampler.cpp assume a directionality to the cartesian coordinate system...i.e. the 2 component is the cos of elevation...does this mean we can only use this in cases where "z" is the "up" direction? If so, we may want to consider some user input to define this directionality, or try and intuit it from the boundary conditions?

amr-wind/utilities/sampling/LidarSampler.H
```diff
@@ -46,6 +46,8 @@ protected:
     amrex::Vector<amrex::Real> m_azimuth_table;
     amrex::Vector<amrex::Real> m_elevation_table;
     amrex::Real m_length{0};
+    bool m_periodic{true};
+    amrex::Real m_period{99999999};
```
michaeljbrazell, line 50: use a std max numeric limits please

amr-wind/utilities/sampling/LidarSampler.cpp
```diff
@@ -55,32 +65,23 @@ void LidarSampler::update_sampling_locations()
...
     amrex::Real time = m_sim.time().current_time();
 
-    // The current azimuth angle
-    const amrex::Real current_azimuth =
-        ::amr_wind::interp::linear(m_time_table, m_azimuth_table, time);
-
-    const amrex::Real current_elevation =
-        ::amr_wind::interp::linear(m_time_table, m_elevation_table, time);
+    // If the time table is periodic, make sure to get the correct time
+    if (m_periodic) time = std::fmod(time, m_period);
```
michaeljbrazell, line 69: add braces here please

tonyinme: fixed

amr-wind/utilities/sampling/LidarSampler.H
tonyinme: fixed

tonyinme:
Thanks @michaeljbrazell , the fixes have been implemented.

michaeljbrazell (approved):

PR #576, opened 2022-02-21
one-line fix to PR #571

mbkuhn:
uncommenting fillpatch for density, necessary for interpolating density in MAC projection


gdeskos (approved):

gdeskos:
Trivial one-line fix that does not affect anything else other than the multiphase stuff

PR #577, opened 2022-02-24
Spinnerlidar

alhs6577:
First draft of spinnerlidar model into amr-wind for the RAAW Project


amr-wind/utilities/sampling/DTUSpinnerSampler.cpp
```diff
@@ -109,16 +199,11 @@ void DTUSpinnerSampler::update_sampling_locations()
     if (time == start_time) {
         m_time_sampling = time;
     }
-    amrex::Real time_tmp = m_time_sampling;
-    bool cond = true;
-    constexpr double eps = 1.0e-12;
-    amrex::Real time_new = time + dt_sim;
     // Loop to see how many times we will subsample
```
tonyinme, line 202: update comment

tonyinme (changes requested):
Excellent job writing this code @alhs6577 !

Can you please adjust the inputs for this test to make sure it works with this new implementation.
https://github.com/Exawind/amr-wind/blob/main/test/test_files/abl_virtual_lidar/abl_virtual_lidar.i

amr-wind/utilities/sampling/DTUSpinnerSampler.cpp
alhs6577: Whoops, forgot that. The updated test file should be added now.

tonyinme (approved):

tonyinme:
hey @michaeljbrazell , it looks like this is ready to merge.
@alhs6577 , can you confirm?

alhs6577:
> hey @michaeljbrazell , it looks like this is ready to merge. @alhs6577 , can you confirm?

I'm actually not sure since Nate recently added a couple fast interface function changes to the pull request. Since those haven't been fully tested, there are potentially issues there so we might want to hold off for now.

tonyinme:
> 

oh, yes, I missed the new commits. Thanks for clarifying @alhs6577 !

ndevelder:
Hi @all, This is building and running fine for me on RHEL7 with intel 2019 compiler....but seems to be failing on build tests. Guessing I'm doing something wrong, but having trouble figuring out what. Any thoughts?

ndevelder:
Ok. Added a bunch of idefs for openfast so that all the CI builds would go through. There are a few warnings in Lint-cppcheck that don't have to do with our spinner code. How much more V&V do we want to do before merging this?

tonyinme:
Hi @ndevelder and @alhs6577 , I thought this had been merged a while ago. 
Is this ready to be merged?

ndevelder:
Hi @tonyinme I've been meaning to get back to this, and will try to finalize this afternoon.

ndevelder:
@tonyinme and @alhs6577, this is in the state it was prior to being out of sync with main. I think there could be some more work in the following areas:
- Tests aren't in a great state...but I'm not sure we need to put more work in to merge. 
- Not sure if we need to do anything GPU related...and the cuda test seemingly fails in cmake?
- Documentation and an example case would be nice

ndevelder:
Note, Tommy found that DTU changed the spinner rotational direction, so this PR is on hold until we switch direction and do another verification exercise.

ndevelder:
This work is being continued in https://github.com/Exawind/amr-wind/pull/828

PR #581, opened 2022-03-08
Update .gitignore

psakievich:
Add .cache to .gitignore.  This has been driving me crazy

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@ spack*
 compile_commands.json
 settings.json
 .vscode*
+.cache*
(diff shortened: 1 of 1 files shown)


jrood-nrel (approved):

marchdf (approved):

marchdf:
Feels like we almost should be adding even more stuff in there...

.gitignore
```diff
@@ -18,3 +18,4 @@ spack*
 compile_commands.json
 settings.json
 .vscode*
+.cache*
```
marchdf, line 21: Feels like we almost should be adding even more stuff in there... I would like
```
.ccls
.ccls-cache/
build/
```
to be there too

PR #583, opened 2022-03-09
Implement Joukoswky Acutator Disk

psakievich:
This PR pulls out the operations that can be shared between the UniformCt disk and the pending Joukowski disk. 

The new input will be `Actuator.JoukowskiDisk` and should accept all the same inputs.  I think this PR is good to merge on it's own, but I can convert it to a draft and do the development of the Joukowski disk in this PR if people desire.


amr-wind/wind_energy/actuator/disk/disk_ops.cpp
```diff
@@ -384,6 +376,7 @@ void compute_disk_points(
         }
     }
 }
+} // namespace base
 } // namespace ops
 } // namespace actuator
-} // namespace amr_wind
+} // namespace amr_wind
```
tonyinme, line 382: This shows a warning about there not being a line at the end of the file. Not sure if this matters.

amr-wind/wind_energy/actuator/disk/disk_ops.cpp
```diff
@@ -126,7 +127,7 @@ void collect_parse_conflicts(
     std::ostringstream& ss)
 {
     if (pp.contains(p1) && pp.contains(p2)) {
-        ss << "UniformCt Conflict: " << p1 << " and " << p2 << std::endl;
+        ss << "ActuatorDisk Conflict: " << p1 << " and " << p2 << std::endl;
```
tonyinme, line 130: I'm wondering if it might be best to use an identifier in here instead of spelling "ActuatorDisk" all the time?

tonyinme (commented):
This looks good. I only had a few minor comments.

amr-wind/wind_energy/actuator/disk/disk_ops.cpp
psakievich: Sure. We can add an argument for the identifier. This is a free function so it has no idea what the identifier is for the object calling it which is why I generalized it, but it would make things cleaner to give a specific identifier as you suggest.

psakievich:
@lawrenceccheung @ndevelder @neilmatula this is ready for beginning verification runs.  I've implemented all the infrastructure for the Joukowsky disk from the Sorenson 2020 paper. You can see the spatial variation in the source term in the image below:
image

psakievich:
@michaeljbrazell if you can start looking at this soon that would be great. I think there is a little clean up still, but it should be pretty close to done.  I don't want to add any more features at this point due to the size of the PR.

michaeljbrazell (approved):
looks good, tested on GPUs and CPUs and gives the same result. Will cause a few reg test diffs for actuator lines

PR #585, opened 2022-03-11
Add braces to headers

jrood-nrel:


gdeskos:
It all looks good to me. I only looked at VOF and WENO parts though

gdeskos (commented):
It all looks good

PR #604, opened 2022-03-30
Prescribe advection velocity

mbkuhn:
PR adds ability to directly prescribe the velocity field at the faces. This is very important for certain multiphase test cases, allowing more accurate specification of the velocity without contamination from unnecessary solves and interpolations. It also makes these cases much faster to run.


michaeljbrazell (approved):
looks good, could clean up some comments in ApplyPrescribeStep if you feel like it

amr-wind/incflo_advance.cpp
```diff
@@ -566,3 +566,151 @@ void incflo::ApplyCorrector()
...
+    // - No source terms or viscous terms are used for icns
+    // - The MAC velocity is prescribed before this routine, not calculated here
+    // - The nodal projection is omitted
+
+    if (m_verbose > 2) {
+        PrintMaxValues("before prescribe step");
+    }
+
+    if (m_use_godunov) {
+        amr_wind::io::print_mlmg_header("Godunov:");
```
michaeljbrazell, line 593: probably don't need this output either

gdeskos (approved):

Issue #611, opened 2022-04-22
First timestep for boundary planes varies with data format


rybchuk:
I ran into unexpected behavior when launching a turbine simulation using boundary plane data from a precursor simulation. I ran two more precursor simulations to investigate this behavior.

Let's say my precursor simulation has `time.fixed_dt = 1.0` and I set `ABL.bndry_output_start_time = 8999.0`. If I set the boundary output format to be netCDF `ABL.bndry_output_format = netcdf`, then the first value of `time` in `bndry_file.nc` is `8999`. If I set the boundary output format to be the default native format, then the first value of time in `time.dat` is `9000`. I've attached the two input files I used to test this behavior.

[abl_precursor_native.txt](https://github.com/Exawind/amr-wind/files/8542141/abl_precursor_native.txt)

[abl_precursor_netcdf.txt](https://github.com/Exawind/amr-wind/files/8542146/abl_precursor_netcdf.txt)

tonyinme:
Hey @rybchuk , thanks for pointing this out. 

@michaeljbrazell , I am surprised that this does not show up in the [test](https://github.com/Exawind/amr-wind/blob/main/test/test_files/abl_bndry_input_native/abl_bndry_input_native.i).
Do you have any ideas?

rybchuk:
@tonyinme, @michaeljbrazell helped me look into this issue. Turns out the tests were passing because `abl_bndry_output_native.i` started outputting data at `2.0` seconds (which is the 4th timestep), but `abl_bndry_input_native.i` used a restart file that started at the 5th timestep, which circumvents this issue. I'll submit a PR in just a second to resolve this.

PR #613, opened 2022-05-03
Enable native boundary plane restart for first timestep

rybchuk:
This PR addresses #611. Thanks to @michaeljbrazell for the solution.

The jist of the issue was as follows: if you generated a precursor boundary plane that started at a given time (e.g. 2 seconds) using the native data format, your restart simulation would need to start one timestep later. This was a little funny, because that boundary plane data *was* being generated. But this bug was happening because the `time.dat` file was missing the first timestep.

I modified two files in this PR:
* `amr-wind/test/test_files/abl_bndry_input_amr/abl_bndry_input_native.i`: I change the restart time to start at 2 seconds (which is the first `abl_bndry_output_native` timestep for the boundary planes) instead of 2.5 seconds, which will now cause the regression test to fail
* `amr-wind/wind_energy/ABLBoundaryPlane.cpp`: I explicitly create the `bndry_files` directory where the time file is generated, that way the first `time.dat` timestep is written in time. *This also has a side effect:* If you run a precursor simulation and then re-run that same simulations, the initial `bndry_files` directory will be renamed `bndry_files.old*`.


michaeljbrazell:
Thanks for submitting this PR. It should cause a failure in the reg tests, once that happens can you let @jrood-nrel know and he can update the golds. Thanks!

tonyinme (approved):

tonyinme:
This looks good. I think it's ready to merge.

PR #616, opened 2022-05-05
Add ZFP compression support for HDF5 plot files

jrood-nrel:
Depends on https://github.com/AMReX-Codes/amrex/pull/2753


amr-wind/utilities/IOManager.cpp
```diff
@@ -159,9 +162,14 @@ void IOManager::write_plot_file()
...
 #ifdef AMR_WIND_USE_HDF5
     if (m_output_hdf5_plotfile) {
-        amrex::WriteMultiLevelPlotfileHDF5(
+        amrex::WriteMultiLevelPlotfileHDF5SingleDset(
             plt_filename, nlevels, outfield->vec_const_ptrs(), m_plt_var_names,
-            mesh.Geom(), m_sim.time().new_time(), istep, mesh.refRatio());
+            mesh.Geom(), m_sim.time().new_time(), istep, mesh.refRatio()
+#ifdef AMR_WIND_USE_HDF5_ZFP
+                                                             ,
+            m_hdf5_compression
```
sayerhs, line 170:
```suggestion
            , m_hdf5_compression
```

jrood-nrel: `clang-format` does this unfortunately. We would need to use a no-format block.

sayerhs: Seems to me that the merged code will fail compilation with zfp ifdef activated?

PR #617, opened 2022-05-05
Unstable inflow outflow test case

rybchuk:
With @tonyinme 's help, I've added a test case that gives some guidance on how to run an unstable inflow/outflow simulation. As part of this, I've also added a test case that is the precursor for the inflow/outflow simulation, as well as a script by @gantech to help take data from the precursor simulation and feed it into the inflow/outflow input file.

I ran into some `ctest` behavior I don't understand, but I think I'm missing something obvious. If I manually run the inflow/outflow simulation, it successfully runs. However, if I run it through `ctest` I get errors. `Backtrace.0` reads

```
=== If no file names and line numbers are shown below, one can run
            addr2line -Cpfie my_exefile my_line_address
    to convert `my_line_address` (e.g., 0x4a6b) into file name and line number.
    Or one can use amrex/Tools/Backtrace/parse_bt.py.

=== Please note that the line number reported by addr2line may not be accurate.
    One can use
            readelf -wl my_exefile | grep my_line_address'
    to find out the offset for that line.

 0: /scratch/orybchuk/tmp/amr-wind/build/submods/amrex/Src/libamrex.so(_ZN5amrex11BLBackTrace20print_backtrace_infoEP8_IO_FILE+0x31) [0x2ac41dc63f71]
    ?? ??:0

 1: /scratch/orybchuk/tmp/amr-wind/build/submods/amrex/Src/libamrex.so(_ZN5amrex11BLBackTrace7handlerEi+0x335) [0x2ac41dc65e45]
    ?? ??:0

 2: /lib64/libc.so.6(+0x363b0) [0x2ac41edf53b0]

 3: /scratch/orybchuk/tmp/amr-wind/build/amr_wind() [0x5e255f]
    amr_wind::MOData::update_fluxes(int) at ??:?

 4: /scratch/orybchuk/tmp/amr-wind/build/amr_wind() [0x5ad0a0]
    amr_wind::ABL::post_init_actions() at ??:?

 5: /scratch/orybchuk/tmp/amr-wind/build/amr_wind() [0x441c25]
    incflo::init_amr_wind_modules() at ??:?

 6: /scratch/orybchuk/tmp/amr-wind/build/amr_wind() [0x443b41]
    incflo::InitData() at ??:?

 7: /scratch/orybchuk/tmp/amr-wind/build/amr_wind() [0x436154]
    main at ??:?

 8: /lib64/libc.so.6(__libc_start_main+0xf5) [0x2ac41ede1505]

 9: /scratch/orybchuk/tmp/amr-wind/build/amr_wind() [0x43e6bb]
    _start at ??:?
```

Does this log file point to an obvious bug that I'm missing? Thanks!


sayerhs:
@rybchuk currently testing in AMR-Wind enables FPE trapping (https://github.com/Exawind/amr-wind/blob/9e008bb69837e7f02912ef191e138996d3280b5f/CMakeLists.txt#L24) so there's some división by zero (for example) in the Flux computations that us getting caught in ctest, but not in regular runs.

rybchuk:
Hi @sayerhs, thanks for the input. I'm running this test on Eagle with gcc. Tony and I tried a few things that might be messing with initialization: switching from M84 to the Smagorinsky turbulence model, no longer using the `tke` boundary plane, turning off `BodyForce` and ``ABLMeanBoussinesq`, and starting from a different checkpoint timestep. The issue still persisted.

I'm starting to realize that this test case is pretty similar to the neutral input/output one. I'll cross reference the two input files and let you know if I find anything.

rybchuk:
I made some edits and I'm now passing the regression tests!
* The main issue was I had a typo in the input file
* I'll also note something weird: I initially had my `stop_time` set to some large value but my `max_step` set to some small value. However, when the simulation reached the `max_step`, it blew past it and continued on to trying to reach the `stop_time`. I fixed this issue in my test by setting the `stop_time` to line up with the `max_step`. I would be surprised if any of my code modifications brought about this behavior, so I think problem stems from a different part of the code.

Issue #619, opened 2022-05-09
HIP and CUDA backends failing unit test


afanfa:
Hello,
when using either the HIP or CUDA backend, the first unit test fails due to an illegal memory access; likely a CPU pointer is being read on the GPU or vice versa.
The problem started with commit: 9e008bb69837e7f02912ef191e138996d3280b5f . 
Is there a plan to fix this anytime soon?

jrood-nrel:
Thanks for reporting this. It has showed up in our tests as well. https://my.cdash.org/test/55248308

PR #621, opened 2022-05-09
Fix global variable read on GPU error

jrood-nrel:
This solves the global variable read error on device from #580 . Member variables are global variables to the device and need to be captured locally to propagate to the device during kernel launch.


jrood-nrel:
Closes https://github.com/Exawind/amr-wind/issues/619

mbkuhn (approved):
Thanks for fixing this so quickly, Jon. Looks good.

PR #623, opened 2022-05-12
Segregated momentum solve functionality

sbidadi9:
The purpose of this PR is to

1. Utilize nsolve (for overset simulations) which currently works only for scalar solves which is enabled by segregated solve.
2. Utilize Hypre linear solver which currently works only for scalar solves which is enabled by segregated solve.
3. New _MLABecCecLaplacian_ operator which treats convective term implicitly has been developed only for scalar equations.

Highlights:

* Segregated momentum solve algorithm solves u, v and w momentum equations separately. 
  The original PR can be found here: https://github.com/Exawind/amr-wind/commit/7ee3a5a7a0794022cdf092589fd4393d7bd215f4

* Feature can be enabled by setting,
   `velocity_diffusion.use_tensor_operator = false`, and
   `velocity_diffusion.use_segregated_operator = true `

* Multicomponent solve can enabled by setting both input parameters to false. 

* Supports mesh mapping

** Note: The original tensor (default) and multicomponent solve workflows are not affected by the changes.


sbidadi9:
Results of strong scaling study:

image

michaeljbrazell:
I think it would be better to generalize MLABecCecLaplacian to handle multiple components than do muck up amr-wind. Also, we shouldn't invest too much in this segregated approach since it doesn't handle the off-diagonal terms correctly

amr-wind/equation_systems/icns/icns_diffusion.H
```diff
@@ -293,25 +541,41 @@ struct DiffusionOp<ICNS, Scheme>
 {
     std::unique_ptr<ICNSDiffTensorOp> m_tensor_op;
     std::unique_ptr<ICNSDiffScalarOp> m_scalar_op;
+    std::unique_ptr<ICNSDiffScalarSegregatedOp> m_scalar_segregated_op;
```
marchdf, line 544: Just a question here. `ICNSDiffScalarOp` and `ICNSDiffScalarSegregatedOp` share a lot of code. Is there any way they can be consolidated by maybe sharing a base class? The additional advantage of that is that we can avoid many of the `if (use_segregated_op)` below by just instantiating with the right args.

sbidadi9: @marchdf 

Thank you for reviewing the PR. 

The segregated solve class `ICNSDiffScalarSegregatedOp` constructs and performs the solve operation for each of the momentum equation separately. On the other hand, `ICNSDiffScalarOp` constructs a tensor solve similar to what is done in `ICNSDiffTensorOp`. This is why m_solver_scalar is of type`amrex::Array<std::unique_ptr<amrex::MLABecLaplacian>, AMREX_SPACEDIM>`  in `ICNSDiffScalarSegregatedOp` and `std::unique_ptr<amrex::MLABecLaplacian>` in `ICNSDiffScalarOp`.

This PR is the first step towards incorporating the implicit time-stepping solver in AMR-Wind. In the next PR, we will be extending `ICNSDiffScalarSegregatedOp` to compute "C" coefficients in addition to "A" and "B". This is another reason for having a separate class for true segregated solve functionality in AMR-Wind.

marchdf: Ok cool, that makes total sense!

PR #629, opened 2022-06-02
Fix more warnings.

jrood-nrel:


amr-wind/wind_energy/ABLBoundaryPlane.cpp
```diff
@@ -748,14 +748,12 @@ void ABLBoundaryPlane::read_header()
...
             const amrex::Box& minBox = m_mesh.boxArray(lev).minimalBox();
-            const auto& lo = minBox.loVect();
-            const auto& hi = minBox.hiVect();
 
-            amrex::IntVect plo(lo);
-            amrex::IntVect phi(hi);
+            amrex::IntVect plo(minBox.loVect());
+            amrex::IntVect phi(minBox.hiVect());
             const int normal = ori.coordDir();
-            plo[normal] = ori.isHigh() ? hi[normal] + 1 : -1;
```
jrood-nrel, line 757: @michaeljbrazell I get the feeling this should be `plo[normal] = ori.isHigh() ? lo[normal] + 1 : -1;`. Is this a bug?

amr-wind/setup/set_background_pressure.cpp
```diff
@@ -44,6 +44,7 @@ void incflo::set_background_pressure()
 
             if (delp_dir == -1) {
                 delp_dir = dir;
+                // cppcheck-suppress unreadVariable
```
michaeljbrazell, line 47: let's delete this entire function, we have bodyForce now

michaeljbrazell, line 47: let's delete this entire function, we have bodyForce now

jrood-nrel: It's not obvious to me how to do so. It's called here https://github.com/Exawind/amr-wind/blob/da8509fed40eb7508ae7b70f2b0f7996f161aef9/amr-wind/incflo.cpp#L32

michaeljbrazell: yep, just delete that too! I'm pretty sure we never enter the logic inside that function. And if we did (mix of pressure inflow and pressure outflow) it would abort

PR #633, opened 2022-06-13
Joukowsky: Fix bug and add Cp

### LINKED ISSUE #632: Joukowsky disk `num_points_t` dependence
There might be an odd dependence of the Joukowsky disk forces and velocities on the value of `num_points_t`.  As you increase `num_points_t` (keeping everything else constant), the rotor disk velocity and actuator disk forces goes down monotonically:

image

FWIW, the averaged rotor velocity when `num_points_t=1` is closest to what I expect.

Looking at `Joukowksy_opts.H`, there might an extra division by `ddata.num_vel_pts_t` in the `grid.force` computation:  
https://github.com/Exawind/amr-wind/blob/8dea2fbab479a2c75d54c9b0b1e04021a1da7d21/amr-wind/wind_energy/actuator/disk/Joukowsky_ops.H#L207-L209

The area factor computation already accounts for a division by `ddata.num_vel_pts_t`:
```c++
        const amrex::Real dTheta =
            amr_wind::utils::two_pi() / ddata.num_vel_pts_t;

        const amrex::Real geometry_factor = ddata.dr * ddata.dr * dTheta * 0.5;
[..]
                const amrex::Real dArea =
                    geometry_factor * (std::pow(i + 1.0, 2) - std::pow(i, 2));
```

I can try removing the extra division to see if that fixes the problem.

Lawrence

psakievich:
Closes #632

### DIFF
--- a/amr-wind/wind_energy/actuator/disk/ActuatorDisk.H
+++ b/amr-wind/wind_energy/actuator/disk/ActuatorDisk.H
@@ -29,6 +29,10 @@ struct DiskBaseData
     RealList thrust_coeff;
     RealList table_velocity;
     std::string spreading_type{"LinearBasis"};
+    AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE amrex::Real radius() const
+    {
+        return 0.5 * diameter;
+    }
 };
 
 } // namespace actuator
--- a/amr-wind/wind_energy/actuator/disk/Joukowsky.H
+++ b/amr-wind/wind_energy/actuator/disk/Joukowsky.H
@@ -15,6 +15,7 @@ struct JoukowskyData : public DiskBaseData
     RealList root_correction;
     amrex::Real current_angular_velocity{0.0};
     amrex::Real vortex_core_size;
+    amrex::Real current_cp;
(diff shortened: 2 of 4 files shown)


lawrenceccheung:
Thanks Phil.  By removing the extra `num_points_t` division, the rotor averaged velocities are now much better behaved:
```
num_points_t: 1 rotoravgv = 0.906619
num_points_t: 4 rotoravgv = 0.872154
num_points_t: 8 rotoravgv = 0.872165
num_points_t: 12 rotoravgv = 0.872167
```

Lawrence

psakievich:
Okay looks good.  I expect 1 to always be wrong. This is not designed to work with 1 point in theta like the uniform ct disk.  I should probably add a check and error if that is supplied. I was hoping this would converge in 3-5 points in theta.

lawrenceccheung (approved):
Looks good, I think we caught all of the errors!

psakievich:
I'm going to merge this since it is fixing a bug. The linting and docker errors have been happening prior to this PR and I know that none of those issues are related to this PR. If we run into any issues then we can revert it. @michaeljbrazell and @jrood-nrel I expect this to create a diff in the Joukowsky regression test.

PR #640, opened 2022-06-22
Actuators: Make density a sampled quantity

psakievich:
Actuators require users to specify density again.  This makes density a sampled quantity, and a rotor averaged density can now be computed.

This will make it so that actuators don't need any inputs for density, and will allow for more flexibility in future research projects.


lawrenceccheung (approved):
The changes look good.  I tested it with a sample V27 case, and the difference in results are acceptable. 

Lawrence

PR #645, opened 2022-06-29
For the Joukowsky ADM, change input from rad/s to RPM, and add a defa…

neilmatula:
…ult for the vortex core size.


psakievich (approved):
Looks good to me.

PR #647, opened 2022-06-30
ADM: Mesh Mapping

psakievich:
Draft PR for doing the development for mesh mapping.  ~~I've stubbed out everything I think we need to do to get it working via TODO's.~~

This currently will not compile on GPU's but converting one loop to host (comment in this PR) should make it device compatible.

To verify the implementation the following things need to be done:

- [ ]  Mesh mapping fixes need to be made to the mass inflow bc's (required for next step)
- [ ] The uniform or joukowsky disk regression tests should be compared with and without mapping
- [ ] A verification study of mesh convergence for the UniformCt disk should be done under mesh refinement. @lawrenceccheung has done this for the un mapped case, so it should be re-running his case

There is a fundamental limitation for getting converged results for the verification studies without nested refinement boxes.  However this code _should_ work with nested refinement once the technical details for that are resolved.


amr-wind/wind_energy/actuator/actuator_utils.cpp
```diff
@@ -44,6 +44,8 @@ std::set<int> determine_influenced_procs(
     std::set<int> procs;
     const int finest_level = mesh.finestLevel();
     const int nlevels = mesh.finestLevel() + 1;
+    // TODO bx should be mapped to unstretched coordinates here when using mesh
```
psakievich, line 47: Probably best to do the implementation inside `realbox_to_box` as I look closer.

amr-wind/wind_energy/actuator/ActuatorContainer.cpp
```diff
@@ -185,6 +199,46 @@ void ActuatorContainer::update_positions()
...
+
+    const auto dpos = gpu::device_view(m_data.position);
+    const auto* const dptr = dpos.data();
+    const int nlevels = m_mesh.finestLevel() + 1;
+    for (int lev = 0; lev < nlevels; ++lev) {
+        for (ParIterType pti(*this, lev); pti.isValid(); ++pti) {
+            const int np = pti.numParticles();
+            auto* pstruct = pti.GetArrayOfStructs()().data();
+
+            amrex::ParallelFor(np, [=] AMREX_GPU_DEVICE(const int ip) noexcept {
```
psakievich, line 220: The simplest thing to do to make this GPU compliant is to make this a non-device loop.  The code path is preserved for the standard actuator points to still execute on the gpu.

psakievich:
Closing for now. This can be resurrected if/when this work is picked up again.

PR #649, opened 2022-07-15
Merge wrf forcing capability

lawrenceccheung:
Merging in the WRF forcing branch that we used in the 2021 Q4 HFM milestone.  The branch is almost the same as the branch at https://github.com/shashankNREL/amr-wind/tree/sy/wrfForcing developed by @shashankNREL, but updated to be compatible with the latest main.  See https://github.com/Exawind/wind-energy/tree/master/Atlantic-Vineyard for runs with his branch

We will be using this capability in the 2022 Q4 milestone, so I'm adding it now such that others can use the latest AMR-Wind in their simulations.  Additional updates/PR's may follow after this as well.

Lawrence


amr-wind/equation_systems/icns/source_terms/ABLMeanBoussinesq.H
```diff
@@ -1,6 +1,9 @@
 #ifndef ABLMEANBOUSSINESQ_H
 #define ABLMEANBOUSSINESQ_H
 
+#include <AMReX_REAL.H>
+#include <AMReX_Vector.H>
+#include <optional>
```
psakievich, line 6: Is `<optional>` used? I don't see it in the PR.  Without this I don't think there is a need to go to C++17.

ewquon:
@lawrenceccheung, I had actually made a number of improvements to the WRF-forcing code — see https://github.com/shashankNREL/amr-wind/pull/2. Can we pull in those changes as well?

psakievich:
@ewquon can you open a pr on Lawrence's branch and potentially squash it? It looks like there is a lot of git history missing from @shashankNREL 's branch. 13k lines and 420 files feels more than a tad overwhelming to process. 😬

psakievich:
We can/should do this in stages if the changes are really large. So maybe we do yours in a follow on PR. I'm on my iPad and I can't even get it to load your PR.

shashankNREL:
Lawrence and Phil: Thanks for putting this PR in. However, this needs to be cleaned up considerably. For instance, the matrix inverse for momentum and temperature for in-direct forcing is repeated and is hard coded to 3rd order polynomial. Eliot tried to fix some of these issues in his PR but there a number of things that need to be resolved before it is PR state. 
Merging this into AMR-Wind was one of things on my radar but I don't have the time in the 2-3 weeks to do that. But I welcome any help I can get from all of you to get it to a PR stage.

ewquon:
> @ewquon can you open a pr on Lawrence's branch and potentially squash it? It looks like there is a lot of git history missing from @shashankNREL 's branch. 13k lines and 420 files feels more than a tad overwhelming to process. 😬

I had already git merged in a lot of the missing git history in that PR...

psakievich:
@ewquon my point is that your PR against @shashankNREL's  branch includes all the development in amr-wind that has been done over the last year since you've updated the code.  It makes it hard to tell what has been done in the context of this specific PR.  If you instead open a PR against @lawrenceccheung 's branch with your exact same branch then we should have a fraction of the changes to look at and hopefully they will be specific to this work.

amr-wind/equation_systems/icns/source_terms/ABLMeanBoussinesq.H
psakievich: Okay cool.  We probably want to move to 17 at some point, but it would be best to do that as a single commit, and not part of a feature PR.

lawrenceccheung:
Thinking in terms of the work that needs to be done and the milestones that depend on this, I'm wondering if we should review this current PR, get it merged in, and then repeat some of the earlier wrf forced simulations before moving on to the next stage.  

I know that there are some imperfections in terms of the indirect forcing, but since the code has changed so much in the last year since the simulations were run (e.g., with the boundary plane stuff), it would be good to be aware of any adverse interactions before trying an even larger PR.  What are your thoughts @ewquon @shashankNREL?

Lawrence

psakievich:
> Thinking in terms of the work that needs to be done and the milestones that depend on this, I'm wondering if we should review this current PR, get it merged in, and then repeat some of the earlier wrf forced simulations before moving on to the next stage.

This makes a lot of sense to me.  It doesn't need to be perfected to be pushed. I think it is best to do this in stages. If you've already generated benchmarks then this seems like a good state to confirm things are working as before. Merge to get it in the git history and then fix it up and generalize in follow on PR's

ashesh2512:
#652 build on this PR. closing it.

PR #650, opened 2022-07-15
Missing destruction of Omega

ndevelder:
Hi All! 

A couple fixes to RANS model here. Seems like there was a typo in the omega src term? 

Also, if I print out "wd_arr" during a kOmegaSST sim, I get only zeros, and I can't seem to find anywhere that we're computing wall distance. This is important to the blending behavior of k-Omega SST. Maybe we should try and fix that on this PR as well?

### DIFF
--- a/amr-wind/equation_systems/sdr/source_terms/SDRSrc.cpp
+++ b/amr-wind/equation_systems/sdr/source_terms/SDRSrc.cpp
@@ -23,7 +23,7 @@ void SDRSrc::operator()(
     const amrex::Array4<amrex::Real>& src_term) const
 {
     const auto& sdr_src_arr = (this->m_sdr_src)(lev).array(mfi);
-    const auto& sdr_diss_arr = (this->m_sdr_src)(lev).array(mfi);
+    const auto& sdr_diss_arr = (this->m_sdr_diss)(lev).array(mfi);
 
     const amrex::Real factor = (fstate == FieldState::NPH) ? 0.5 : 1.0;
 


psakievich (approved):

ndevelder:
Just tested this "fix" on the channel flow regression test and it blows up quite quickly. Let's hold off on merging until we get a comment from Ganesh and do some more troubleshooting?

marchdf:
@ndevelder I noticed the lack of wall distance calculation in amr-wind as well. This is most likely the reason why this source term was turned off. It would be great to have a wall distance calculator but I think that's a bunch of work... We can see what we can port from nalu-wind to make this happen.

psakievich:
For just amr-wind the wall distance is pretty trivial right? You don't really need an elliptic solve. I think we would need to do a lot of work to get it working with the hybrid solver for lots of reasons.  But for pure amr-wind I think it is just boundary condition logic and some integer math.  For IB it is already available through the level set.

marchdf:
Yeah that totally makes sense for pure amr-wind. I was thinking about the hybrid solver stack and what I want to do there. But that's not needed for this.

psakievich (changes requested):
Putting a blocker on this.

gantech:
I think we should merge this right away even if the wall distance issue is not fixed. This is important for hybrid RANS simulations of wind turbines right now

psakievich:
@ndevelder do you want to comment on this?  This initial fix was unstable correct?

ndevelder:
I'd vote for committing this with the assumption that the physics type or a new version of the turbulence model will handle the wall distance correctly (even though we know this isn't always true?) It's clearly a bug, but I'd rather have it crash or go unstable than never show up in the simulation?

psakievich (approved):

PR #652, opened 2022-07-25
Add new "WRF" mesoscale forcing capabilities

ewquon:
This PR builds upon https://github.com/Exawind/amr-wind/pull/649 and in addition to adding several new profile assimilation capabilities, takes steps to generalize the different `ABLWrfForcing*` source term classes. The new capabilities are documented in https://github.com/shashankNREL/amr-wind/pull/2 and have been designed with backwards compatibility in mind.

This *should* reproduce the results that @lawrenceccheung used in his 2021 Q4 HFM milestone. We're opening a new PR because after discussion with @lawrenceccheung, @shashankNREL, @gdeskos, and @psakievich there appears to be consensus that 1) the code on which this branch is based (https://github.com/shashankNREL/amr-wind/tree/sy/almForceWrf) and the code that Lawrence's branch is similar to (https://github.com/shashankNREL/amr-wind/tree/sy/wrfForcing) have identical functionality and differ only by merges and/or a rebase, 2) this version of the WRF forcing implementation is significantly more advanced, and 3) this is easier than resolving merge conflicts between my branch and Lawrence's branch. With a reminder to not let perfection be the enemy of the good, here are some priority, possibly low-hanging fruits for cleaning up the current code that came up today:
- [ ] A regression test should be set up to support any future development of mesoscale forcing capabilities
- [ ] The `ABLWrfForcingMom` and `ABLWrfForcingTemp` momentum and temperature source terms should be refactored to match the code structure of the `ActuatorForcing` source terms.
- [x] `ABLWrf.*` should be separated into separate header and source files for the `ABLWrfForcing` and `ABLWRFfile` classes
- [x] The nomenclature should be updated for consistency and generality: "WRF" and "Wrf" should be renamed "Meso" ("mesoscale")

An ongoing TODO list can be found here: https://github.com/ewquon/amr-wind/blob/partialProfileAssim/amr-wind/wind_energy/README.profileAssim.md


ewquon:
I believe this build is failing because netcdf needs to be enabled. Anyone foresee any issue with adding `DAMR_WIND_ENABLE_NETCDF:BOOL=ON` to the CI?

psakievich:
I believe we want the code to be able to build without netcdf and if the CI is failing it is making it a hard dependency.  Anything that has netcdf calls should be ifdef'd out so the code can still build.

ewquon:
OK, makes sense.

ashesh2512:
@ewquon Should we be closing #649 ?

ewquon:
@ashesh2512 et al — I think this is ready to merge, considering that other recent PRs have also failed the Lint-cppcheck and the warnings are not only related to the changes made in this PR. The exception would be if @gdeskos wants to take a crack at refactoring the source term or @lawrenceccheung wants to put in a r-test first.

ashesh2512:
@lawrenceccheung Do you have a timeline on a regression test ?

lawrenceccheung:
@ashesh2512 I don't have a timeline on getting a regression test put in yet, but I also don't want to hold up this PR.   We have a couple of example cases that we've used previously, but we need to scale it down in order to get something that can be run quickly.

ewquon:
> @ashesh2512 I don't have a timeline on getting a regression test put in yet, but I also don't want to hold up this PR. We have a couple of example cases that we've used previously, but we need to scale it down in order to get something that can be run quickly.

Could you kick off at least one of those cases to verify that results are comparable? Otherwise I can dig up an old case of mine to do the same.

lawrenceccheung:
@ewquon Yup, I will definitely do that as we need this for upcoming AWAKEN runs (and also, at the very least, provide instructions to others on how to set it up.)  

Lawrence

ewquon:
Hi @lawrenceccheung, I just discovered and fixed a very strange bug. It's unclear why declaring a member string var caused an apparent memory leak (manifested as the CFL blowing up after 1-2 steps; initial residuals changed with each run). Hopefully this hasn't caused you any grief.

ewquon:
Some bug was introduced between amrex v22.06 and v22.07! See attached log files, for example. The latest (v22.08) has the same issue as 22.07. 

[log.latest_amrex2205.txt](https://github.com/Exawind/amr-wind/files/9479795/log.latest_amrex2205.txt)
[log.latest_amrex2206.txt](https://github.com/Exawind/amr-wind/files/9479794/log.latest_amrex2206.txt)
[log.latest_amrex2207.txt](https://github.com/Exawind/amr-wind/files/9479793/log.latest_amrex2207.txt)

ewquon:
@mchurchf I've finished merging the latest upstream changes to my feature branch. It compiles for me on eagle. You can evaluate and merge now as you see fit. There is some discussion of the input options [here](https://github.com/shashankNREL/amr-wind/pull/2).

mchurchf:
So far, this looks good, but do you have an example input file?  To do time-varying profile forcing requires a lot of input, and I can see in your code options for netcdf or maybe text input, but I can't tell how the user needs to input it.

mchurchf (approved):
This looks good to me.  Good job, Eliot, in getting this capability ported over to this code!

PR #656, opened 2022-08-10
added new plane averaged term for wall model

michaeljbrazell:


michaeljbrazell:
PR #667 also adds this functionality through the new fine plane averaging so closing for now

PR #665, opened 2022-08-18
F/geostrophic forcing

JarredK21:
Removed requirement for latitude = 90deg when Geostrophic forcing source term is added to ICNS and added user input to specify three_componentForcing, default is set to two component forcing i.e. only X and Y components of the Coriolis and Geostrophic source terms are calculated. Modified Coriolis and Geostrophic unit tests to test for two and three components. Added regression test for abl_unstable_three_componentForcing.


amr-wind/equation_systems/icns/source_terms/CoriolisForcing.cpp
```diff
@@ -83,9 +95,10 @@ void CoriolisForcing::operator()(
                                up[1] * vel(i, j, k, 1) +
                                up[2] * vel(i, j, k, 2);
 
-        const amrex::Real ae = +corfac * (un * sinphi - uu * cosphi);
+        const amrex::Real ae =
+            +(corfac * un * sinphi) - (corfac * uu * cosphi * S);
         const amrex::Real an = -corfac * ue * sinphi;
-        const amrex::Real au = +corfac * ue * cosphi;
+        const amrex::Real au = +corfac * ue * cosphi * S;
```
gdeskos, line 101: S is a boolean why are you setting it equal to an amrex::Real ?

amr-wind/equation_systems/icns/source_terms/CoriolisForcing.cpp
```diff
@@ -69,6 +80,7 @@ void CoriolisForcing::operator()(
     const auto sinphi = m_sinphi;
     const auto cosphi = m_cosphi;
     const auto corfac = m_coriolis_factor;
+    const auto S = m_S;
```
gdeskos, line 83: if you are trying to use a boolean to take a value of 0 or 1 then create an if statement. I am not sure if this even compiles

test/test_files/abl_unstable_three_componentForcing/abl_unstable_three_componentForcing.i
```diff
@@ -0,0 +1,74 @@
...
+turbulence.model = OneEqKsgsM84
+incflo.physics = ABL
+ICNS.source_terms = BoussinesqBuoyancy CoriolisForcing GeostrophicForcing
+TKE.source_terms = KsgsM84Src
+BoussinesqBuoyancy.reference_temperature = 290.0
+CoriolisForcing.east_vector = 1.0 0.0 0.0
+CoriolisForcing.north_vector = 0.0 1.0 0.0
+CoriolisForcing.latitude = 90.0
+CoriolisForcing.rotational_time_period = 125663.706143592
+GeostrophicForcing.geostrophic_wind = 10.0 0.0 0.0
```
gdeskos, line 35: The reg test does not use the       
 ```
 if (!pp.query("three_ComponentForcing", m_S)) {
         };
```

File: unit_tests/wind_energy/test_abl_src.cpp [hunk 4]
```diff
@@ -135,13 +135,17 @@ TEST_F(ABLMeshTest, body_force)
     }
 }
 
-TEST_F(ABLMeshTest, geostrophic_forcing)
```
gdeskos, line 138: Instead of overwriting create an additional test!

amr-wind/equation_systems/icns/source_terms/CoriolisForcing.cpp
JarredK21: Since a bool can be interpreted as 0 or 1, I was using S to remove or include those terms.

amr-wind/equation_systems/icns/source_terms/CoriolisForcing.cpp
JarredK21: Is this 100% necessary to include? I can confirm it does compile.

test/test_files/abl_unstable_three_componentForcing/abl_unstable_three_componentForcing.i
JarredK21: I have made the changes to the regression test. latitude = 45 and rotational_time_period = 86400.0 (default)

unit_tests/wind_energy/test_abl_src.cpp
JarredK21: I spilt the geostrophic unit test into two (two component and then three component) to make sure the tests were clear.

amr-wind/equation_systems/icns/source_terms/CoriolisForcing.cpp
gdeskos: Jarred, casting a double or int into a boolean is plain wrong. It will be a great source of memory leakage and I cannot approve that. Please add a double and use the boolean only as the logic to assign the correct value to that double S.

JarredK21: I wasn't aware this would be a source of memory leakage. I have added logic to specify double S.

amr-wind/equation_systems/icns/source_terms/CoriolisForcing.cpp
```diff
@@ -80,7 +80,14 @@ void CoriolisForcing::operator()(
...
     const auto corfac = m_coriolis_factor;
-    const auto S = m_S;
+
+    amrex::Real S;
+    if (m_S == true) {
+        S = 1.0
+    } else {
+        S = 0.0
+    }
+
```
gdeskos, line 90: Thanks, just a small suggestion, this can be an one-liner like this
`amrex::Real S = (m_three_dimensional_forcing==true) ? 1.0 : 0.0;`. I would also recommend changing m_S to something that makes more sense when you read the code for the first time.

JarredK21: I have made the changes thanks for the suggestion.

amr-wind/equation_systems/icns/source_terms/CoriolisForcing.cpp
```diff
@@ -21,33 +21,44 @@ namespace icns {
...
+        pp.queryarr("north_vector", m_north, 0, AMREX_SPACEDIM);
+        utils::vec_normalize(m_east.data());
+        utils::vec_normalize(m_north.data());
+        utils::cross_prod(m_east.data(), m_north.data(), m_up.data());
+    }
+
+    {
+        amrex::ParmParse pp("ABL");
+        // 3-component forcing (Default: false)
+        if (!pp.query("three_ComponentForcing", m_three_dimension_forcing)) {
```
gdeskos, line 59: you don't need the if statement anymore. Simply write `pp.query("three_ComponentForcing", m_three_dimensional_forcing)`. Note that you had m_three_dimension_forcing instead of "dimensional"

amr-wind/equation_systems/icns/source_terms/GeostrophicForcing.cpp
```diff
@@ -30,26 +31,40 @@ GeostrophicForcing::GeostrophicForcing(const CFDSim& /*unused*/)
...
         pp.getarr("geostrophic_wind", m_target_vel);
+        latitude = utils::radians(latitude);
+        sinphi = std::sin(latitude);
+        cosphi = std::cos(latitude);
     }
 
+    {
+        amrex::ParmParse pp("ABL");
+        pp.query("three_ComponentForcing", m_three_dimensional_forcing)
+        {
```
gdeskos, line 54: remove this or add a conditional to print the statement. Also, make sure that you use a `;` at the end of every line

amr-wind/equation_systems/icns/source_terms/GeostrophicForcing.cpp
```diff
@@ -30,26 +31,40 @@ GeostrophicForcing::GeostrophicForcing(const CFDSim& /*unused*/)
...
+            amrex::Print() << "cannot find three_componentForcing, "
+                           << "so the default will be used\n";
+        };
+    }
+
+    const auto corfac = coriolis_factor;
```
gdeskos, line 60: What is the need to define `corfac` when you already have the local variable `coriolis_factor` ? Just use it directly

gdeskos (changes requested):

amr-wind/equation_systems/icns/source_terms/GeostrophicForcing.cpp
JarredK21: Removed the print statement and added the ; to the end of the line

amr-wind/equation_systems/icns/source_terms/GeostrophicForcing.cpp
JarredK21: Removed corfac and used coriolis_factor in the calculation.

amr-wind/equation_systems/icns/source_terms/GeostrophicForcing.cpp
```diff
@@ -30,26 +31,35 @@ GeostrophicForcing::GeostrophicForcing(const CFDSim& /*unused*/)
...
-        pp.query("latitude", latitude);
-        AMREX_ALWAYS_ASSERT(
-            amrex::Math::abs(latitude - 90.0) <
-            static_cast<amrex::Real>(vs::DTraits<float>::eps()));
     }
 
     {
         // Read the geostrophic wind speed vector (in m/s)
         amrex::ParmParse pp("GeostrophicForcing");
         pp.getarr("geostrophic_wind", m_target_vel);
```
gdeskos, line 45: add this `pp.query("three_ComponentForcing", m_three_dimensional_forcing);` under GeostrophicForcing

amr-wind/equation_systems/icns/source_terms/CoriolisForcing.cpp
```diff
@@ -83,9 +96,10 @@ void CoriolisForcing::operator()(
                                up[1] * vel(i, j, k, 1) +
                                up[2] * vel(i, j, k, 2);
 
-        const amrex::Real ae = +corfac * (un * sinphi - uu * cosphi);
+        const amrex::Real ae =
+            +(corfac * un * sinphi) - (corfac * uu * cosphi * S);
```
michaeljbrazell, line 100: for cosphi != 0 this will cause a diff

amr-wind/equation_systems/icns/source_terms/GeostrophicForcing.cpp
```diff
@@ -30,26 +31,35 @@ GeostrophicForcing::GeostrophicForcing(const CFDSim& /*unused*/)
...
+        pp.query("three_ComponentForcing", m_three_dimensional_forcing);
+    }
+
+    amrex::Real S = (m_three_dimensional_forcing == true) ? 1.0 : 0.0;
+
     m_g_forcing = {
-        -coriolis_factor * m_target_vel[1], coriolis_factor * m_target_vel[0],
-        0.0};
+        -coriolis_factor * m_target_vel[1] * sinphi +
+            coriolis_factor * m_target_vel[3] * cosphi * S,
```
michaeljbrazell, line 60: this is causing the unit test to fail since m_target_vel is length 3 so you can only use indices 0,1,2. This probably means the unit tests are not sufficient either if you didn't catch this before

amr-wind/equation_systems/icns/source_terms/CoriolisForcing.cpp
JarredK21: I believe this is correct when S = 0 the cosine components are removed (two component forcing) when the three_component forcing flag is true S = 1, thus all the components are calculated.

amr-wind/equation_systems/icns/source_terms/GeostrophicForcing.cpp
JarredK21: I'm not sure this is what I want. The way I currently have it:
amrex::ParmParse pp("ABL");
        pp.query("three_ComponentForcing", m_three_dimensional_forcing);

Makes the user set the number of forcing components under the ABL section, which I think makes sense.

PR #666, opened 2022-08-18
Added power output to Joukowsky disk

lawrenceccheung:
For the Joukowsky actuator disk, I added the power output (in kW) to the netcdf output file.  Could be useful when we compute wind turbine and wind farm performance metrics.

Lawrence


psakievich (approved):

michaeljbrazell (approved):
looks good

PR #667, opened 2022-08-18
added new plane average that works on fine levels

michaeljbrazell:


ashesh2512 (approved):

michaeljbrazell:
This will cause small diffs in all ABL reg tests. For single level this is due to the round off error in the denominator term when calculating plane averages, for multiple level ABL's it's a new implementation using the finest level for wall model and abl forcing. @jrood-nrel

PR #668, opened 2022-08-23
fch uses finest level

michaeljbrazell:
This could be risky if the fine level is not on the ground... an alternative could be to make ABL.log_law_height a required input from now on?


lawrenceccheung:
Thanks for taking a look at this @michaeljbrazell!  Yes, there is a risk that `zref` could end up being in a random location if the finest level is not near the ground.  Is it worth it to add a check to make sure that `zref` lies within the first cell off the ground at level 0, otherwise error out?  I could imagine other checks too, but this way it would prevent something possibly disastrous from happening.

Lawrence

PR #669, opened 2022-08-25
Add BDS advection scheme

marchdf:
This PR does the following:
- adds the BDS advection scheme for velocity and scalars
- uses amrex-hydro calls for PPM and PLM (and so we get to remove a bunch of code)
- fixes a bug in that we were setting `godunov_use_forces_in_trans` to be `false` but by (just) checking the existence of `fq` in the `compute_fluxes` we were actually making implicitly using `godunov_use_forces_in_trans = true` in the postmac, while setting it to `false` in the `premac`. That's inconsistent. That is fixed in this PR. But it also means that we expect diffs for PPM and PLM.
- adds some BDS tests

Draft until I do some extra due diligence and wait for amrex-hydro to merge this: https://github.com/AMReX-Codes/AMReX-Hydro/pull/94


amr-wind/equation_systems/AdvOp_Godunov.H
```diff
@@ -123,17 +133,52 @@ struct AdvectionOp<
...
+                        lev, bx, PDE::ndim, (*flux_x)(lev).array(mfi),
+                        (*flux_y)(lev).array(mfi), (*flux_z)(lev).array(mfi),
+                        (PDE::multiply_rho ? rhotrac : tra_arr),
+                        u_mac(lev).const_array(mfi),
+                        v_mac(lev).const_array(mfi),
+                        w_mac(lev).const_array(mfi),
+                        src_term(lev).const_array(mfi),
+                        dof_field.bcrec_device().data(), iconserv.data(),
+                        tmpfab.dataPtr(), geom, dt, godunov_scheme);
+                } else {
```
jrood-nrel, line 151: I presume this is for BDS? Can it be an `else if` to make it clear?

marchdf: It's not! lol it's for PPM, PLM, _and_ BDS. So we don't have to roll our own code! it's super sweet. Unfortunately PPM_NOLIM, and WENO variants are not in amrex-hydro so I can't blow away as much could as I would have liked ;)

jrood-nrel: Is it still possible to make it an `else if`?

marchdf: yup, done

marchdf:
Feeling pretty good about the way things stand. This is the MMS verification showing 2nd order:
Screen Shot 2022-08-25 at 15 34 54

jrood-nrel (approved):

michaeljbrazell (approved):
looks good, thank you! I'm excited to try out BDS on an ABL flow

marchdf:
Scalar advection also works (thanks @diederikb), and BDS avoids the weird undershoots that PPM creates:

PPM:
image

BDS:
image (1)

PR #671, opened 2022-08-25
Fix dimensionless vortex core size computation

### LINKED ISSUE #670: Typo in root correction for Joukowski disk
There might be a typo in the calculation for the root correction in the Joukowski disk model.  In the [Joukowsky_ops.H](https://github.com/ExaWind/amr-wind/blob/main/amr-wind/wind_energy/actuator/disk/Joukowsky_ops.H):  

https://github.com/Exawind/amr-wind/blob/4805d0038b728eefd79899cb74e8ccf032538abc/amr-wind/wind_energy/actuator/disk/Joukowsky_ops.H#L75-L90

should that be `vortex_core_size` divided by `radius()` for delta?  Otherwise the units don't work out for the arguments in the exponential function.  

Right now the behavior is such that for very large turbine diameters, the root correction is killing all power production, unless we go to ridiculous values of vortex core size or a/b coefficients.

Lawrence

psakievich:
Closes #670

### DIFF
--- a/amr-wind/wind_energy/actuator/disk/Joukowsky_ops.H
+++ b/amr-wind/wind_energy/actuator/disk/Joukowsky_ops.H
@@ -79,7 +79,7 @@ struct InitDataOp<Joukowsky, ActSrcDisk>
             const auto b = meta.root_correction_exponent;
 
             const auto delta = std::max(
-                meta.vortex_core_size * meta.radius(),
+                meta.vortex_core_size / meta.radius(),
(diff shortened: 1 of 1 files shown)


lawrenceccheung:
Thanks @psakievich.  Trying this fix out now with against a real power curve.

Lawrence

lawrenceccheung (approved):
Okay, this looks good, the results are much better behaved.  Now I'm getting things to be on the same order as the real power curve, instead of being like 13 orders of magnitude off.

Lawrence

PR #678, opened 2022-09-09
Updates to SyntheticTurbulence class

diederikb:
This PR introduces the following changes to the SyntheticTurbulence class: 
- Added duration parameter to stop the injection of turbulence after a specified time (ignored if < 0)
- Fixes a bug when calculating the interpolation weights


amr-wind/physics/SyntheticTurbulence.H
```diff
@@ -199,6 +199,8 @@ private:
     amrex::Real m_gauss_scaling;
 
     amrex::Real m_time_offset{0.0};
+    // Duration of the synthetic turbulence injection. Used if < 0
```
marchdf, line 202: Used -> Unused ?

diederikb: Good catch, thanks!

marchdf (approved):

amr-wind/physics/SyntheticTurbulence.cpp
```diff
@@ -268,8 +268,8 @@ AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE void get_lr_indices(
     }
 
     const amrex::Real xfrac = xbox - turb_grid.dx[dir] * il;
-    rxl = xfrac / turb_grid.dx[dir];
-    rxr = (1.0 - rxl);
+    rxr = xfrac / turb_grid.dx[dir];
+    rxl = (1.0 - rxr);
```
gantech, line 272: Hmm... are we sure about this?

amr-wind/physics/SyntheticTurbulence.cpp
```diff
@@ -506,12 +509,18 @@ void SyntheticTurbulence::initialize()
...
     BL_PROFILE("amr-wind::SyntheticTurbulence::update");
+
     // Convert current time to an equivalent length based on the reference
     // velocity to determine the position within the turbulence grid
     const amrex::Real cur_time = m_time.new_time() - m_time_offset;
     const amrex::Real eqiv_len =
         m_wind_profile->reference_velocity() * cur_time;
 
+    // Stop update if the current time is past the duration of the synthetic
+    // turbulence and if the duration is positive
```
gantech, line 520: I thought the approach we wanted was to recycle the input turbulence box after it's finished as implemented here https://github.com/Exawind/amr-wind/blob/main/amr-wind/physics/SyntheticTurbulence.cpp#L266


May be this is not consistent throughout the code?

gantech (changes requested):

amr-wind/physics/SyntheticTurbulence.cpp
diederikb: If `xfrac` is zero, the weight of the right point should be zero and the weight of the left point should be 1, right? This was a sample of some distributions on a grid finer than the turbulence grid before the fix:
pre_fix_distributions
And this is after the fix:
post_fix_distributions

amr-wind/physics/SyntheticTurbulence.cpp
diederikb: Sorry for the confusing language. I was using the word 'duration' to refer to the value of whatever was set in the parameter `m_duration`, not the amount of convective times of the turbulence available in the turbulence box. If `m_duration` is negative, it would indeed recycle the input turbulence box. I just needed a parameter that would allow me to get an isolated patch of turbulence that would be bounded by an undisturbed mean flow both upstream and downstream. So I chose this parameter to be the duration of the injection of the turbulence.

marchdf:
@gantech I think @diederikb addressed your comments? Can you make sure that's the case and then we can merge?

jrood-nrel (approved):

PR #681, opened 2022-09-13
Update up unit vector definition in CoriolisForcing.H

misi9170:
Update to fix typo (?) in "up" unit vector. Please confirm typo.


michaeljbrazell (approved):

ashesh2512:
For those like me who alarmed themselves at first glance - The `up` vector is computed as a cross product of the `north` and `east` vectors. Consequently, the default definition of the `up` vector addressed by this PR was never used.

PR #691, opened 2022-09-29
HELICS interface

michaeljbrazell:
connect to helics, still WIP, need to #ifdef helics


CMakeLists.txt
```diff
@@ -166,6 +167,15 @@ if(AMR_WIND_ENABLE_ASCENT)
   target_compile_definitions(${amr_wind_lib_name} PRIVATE AMR_WIND_USE_ASCENT)
 endif()
 
+# Link with HELICS module 
+if(AMR_WIND_ENABLE_HELICS)
+  set(CMAKE_PREFIX_PATH ${HELICS_DIR} ${CMAKE_PREFIX_PATH})
+  find_package(HELICS 3 REQUIRED)
+  include_directories(AFTER ${HELICS_INSTALL_DIR}/include)
```
jrood-nrel, line 174: Can we use `target_include_directories`?

amr-wind/helics.H
```diff
@@ -0,0 +1,60 @@
...
+#endif
+    double deltat =
+        0.01; // what is this for? should we use m_time.deltaT() instead?
+
+    int m_num_turbines{1};
+
+    double m_inflow_wind_speed_to_amrwind{8.0};
+    double m_inflow_wind_direction_to_amrwind{270.0};
+
+    std::vector<double> m_turbine_power_to_controller;
```
jrood-nrel, line 39: I feel like we should just use `amrex::Vector` and avoid `std` to be consistent.

amr-wind/helics.cpp
```diff
@@ -0,0 +1,239 @@
...
+                abl.abl_statistics().vel_profile().line_average_interpolated(
+                    height, 0);
+            amrex::Real vely =
+                abl.abl_statistics().vel_profile().line_average_interpolated(
+                    height, 1);
+            const amrex::Real turbine_angle = std::atan2(vely, velx);
+            wind_direction = -amr_wind::utils::degrees(turbine_angle) + 270.0;
+        }
+
+        std::cout << "pub count: " << pubCount << std::endl;
```
jrood-nrel, line 187: Even though we're in an `if` statement for the print, I still think we might want to use `amrex::Print()` in case people copy and paste. I'm not sure we want `std::cout` anywhere in the code.

jrood-nrel (approved):

PR #693, opened 2022-09-29
VOF improvements

mbkuhn:
- updating computation and summation of volume fraction fluxes to conserve VOF across coarse-fine boundaries
- deprecating "lagrangian" transport of volume fractions
- setting up arrays that will be components of multiphase momentum solve
- updating existing unit test to check VOF conservation across coarse-fine boundaries
- adding new unit test to check saved VOF advection quantities


amr-wind/equation_systems/vof/split_advection.H
```diff
@@ -222,50 +140,60 @@ AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE void balance_eulerian_fluxes(
         if (bclo == amrex::BCType::ext_dir || bclo == amrex::BCType::hoextrap) {
```
gdeskos, line 140: Can you separate `amrex::BCType::ext_dir` and `amrex::BCType::hoextrap`. The former should be the Dirichlet BC we use for example In wave generation while the second should be a free slip condition

gdeskos (approved):

gdeskos (approved):

PR #694, opened 2022-10-03
Reformulating the ICNS for momentum

mbkuhn:
This converts the setup of the ICNS equation system to update the velocity by solving the momentum equation, as opposed to solving the velocity equation. It involves:
- multiplying source terms by rho^(n+1/2)
- multiplying viscous terms by rho^(n+1/2)
- setting multiply_rho = true for the icns equation_system
- incorporating density into the calculation of advection fluxes

### DIFF
--- a/amr-wind/incflo_advance.cpp
+++ b/amr-wind/incflo_advance.cpp
@@ -325,6 +325,7 @@ void incflo::ApplyPredictor(bool incremental_projection)
     //    and using the half-time density
     // *************************************************************************************
     icns().compute_source_term(amr_wind::FieldState::New);
+    icns().compute_diffusion_term(amr_wind::FieldState::New);
 
     // *************************************************************************************
     // Evaluate right hand side and store in velocity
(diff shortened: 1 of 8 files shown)


amr-wind/incflo_advance.cpp
```diff
@@ -564,7 +565,7 @@ void incflo::ApplyCorrector()
     // *************************************************************************************
     bool incremental = false;
     ApplyProjection(
-        (density_nph).vec_const_ptrs(), new_time, m_time.deltaT(), incremental);
+        (density_new).vec_const_ptrs(), new_time, m_time.deltaT(), incremental);
```
gdeskos, line 568: Should that remain density_nph ?

amr-wind/incflo_advance.cpp
```diff
@@ -349,7 +350,7 @@ void incflo::ApplyPredictor(bool incremental_projection)
     //
     // ************************************************************************************
     ApplyProjection(
-        (density_nph).vec_const_ptrs(), new_time, m_time.deltaT(),
```
gdeskos, line 352: Should that remain density_nph ?

amr-wind/incflo_advance.cpp
mbkuhn: as we have discussed, density_new is intended, and it corresponds to the rest of the changes.

gdeskos (approved):

PR #703, opened 2022-10-17
Bug fix for constant, non-unity density simulations

mbkuhn:
- changing parser query to match norm in input files, regression tests
- only effect is slight difference in convergence of MAC projection


michaeljbrazell (approved):
good catch!

PR #706, opened 2022-10-21
Initial condition input

tonyinme:
This is a feature to read in a netcdf file with initial conditions for velocity generated outside of amr-wind.
The feature lives within the ABL section and it is designed for it.
The number of points in the netcdf file (nx, ny, nz) needs to be the same as in the simulation.

The current implementation works when using one processor, but it doesn't work correctly with more than one processor (reads in the wrong ordering). 
This is because the netcdf reading is still missing something (need help here).
Once this feature is fixed, the PR should become ready for review.

I am tagging the group that has been helping with this development.
@michaeljbrazell @gantech @mbkuhn @rybchuk

### DIFF
--- a/amr-wind/wind_energy/ABLFieldInit.cpp
+++ b/amr-wind/wind_energy/ABLFieldInit.cpp
@@ -184,7 +184,7 @@ void ABLFieldInit::operator()(
         // Read the velocity components u and v
         uvel.get(uvel2.data(), start, count);
         vvel.get(vvel2.data(), start, count);
-        wvel.get(vvel2.data(), start, count);
+        wvel.get(wvel2.data(), start, count);
 
         // Amrex parallel for to assign the velocity at each point
         amrex::ParallelFor(
--- a/amr-wind/wind_energy/ABLFieldInit.cpp
+++ b/amr-wind/wind_energy/ABLFieldInit.cpp
@@ -157,6 +157,7 @@ void ABLFieldInit::operator()(
         // The x, y and z velocity components (u, v, w)
         auto uvel = ncf.var("uvel");
         auto vvel = ncf.var("vvel");
+        auto wvel = ncf.var("wvel");
 
         // Loop through all points in the domain and set velocities to values
         // from the input file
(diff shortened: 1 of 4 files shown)


tonyinme:
This code is ready for review.  Unit tests pass.

Another PR will follow with a python script to generate the netcdf file used in the input.

gantech:
This looks good. If you can get rid of the extra comments, that'll be good.

amr-wind/wind_energy/ABLFieldInit.cpp
```diff
@@ -33,6 +34,9 @@ ABLFieldInit::ABLFieldInit()
 
     pp_abl.query("init_tke", m_tke_init);
 
+    // Use input from netcdf file
+    pp_abl.query("initial_condition_input_file", m_ic_input);
+
```
gantech, line 39: We probly need to perform a few checks. First, that there are no mesh refinement zones enabled. Second, this option will fail or default to some baseline if netcdf is not enabled.

gantech (approved):

amr-wind/wind_energy/ABLFieldInit.cpp
```diff
@@ -126,6 +130,77 @@ void ABLFieldInit::operator()(
...
+        // Set the size of the arrays to the total number of points in this
+        // processor
+        uvel2.resize(count[0] * count[1] * count[2]);
+        vvel2.resize(count[0] * count[1] * count[2]);
+        wvel2.resize(count[0] * count[1] * count[2]);
+
+        // Read the velocity components u and v
+        uvel.get(uvel2.data(), start, count);
+        vvel.get(vvel2.data(), start, count);
+        wvel.get(vvel2.data(), start, count);
```
michaeljbrazell, line 187: should this be wvel2.data()?

tonyinme: good catch! just fixed it :)

amr-wind/wind_energy/ABLFieldInit.cpp
tonyinme: these fixes have been added.

tonyinme:
This PR is ready. 
Here is a cool video showing the turbulence from an ABL simulation initialized with an initial condition from a machine learning algorithm.

Thanks @mbkuhn @gantech @michaeljbrazell @rybchuk 


https://user-images.githubusercontent.com/11055361/200077352-6d3c4a43-a414-40ca-94ee-557df909dbe9.mp4

PR #707, opened 2022-10-25
 Establishing regression tests targeted at multiphase features

mbkuhn:
- rain_drop: water droplet falling under influence of gravity in periodic domain
- inertial_drop: water droplet with initial velocity in quiescent air; inviscid, multiple mesh levels
- vortex_patch_scalar_vel and zalesak_disk_scalar_vel: test momentum 
advection step with a prescribed velocity field
- separate computation of dt for simulations that use a prescribed velocity field


gdeskos (approved):
It looks good!

PR #716, opened 2022-11-10
Remedying FPEs on intel compilers

mbkuhn:
Changing the manner of momentum calculation
- to avoid floating-point errors on intel compilers
- no change to actual function
- FPEs occurred in select MOL reg tests
- (introduced by PR 694)
- these tests now pass on rhodes
- no changes to reg test fcompare results

### DIFF
--- a/amr-wind/equation_systems/icns/icns_advection.H
+++ b/amr-wind/equation_systems/icns/icns_advection.H
@@ -566,10 +566,8 @@ struct AdvectionOp<ICNS, fvm::MOL>
                 amrex::Box gbx = grow(bx, fvm::MOL::nghost_state);
 
                 // Set up momentum array
-                amrex::FArrayBox qfab;
-                qfab.resize(gbx, ICNS::ndim);
-                amrex::Array4<amrex::Real> q;
-                q = qfab.array();
+                amrex::FArrayBox qfab(gbx, ICNS::ndim);
+                const auto& q = qfab.array();
                 // Calculate momentum
                 auto rho_arr = rho(lev).const_array(mfi);
                 auto vel_arr = dof_field(lev).const_array(mfi);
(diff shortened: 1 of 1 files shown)


amr-wind/equation_systems/icns/icns_advection.H
```diff
@@ -576,6 +563,24 @@ struct AdvectionOp<ICNS, fvm::MOL>
             for (amrex::MFIter mfi(dof_field(lev), mfi_info); mfi.isValid();
                  ++mfi) {
                 amrex::Box const& bx = mfi.tilebox();
+                amrex::Box gbx = grow(bx, fvm::MOL::nghost_state);
+
+                // Set up momentum array
+                amrex::FArrayBox qfab;
```
marchdf, line 569: can we do: `amrex::FArrayBox qfab(gbx, ICNS:ndim);` instead of splitting on 2 lines?

amr-wind/equation_systems/icns/icns_advection.H
```diff
@@ -576,6 +563,24 @@ struct AdvectionOp<ICNS, fvm::MOL>
...
+                qfab.resize(gbx, ICNS::ndim);
+                amrex::Array4<amrex::Real> q;
```
marchdf, line 571: and do `const auto& q = qfab.array();` ?

amr-wind/equation_systems/icns/icns_advection.H
mbkuhn: good idea. I'll make this change.

amr-wind/equation_systems/icns/icns_advection.H
mbkuhn: good catch. Added this change to a new commit as well. Will push soon.

marchdf (approved):

jrood-nrel (approved):

PR #717, opened 2022-11-10
Initial condition input

tonyinme:
There was an issue with compiling on the GPUs with this [PR](https://github.com/Exawind/amr-wind/pull/706).
Thanks @jrood-nrel for catching this.

We have worked on a fix.
A few more tests are needed before this is ready for review.
Working with @mbkuhn


tonyinme:
This PR has been updated to work on GPUs. 
Thanks @jrood-nrel for pointing out the issues with the original one.

Worked with @mbkuhn on the latest fix.

amr-wind/wind_energy/ABLFieldInit.cpp
```diff
@@ -126,6 +141,97 @@ void ABLFieldInit::operator()(
...
+        // This file should have the same dimensions as the simulation
+        auto ncf = ncutils::NCFile::open(m_ic_input, NC_NOWRITE);
+
+        // Ensure that the input dimensions match the coarsest grid size
+        const auto& domain = geom.Domain();
+
+        // The indices that determine the start and end points of the i, j, k
+        // arrays. The max and min are there to ensure that the points form
+        // ghost cells are not used
+        auto i0 = std::max(vbx.smallEnd(0), domain.smallEnd(0));
```
jrood-nrel, line 163: Let's use `amrex::max`, etc.

amr-wind/wind_energy/ABLFieldInit.cpp
```diff
@@ -126,6 +141,97 @@ void ABLFieldInit::operator()(
...
+        amrex::Gpu::copy(
+            amrex::Gpu::hostToDevice, &tmp[0], &tmp[dlen - 1] + 1,
+            vvel_d.begin());
+        wvel.get(tmp.data(), start, count);
+        amrex::Gpu::copy(
+            amrex::Gpu::hostToDevice, &tmp[0], &tmp[dlen - 1] + 1,
+            wvel_d.begin());
+
+        // Pointers to velocity objects
+        auto* uvel_dptr = uvel_d.data();
```
jrood-nrel, line 212: Can these be `const`?

amr-wind/wind_energy/ABLFieldInit.cpp
```diff
@@ -126,6 +141,97 @@ void ABLFieldInit::operator()(
...
+
+        // The x, y and z velocity components (u, v, w)
+        auto uvel = ncf.var("uvel");
+        auto vvel = ncf.var("vvel");
+        auto wvel = ncf.var("wvel");
+
+        // Loop through all points in the domain and set velocities to values
+        // from the input file
+        // start is the first index from where to read data
+        std::vector<size_t> start{
```
jrood-nrel, line 180: Let's use `amrex::Vector`.

amr-wind/wind_energy/ABLFieldInit.cpp
tonyinme: yes, they can. fixed.

amr-wind/wind_energy/ABLFieldInit.cpp
tonyinme: fixed

tonyinme:
Hi @jrood-nrel , it looks like this is ready to be merged.
Let me know if you need anything else from our end.

amr-wind/wind_energy/ABL.cpp
```diff
@@ -65,12 +65,19 @@ void ABL::initialize_fields(int level, const amrex::Geometry& geom)
...
-        (*m_field_init)(
-            vbx, geom, velocity.array(mfi), density.array(mfi),
-            temp.array(mfi));
+        interp_fine_levels = (*m_field_init)(
+            vbx, geom, velocity.array(mfi), density.array(mfi), temp.array(mfi),
+            level);
+    }
+
+    if (interp_fine_levels) {
+        // Fill the finer levels with coarse data
```
michaeljbrazell, line 78: this will break if level==2 and Smagorinsky and viscosity=0 since you'll have no strainrate and therefore no eddy viscosity

amr-wind/wind_energy/ABLFieldInit.cpp
```diff
@@ -126,6 +141,97 @@ void ABLFieldInit::operator()(
             velocity(i, j, k, 1) += vfac * damp * z * std::cos(bval * xl);
         }
     });
+
+#ifdef AMR_WIND_USE_NETCDF
+    // Skip fine levels and interpolate data from already loaded coarse levels
```
michaeljbrazell, line 146: can you move this into it's own function kind of like perturb_temperature or init_tke. You could also move this into its own physics class but then you'll have to be careful about order in the input file and call it after ABL... maybe the physics class could check if ABL was created first otherwise abort?

michaeljbrazell: another option is to make another field initializer kind of like field_init. Then this can go in it's own class and you won't have to worry about ordering physics classes

mbkuhn:
This PR was replaced by PR #764, which was merged.

PR #719, opened 2022-11-14
Use amrex-hydro function to compute divergence

marchdf:
I want another set of eyes for when `m_cons==0` in `icns_advection.H`. I mimicked IAMR...


marchdf:
Converting to draft because it needs the latest amrex-hydro and we had to revert that.

michaeljbrazell (approved):

PR #720, opened 2022-11-15
Fix for MOL on GPUs

mbkuhn:
need synchronization after forming momentum array
without this, results vary from run-to-run on GPUs
follow-up to the changes that fixed FPEs on intel

### DIFF
--- a/amr-wind/equation_systems/icns/icns_advection.H
+++ b/amr-wind/equation_systems/icns/icns_advection.H
@@ -579,6 +579,7 @@ struct AdvectionOp<ICNS, fvm::MOL>
                 // Doing this explicitly, instead of through a Multiply command,
                 // helps avoid floating-point errors with intel compilers and
                 // mimics the implementation in equation_systems/AdvOp_MOL.H
+                amrex::Gpu::streamSynchronize();
 
                 amrex::Box tmpbox = amrex::surroundingNodes(bx);
                 const int tmpcomp = nmaxcomp * AMREX_SPACEDIM;


amr-wind/equation_systems/icns/icns_advection.H
```diff
@@ -579,6 +579,7 @@ struct AdvectionOp<ICNS, fvm::MOL>
                 // Doing this explicitly, instead of through a Multiply command,
                 // helps avoid floating-point errors with intel compilers and
                 // mimics the implementation in equation_systems/AdvOp_MOL.H
+                amrex::Gpu::streamSynchronize();
```
marchdf, line 582: I am thinking something else is up... Can you remove the synch and do `amrex::FArrayBox qfab(gbx, ICNS::ndim, amrex::The_Async_Arena());` I think that should be the right fix to the problem.

mbkuhn: I'm afraid that this modification isn't an improvement. I end up getting the following error message while running: 
```
  what():  CUDA error 2 in file /home/mkuhn/spack-amr-wind-exawind/amr-wind/submods/amrex/Src/Base/AMReX_PArena.cpp line 44: out of memory
```
To me, the synchronization call mimics what's already done in the Godunov advection implementation

marchdf: Thanks for trying that out. To me, it's the correct fix. But clearly it's creating an OOM. Jon and I have been trying to figure out why that is. We are seeing it in a bunch of codes when we use the async arena. I am guessing this is on Eagle? And we've seen that it goes away if we use managed memory and/or 1 MPI rank. In any event... until we figure this out, lets use the sync route.

marchdf (approved):

amr-wind/equation_systems/icns/icns_advection.H
mbkuhn: Ok, that's interesting, good to know. Yes, it's on Eagle. Works for me.

jrood-nrel (approved):

PR #723, opened 2022-11-18
Use a common CDash site for all projects

psakievich:
See https://github.com/Exawind/nalu-wind/pull/1081


sayerhs:
@psakievich  Can you also update the links in the README file?

PR #730, opened 2022-11-22
Fillpatch umac for Godunov and BDS

marchdf:
amrex-hydro assumes that the mac velocities ghost cells are filled. Both IAMR and incflo (https://github.com/AMReX-Codes/incflo/blob/development/src/convection/incflo_compute_advection_term.cpp#L159) do this. Weird corner effects happen if we don't.

This is draft because I am currently unsure how to get references to the velocity `PhysBCFunct`...


marchdf:
I decided that the way I am going to do this is by implementing a `fillpatch_other`for a field: fillpatches other fields based on this fields BC. Happy to hear other ideas.

marchdf:
This PR fixes 2 issues:

1. BDS needs the mac velocities to contain valid ghost data

Without this PR, a [freestream case](https://github.com/Exawind/amr-wind/files/10106183/freestream_godunov.inp.txt) shows weird corner effects:
Screen Shot 2022-11-28 at 10 38 56

With this PR:

Screen Shot 2022-11-28 at 10 37 20


2. Godunov needs a fillpatch with multilevel AMR. 

WIthout this PR, [this case](https://github.com/Exawind/amr-wind/files/10106200/freestream-amr.inp.txt) shows weird corner effects at the AMR
Screen Shot 2022-11-28 at 10 45 20

With this PR:
Screen Shot 2022-11-28 at 10 47 01

amr-wind/core/Field.H
```diff
@@ -318,6 +318,10 @@ public:
 
     void fillpatch(amrex::Real time) noexcept;
     void fillpatch(amrex::Real time, amrex::IntVect ng) noexcept;
+    void fillpatch(
```
sayerhs, line 321: Seems like the least friction way to achieve what you want. My only suggestion would be to name this method differently (e.g., `fillpatch_sibling_fields` or something along those lines) to indicate in an explicit way that this method is doing something different than what the other `fillpatch` methods are doing.

sayerhs (approved):

amr-wind/core/Field.H
marchdf: Good idea. Done.

amr-wind/core/FieldFillPatchOps.H
```diff
@@ -264,6 +292,92 @@ public:
...
+        const auto& geom = m_mesh.Geom(lev);
+        const auto& domain = geom.growPeriodicDomain(ng);
+        auto& sfab = m_field.state(fstate)(lev);
+        for (amrex::OrientationIter oit; oit != nullptr; ++oit) {
+            auto ori = oit();
+            const int idir = ori.coordDir();
+            const auto& dbx =
+                ori.isLow() ? amrex::adjCellLo(domain, idir, nghost[idir])
+                            : amrex::adjCellHi(domain, idir, nghost[idir]);
+
```
michaeljbrazell, line 316: to fix the abl_bndry_plane_input_native I checked for inflow BC's here and only copied them in when there is an inflow. 
`        if (bctype[ori] != BC::mass_inflow) {
            continue;
        }`
        
what do you think? or are there other BC's we should be copying face data back in?

marchdf: I thought long and hard about that one and decided that it was safest to copy everything. Clearly I was wrong. Great find! I think inflow is probably the only one that matters... I am also moving the fillpatch_siblings call in preadvect as we discussed.

michaeljbrazell (approved):
looks good!

PR #742, opened 2022-12-01
Zalesak reg tests: changes to avoid out-of-bounds

mbkuhn:
These changes return the initialization loops to its original functionality, despite the change in the number of levelset ghost cells.

### DIFF
--- a/amr-wind/physics/multiphase/ZalesakDisk.cpp
+++ b/amr-wind/physics/multiphase/ZalesakDisk.cpp
@@ -51,7 +51,7 @@ void ZalesakDisk::initialize_fields(int level, const amrex::Geometry& geom)
     const amrex::Real depth = m_depth;
 
     for (amrex::MFIter mfi(levelset); mfi.isValid(); ++mfi) {
-        const auto& vbx = mfi.growntilebox(1);
+        const auto& gbx = mfi.growntilebox(1);
         auto uf = u_mac.array(mfi);
         auto vf = v_mac.array(mfi);
         auto wf = w_mac.array(mfi);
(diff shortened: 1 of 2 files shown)


mbkuhn:
I can confirm that these fix the zalesak reg test issues on rhodes with intel compilers and locally with gnu compilers.

amr-wind/physics/multiphase/ZalesakDisk.cpp
```diff
@@ -51,7 +51,7 @@ void ZalesakDisk::initialize_fields(int level, const amrex::Geometry& geom)
     const amrex::Real depth = m_depth;
 
     for (amrex::MFIter mfi(levelset); mfi.isValid(); ++mfi) {
-        const auto& vbx = mfi.growntilebox();
+        const auto& vbx = mfi.growntilebox(1);
```
gdeskos, line 54: I know this is a minor, but can you rename it as gbx (grown box) instead of vex (valid box) ? It is just a matter of convention

amr-wind/physics/multiphase/ZalesakDiskScalarVel.cpp
```diff
@@ -84,7 +84,7 @@ void ZalesakDiskScalarVel::initialize_fields(
     const amrex::Real depth = m_depth;
 
     for (amrex::MFIter mfi(levelset); mfi.isValid(); ++mfi) {
-        const auto& vbx = mfi.growntilebox();
+        const auto& vbx = mfi.growntilebox(1);
```
gdeskos, line 87: same here

gdeskos (approved):

PR #743, opened 2022-12-01
Fix inflow with AMR

### LINKED ISSUE #174: Fix ABL boundary planes inflow when using multiple levels
If more than one level is present, reading inflow boundary conditions from higher levels leads to spurious velocities at the coarse-fine interface (see image). Disabling the velocity solve does not fix the issue. Disabling the nodal projection does fix the issue. It is unclear where this issue comes from. The initial thought was this came from ghost cells not being filled correctly at the fine level. This may not be the case... Currently the code aborts if the user tries to use multiple levels for inflow.

Screen Shot 2020-07-13 at 11 40 45 AM

### LINKED ISSUE #735: AMR at inflow not working
Having AMR levels touch an inflow BC leads to spurious velocity at the AMR corners. This input file illustrates the issue: [freestream-amr.inp.txt](https://github.com/Exawind/amr-wind/files/10116961/freestream-amr.inp.txt). Inflow at xlo and ylo with vx = vy = 1.0.

Screen Shot 2022-11-29 at 13 17 00

This is probably related to #174 as well.

marchdf:
On initialization the nodal projection creates spurious oscillations at the coarse-fine inflow BC. This edit fixes that. But it's not the end of the story because it NaN's after 1 time step. But I am getting closer I think... 

Here's the input file I am using to test [inflow-amr.inp.txt](https://github.com/Exawind/amr-wind/files/10135627/inflow-amr.inp.txt)

Left over tasks:
- [x] add regression tests: `inflow_godunov_amr.inp`, `inflow_bds_amr.inp`, `freestream_bds.inp`
- [x] figure out MPI and smaller `max_grid_size`
- [x] check that this works with other inflow profiles like power law and linear
- [x] check that this works with ABL inflow
- [x] check that this works with MPL

Closes #735 
Closes #174


marchdf:
Seems like the correct fix to all this madness is to actually change the inflow BC functions to accept a dest and orig components (instead of the copy I was doing before). But I don't know how/if this works for non-standard inflow like ABL inflow.

amr-wind/wind_energy/ABLBoundaryPlane.cpp
```diff
@@ -884,18 +879,6 @@ void ABLBoundaryPlane::populate_data(
             continue;
         }
 
-        // Ensure the fine level does not touch the inflow boundary
-        if (lev > 0) {
-            const amrex::Box& minBox = m_mesh.boxArray(lev).minimalBox();
```
michaeljbrazell, line 889: are you sure we can remove this continue?

marchdf: ok I see the problem. If the inflow boundary plane data doesn't have a fine level, then there's a problem. Great catch!

marchdf: I put the continue back in

michaeljbrazell (approved):

PR #749, opened 2022-12-09
Port of radar and lidar sampling code from Nalu-wind to AMR-Wind

ndevelder:


lawrenceccheung:
Reopening PR because it is still active.

PR #763, opened 2022-12-21
Replace floating point equality checks

mbkuhn:
floating pt equality checks now use abs(), epsilon


jrood-nrel (approved):
Excellent! Thanks!

PR #769, opened 2023-01-04
compute l2 error over both fringe and field points

ashesh2512:
Currently l2 error is compute over only field points. This is inconsistent with nalu-wind which computes l2 errors over both field and fringe points. Additionally, the l2 error slope over field points may not always be representative of the l2 error slope over fringe points.

### DIFF
--- a/amr-wind/physics/ConvectingTaylorVortex.cpp
+++ b/amr-wind/physics/ConvectingTaylorVortex.cpp
@@ -248,7 +248,7 @@ amrex::Real ConvectingTaylorVortex::compute_error(const Field& field)
                 const auto& imask_arr = level_mask.array(mfi);
                 amrex::ParallelFor(
                     vbx, [=] AMREX_GPU_DEVICE(int i, int j, int k) noexcept {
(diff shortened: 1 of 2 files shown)


amr-wind/physics/ConvectingTaylorVortex.cpp
```diff
@@ -248,7 +248,7 @@ amrex::Real ConvectingTaylorVortex::compute_error(const Field& field)
                 const auto& imask_arr = level_mask.array(mfi);
                 amrex::ParallelFor(
                     vbx, [=] AMREX_GPU_DEVICE(int i, int j, int k) noexcept {
-                        if (iblank_arr(i, j, k) < 1) {
+                        if (std::abs(iblank_arr(i, j, k)) < 1) {
```
jrood-nrel, line 251: Can you use `amrex::Math::abs` instead?

ashesh2512: @jrood-nrel done

jrood-nrel (approved):

PR #771, opened 2023-01-12
SST bug fixes - Inconsistencies with Model Definition

ndevelder:
Two fixes to the SDR source term...both involve using the clipped version of a term when the full term should be used instead.


gantech (approved):
Thanks for getting this done!

ashesh2512 (approved):

ashesh2512:
@jrood-nrel This might break some of the k-w sst tests, but I can't be certain.

PR #778, opened 2023-02-02
time precision

michaeljbrazell:
Does this fix the precision in your simulation @rthedin ?


rthedin:
Yep, it does. Thanks for the quick and simple fix.

michaeljbrazell:
I think this is relatively safe to merge, it shouldn't go past the current line length

jrood-nrel (approved):

PR #779, opened 2023-02-03
Apply same bug fixes in SST to IDDES implementation

gantech:
This was pointed out by @ashesh2512


ashesh2512:
Closing following #780

PR #783, opened 2023-02-06
Pressure BC for wave generation

gdeskos:
Introducing a pressure bc for wave generation. Currently this BC only works for linear waves

- No unit tests
- No regressions tests needed (already exist)


gdeskos:
This PR is not needed anymore

PR #793, opened 2023-03-02
Perturbational pressure addressed: source terms and more

mbkuhn:
* Previously, GravityForcing and DensityBuoyancy both implemented the same terms, but the latter considered a perturbational form of density, implying a perturbational form of the pressure solution. 
* These perturb. forms enable better dirichlet BCs for the pressure equation without changing the way pressure is solved or implementing new BCs.
* A perturbational source term also implies a modification of the pressure solution from the Poisson solve
* This PR enables a perturbational form for these terms in a multiphase context, allows the true pressure to be reconstructed after being solved. This allows GravityForcing to have the same (and more) functionality as DensityBuoyancy, which can be removed in a future PR if desired.

### DIFF
--- a/amr-wind/equation_systems/icns/source_terms/GravityForcing.cpp
+++ b/amr-wind/equation_systems/icns/source_terms/GravityForcing.cpp
@@ -25,9 +25,9 @@ GravityForcing::GravityForcing(const CFDSim& sim)
     amrex::ParmParse pp_icns("ICNS");
     pp_icns.query("use_perturb_pressure", is_pptb);
     // Determine if rho0 field exists
-    is_rho0 = sim.repo().field_exists("rho0");
+    is_rho0 = sim.repo().field_exists("reference_density");
     if (is_rho0) {
-        m_rho0 = &(sim.repo().get_field("rho0"));
+        m_rho0 = &(sim.repo().get_field("reference_density"));
     } else {
         // Point to existing field, won't be used
         m_rho0 = m_rho;
--- a/amr-wind/equation_systems/icns/source_terms/GravityForcing.cpp
+++ b/amr-wind/equation_systems/icns/source_terms/GravityForcing.cpp
@@ -12,10 +12,22 @@ namespace amr_wind::pde::icns {
  *
  *  - `gravity` acceleration due to gravity (m/s)
  */
-GravityForcing::GravityForcing(const CFDSim& /*unused*/)
+GravityForcing::GravityForcing(const CFDSim& sim)
 {
     amrex::ParmParse pp("incflo");
     pp.queryarr("gravity", m_gravity);
+
+    // Get density fields
+    m_rho = &(sim.repo().get_field("density"));
+
+    // Determine if rho0 field exists
+    is_rho0 = sim.repo().field_exists("rho0");
+    if (is_rho0) {
+        m_rho0 = &(sim.repo().get_field("rho0"));
+    } else {
+        // Point to existing field, won't be used
+        m_rho0 = m_rho;
+    }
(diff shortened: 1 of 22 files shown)


amr-wind/physics/multiphase/hydrostatic_ops.H
```diff
@@ -0,0 +1,70 @@
...
+                    // Integrated (top-down in z) phase heights to pressure node
+                    amrex::Real ih_g = amrex::max(
+                        0.0, amrex::min(probhi[2] - hliq, probhi[2] - hnode));
+                    amrex::Real ih_l = amrex::max(
+                        0.0, amrex::min(hliq - hnode, hliq - problo[2]));
+                    // Integrated rho at pressure node
+                    const amrex::Real irho = rho1 * ih_l + rho2 * ih_g;
+
+                    // Add term to reference pressure
+                    p0_arr(i, j, k) = -irho * grav_z;
```
gdeskos, line 62: Is this "-" meant to take care of the "-" in grav_z ?

docs/sphinx/theory/theory.rst
```diff
@@ -160,6 +160,32 @@ Re-initialization of the level-set
...
+
+Using the perturbational form implies that the hydrostatic pressure is removed from the pressure variable, including its output. This means that the solution to the Poisson equation is actually the perturbational pressure, :math:`p'`, not :math:`p`. If the full pressure, :math:`p`, is desired for analysis or postprocessing purposes, the hydrostatic pressure can be added back to the pressure field via the input argument ``ICNS.reconstruct_true_pressure = true``. In order for this to operate in the code, the reference pressure field must be defined for the specific flow case being run. 
+
+- An example of this is in physics/multiphase/Multiphase.cpp. To construct the reference pressure field, the reference gravity term must be integrated. This particular example assumes that the reference density only varies in z (or is constant), gravity acts only in z, and the hydrostatic pressure at zhi is equal to 0. 
+
+- In mathematical form, the derivation and calculation of the full pressure is as follows:
+
+.. math:: \nabla p = \nabla p + \rho_0 \boldsymbol{g}
```
gdeskos, line 179: should the first right-hand side term be \nabla p' ?

gdeskos (approved):

amr-wind/equation_systems/icns/source_terms/GravityForcing.cpp
```diff
@@ -29,19 +45,30 @@ GravityForcing::~GravityForcing() = default;
...
+
     amrex::ParallelFor(bx, [=] AMREX_GPU_DEVICE(int i, int j, int k) noexcept {
+        const amrex::Real factor =
+            (!ipt ? 1.0
+                  : 1.0 - (ir0 ? rho0_arr(i, j, k) : mr0c) / rho_arr(i, j, k));
+
         vel_forces(i, j, k, 0) += gravity[0];
         vel_forces(i, j, k, 1) += gravity[1];
-        vel_forces(i, j, k, 2) += gravity[2];
+        vel_forces(i, j, k, 2) += gravity[2] * factor;
```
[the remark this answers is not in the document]
mbkuhn: the reconstruction of the full pressure assumes that this is only in z, but you're right, the factor can be multiplied to all the terms to keep it general. I made this change.

amr-wind/physics/multiphase/hydrostatic_ops.H
mbkuhn: basically, yes. But a more accurate way to think about it is that it's from integrating top-down instead of bottom-up, which introduces a negative sign.

PR #795, opened 2023-03-03
Ruvm rebase

itopcuoglu:
plot_power
plot_thrust


itopcuoglu:
Removed unified memory from TiogaInterface

* Host allocated m_qcell_host and m_qnode_host are passed to TIOGA
* Host allocated iblank_cell_host and iblank_node_host are passed to TIOGA
* The newly added IntScratchField class enables integer scratch fields
* A unit test for IntScratchField is added
* amr_to_tioga_mesh is split into two functions

PR #796, opened 2023-03-03
Unified memory removed from TiogaInterface

itopcuoglu:
Removed unified memory from TiogaInterface

* Host allocated m_qcell_host and m_qnode_host are passed to TIOGA
* Host allocated iblank_cell_host and iblank_node_host are passed to TIOGA
* The newly added IntScratchField class enables integer scratch fields
* A unit test for IntScratchField is added
* amr_to_tioga_mesh is split into two functions
* 
plot_power
plot_thrust

### DIFF
--- a/amr-wind/core/field_ops.H
+++ b/amr-wind/core/field_ops.H
@@ -114,6 +114,53 @@ copy(T1& dst, const T2& src, int srccomp, int dstcomp, int numcomp, int nghost)
     copy(dst, src, srccomp, dstcomp, numcomp, amrex::IntVect(nghost));
 }
 
+/** Copy source field to destination field
+ *  \ingroup field_ops
+ *
+ *  \tparam T1 IntScratchField
+ *  \tparam T2 IntScratchField
+ *  \param [out] dst Field that is updated
+ *  \param [in] src Field to be added
+ *  \param [in] srccomp Starting component index of source field
+ *  \param [in] dstcomp Starting component index of destination field
+ *  \param [in] numcomp Number of components to be updated
+ *  \param [in] nghost Number of ghost cells to be updated
+ */
+template <typename T1, typename T2>
+inline void copy_int(
+    T1& dst,
+    const T2& src,
+    int srccomp,
+    int dstcomp,
+    int numcomp,
+    const amrex::IntVect& nghost)
+{
+    const int nlevels = dst.repo().num_active_levels();
+    for (int lev = 0; lev < nlevels; ++lev) {
+        amrex::iMultiFab::Copy(
+            dst(lev), src(lev), srccomp, dstcomp, numcomp, nghost);
+    }
+}
+
+/** Copy source field to destination field
+ *  \ingroup field_ops
+ *
+ *  \tparam T1 IntScratchField
+ *  \tparam T2 IntScratchField
+ *  \param [out] dst Field that is updated
+ *  \param [in] src Field to be added
+ *  \param [in] srccomp Starting component index of source field
+ *  \param [in] dstcomp Starting component index of destination field
+ *  \param [in] numcomp Number of components to be updated
+ *  \param [in] nghost Number of ghost cells to be updated
+ */
+template <typename T1, typename T2>
+inline void
+copy_int(T1& dst, const T2& src, int srccomp, int dstcomp, int numcomp, int nghost)
+{
+    copy(dst, src, srccomp, dstcomp, numcomp, amrex::IntVect(nghost));
+}
+
 /** Perform operation \f$y = a x + y\f$
  *  \ingroup field_ops
  *
--- a/amr-wind/overset/TiogaInterface.H
+++ b/amr-wind/overset/TiogaInterface.H
@@ -93,8 +93,11 @@ public:
 
     AMROversetInfo& amr_overset_info() { return *m_amr_data; }
 
+	// These two are functions that can access the private
+	// m_qcell and m_qnode variables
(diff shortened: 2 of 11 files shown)


ashesh2512:
@jrood-nrel As a test, could you please try this commit on Frontier test bed? Would be good if someone else tried this too.

amr-wind/core/field_ops.H
```diff
@@ -114,6 +114,53 @@ copy(T1& dst, const T2& src, int srccomp, int dstcomp, int numcomp, int nghost)
...
+    const amrex::IntVect& nghost)
+{
+    const int nlevels = dst.repo().num_active_levels();
+    for (int lev = 0; lev < nlevels; ++lev) {
+        amrex::iMultiFab::Copy(
+            dst(lev), src(lev), srccomp, dstcomp, numcomp, nghost);
+    }
+}
+
+/** Copy source field to destination field
```
marchdf, line 145: If I am reading this right, `copy_int` here is the same as `copy` on L112. The templating takes care of things so I don't think this is necessary?

amr-wind/overset/TiogaInterface.H
```diff
@@ -93,12 +94,15 @@ public:
 
     AMROversetInfo& amr_overset_info() { return *m_amr_data; }
 
-    ScratchField& qvars_cell() { return *m_qcell; }
-    ScratchField& qvars_node() { return *m_qnode; }
+    ScratchField& qvars_cell() { return *m_qcell_host; }
```
marchdf, line 97: do you want to add a `_host` to the function names for these 2 functions so it is clear it returns a host object?

amr-wind/overset/TiogaInterface.cpp
```diff
@@ -7,7 +7,7 @@
 
 #include <memory>
 #include <numeric>
-
+#include <AMReX_Print.H>
```
marchdf, line 10: is this needed? Or part of a debugging thing?

amr-wind/overset/TiogaInterface.cpp
```diff
@@ -102,20 +104,45 @@ void TiogaInterface::post_regrid_actions()
...
+    const int num_ghost = m_sim.pde_manager().num_ghost_state();
+    m_iblank_cell_host = repo.create_int_scratch_field_on_host("iblank_cell_host",1,num_ghost, FieldLoc::CELL);
+    m_iblank_node_host = repo.create_int_scratch_field_on_host("iblank_node_host",1,num_ghost, FieldLoc::NODE);
+
+    amr_to_tioga_iblank();
+
     m_iblank_cell.setVal(1);
     m_iblank_node.setVal(1);
+
+    (*m_iblank_cell_host).setVal(1);
```
marchdf, line 118: can just do `m_iblank_cell_host->setVal` no?

amr-wind/overset/TiogaInterface.cpp
```diff
@@ -291,37 +386,68 @@ void TiogaInterface::amr_to_tioga_mesh()
...
     auto& ibnode = m_sim.repo().get_int_field("iblank_node");
+
+
+    // Create standard vector of integer pointers here
+    auto& ad = *m_amr_data;
+    amrex::Vector<int *> tmpdataPtr(ad.iblank_cell.size());
+    amrex::Vector<int *> tmpdataPtrn(ad.iblank_node.size());
     for (int lev = 0; lev < nlevels; ++lev) {
-        auto& ad = *m_amr_data;
+        //auto& ad = *m_amr_data;
```
marchdf, line 420: remove this

amr-wind/overset/TiogaInterface.H
itopcuoglu: I left the function name as it was since it is called from AMRTiogaIface from inside Exawind driver, and I wanted to avoid making a change in Exawind source. I can go ahead and make that change.

amr-wind/overset/TiogaInterface.cpp
itopcuoglu: I must have missed it. Thanks for pointing it out.

amr-wind/overset/TiogaInterface.cpp
itopcuoglu: This was used for debugging. I will delete it. Thanks for pointing it out.

amr-wind/core/field_ops.H
itopcuoglu: The copy function calls amrex::MultiFab::Copy, so the template does not allow the use of iMultiFab objects as far as I can tell.

I think I stopped using the copy_int function when I switched to using IntScratchField instead of a host IntField for iblanks, so I will eliminate this after making sure that it is not called from anywhere. Thanks for pointing this out.

amr-wind/overset/TiogaInterface.cpp
itopcuoglu: Calling setVal here requires the dereferencing of a pointer since m_iblank_cell_host and m_iblank_node_host are unique_ptrs to an IntScratchField, while m_iblank_cell and m_iblank_node are references to IntField objects.

amr-wind/core/field_ops.H
itopcuoglu: It appears that I was not using copy_int anymore, so I deleted it.

jrood-nrel (approved):
This seems to run the 16 turbine case on Crusher fine.

ashesh2512 (approved):

PR #797, opened 2023-03-03
velocity/temperature fluctuations 

michaeljbrazell:
added feature to sample velocity/temperature fluctuations at zref height


marchdf:
This look almost ready to go? Is that right?

michaeljbrazell:
I think that @mchurchf wanted to incorporate this feature into his PR. I would check with him if arbitrary height (and this PR) is necessary

PR #799, opened 2023-03-07
FLLC: Add updated formulation

psakievich:
Add the new formulation for the filtered lifting line correction that can account for variable chord length.

This PR preserves the original formulation we've been using, and it can be accessed by setting `fllc.fllc_type=constant_chord`.

The new formulation should be superior in most instances and is made the new default.


tonyinme (commented):
Hi @psakievich , the implementation looks good.
The only thing missing is a few lines in the documentation to document the new options.
https://exawind.github.io/amr-wind/user/inputs_Actuator.html#input_param-Actuator.FixedWingLine.fllc

tonyinme (approved):

PR #800, opened 2023-03-08
Wall Model as Wall-Adjacent Source Term with Arbitrary Height Sampling

mchurchf:
This is a "draft" pull request for feedback.  This changes the wall shear stress and temperature flux model imposition from a specified gradient times surface effective viscosity/diffusivity to a source term in the cells adjacent to the wall.  The advantage is that this gives control over all three components of wall stress, tau_xz, tau_yz, and tau_zz.  The original formulation gives no control over tau_zz.  This formulation, being a source term, is equivalent to div(tau_wall), which is applied at cell center, but it is fully explicit (although a wall gradient goes into the RHS, so I am not sure the original way was not explicit).  Another advantage to this method is that it is theoretically extensible to immersed boundary methods for terrain and waves as it is a source term that can be placed anywhere on the grid.  There is an ability to sample fluctuating velocity at any height within the grid boxes adjacent to the surface (see Mike Brazell's pull request that makes that functionality completely general using AMReX slices which I will attempt to blend with my PR).  This PR also includes updates to the MO and ShearStress computations per lengthy discussion and derivation with Jim Brasseur.  This PR does not include any unit/regression tests, yet.  Those will come in the coming week.  Note my complete hack on zeroing effective viscosity at the surface that needs to be generalized.  Ideally, I want these source terms to come on if wall model is specified in BCs, not through specification of source term, to make this more logical for the user.


gantech:
Hey @mchurchf Could you please provide a document or a scan of some handwritten pages explaining what was the approach earlier and how this pull request changes the formulation?

amr-wind/equation_systems/icns/source_terms/WallMomentumFluxForcing.cpp
```diff
@@ -0,0 +1,214 @@
...
+               << std::endl; std::cout << "v = " << vLow << " " << vHigh << " "
+               << v << std::endl; std::cout << "S = " << S << std::endl;
+            */
+
+            // Get local tau_wall based on the local conditions and
+            // mean state based on Monin-Obukhov similarity.
+            amrex::Real tau_xz = 0.0;
+            amrex::Real tau_yz = 0.0;
+
+            if (m_wall_shear_stress_type == "constant") {
```
AMLattanzi, line 144: These if statements are causing the GPU compilation to fail. Specifically, the m_wall_shear_stress_type is a data member of the class WallMomentumFluxForcing, which would require a capture of 'this' to access. Instead, you should expose these data members before the kernel so that they may be copied onto the GPU; see below example. The second issue will arise in the use of std::string and ==; the compiler will bark about the capture and using host code for the equality. I would advocate for simply using integers to denote different options; sticking with strings would require you to copy to a device vector and then expose the char* and then the equality test becomes nasty.

`int d_wall_shear_stress_type = m_wall_shear_stress_type;
...
amrex::ParallelFor(box,[=] AMREX_GPU_DEVICE(int i, int j, int k) noexcept {
...
if(d_wall_shear_stress_type == 0){
...`

amr-wind/equation_systems/temperature/source_terms/WallTemperatureFluxForcing.cpp
```diff
@@ -0,0 +1,211 @@
...
+               << v << std::endl; std::cout << "S = " << S << std::endl;
+                        std::cout << "T = " << TLow << " " << THigh << " " << T
+               << std::endl;
+            */
+
+            // Get local tau_wall based on the local conditions and
+            // mean state based on Monin-Obukhov similarity.
+            amrex::Real q = 0.0;
+
+            if (m_wall_shear_stress_type == "constant") {
```
AMLattanzi, line 148: Similar comment as with WallMomentumFluxForcing.cpp

amr-wind/equation_systems/icns/source_terms/WallMomentumFluxForcing.cpp
```diff
@@ -0,0 +1,214 @@
...
+
+            // Adding the source term as surface stress vector times surface
+            // area divided by cell volume (division by cell volume is to make
+            // this a source per unit volume).
+            plotField(i, j, k, 0) = -tau_xz;
+            plotField(i, j, k, 1) = -tau_yz;
+            plotField(i, j, k, 2) = 0.0;
+
+            src_term(i, j, k, 0) -= (tau_xz * dx[0] * dx[1]) / dV;
+            src_term(i, j, k, 1) -= (tau_yz * dx[1] * dx[1]) / dV;
```
AMLattanzi, line 209: Naïve question after just glancing at this; why is the surface are in L209 `dx[1]*dx[1]` instead of `dx[0]*dx[1]`?

PR #485, opened 2021-09-15
Enable OpenFAST restart from checkpoint files

gantech:
Not sure if this pull request is ready to merge. But this enables both writing of OpenFAST checkpoint files consistent with amr-wind's checkpoint files, as well as the ability to read OpenFAST checkpoint files without the need for replay capability.

PR #807, opened 2023-03-22
Sync OpenFAST restart PR to current amr-wind 

ndevelder:
This is a sync to latest amr-wind of https://github.com/Exawind/amr-wind/pull/485

Planning to test asap, will post results.

### DIFF
--- a/amr-wind/wind_energy/actuator/turbine/fast/FastIface.cpp
+++ b/amr-wind/wind_energy/actuator/turbine/fast/FastIface.cpp
@@ -246,7 +246,7 @@ void FastIface::fast_init_turbine(FastTurbine& fi)
     // integral multiple of CFD timestep
     double dt_err =
         fi.dt_cfd / (static_cast<double>(fi.num_substeps) * fi.dt_fast) - 1.0;
-    if (dt_err > 1.0e-4) {
+    if (dt_err > 1.0e-12) {
         amrex::Abort(
             "FastIFace: OpenFAST timestep is not an integral "
             "multiple of CFD timestep");
--- a/amr-wind/wind_energy/actuator/turbine/fast/FastIface.cpp
+++ b/amr-wind/wind_energy/actuator/turbine/fast/FastIface.cpp
@@ -160,7 +160,7 @@ void FastIface::advance_turbine(const int local_id)
         fast_func(FAST_OpFM_Step, &fi.tid_local);
     }
 
-    if ( ( fi.time_index / fi.num_substeps) % fi.chkpt_interval == 0) {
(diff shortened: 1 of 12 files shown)


ndevelder:
After first round of testing, this is close, we just need to update input flags to be consistent. "sim_mode" and "openfast_sim_mode" both seem to affect whether a case gets initialized, restarted, or replayed. If you want to use this PR right now, recommend using both in your input file.

ndevelder:
This PR adds restarts that work as follows:

Note: current setup creates openfast .chkp file for every amr-wind restart slice. 

1. Change your `time.start_time` to the restart time
2. Set `io.restart_file` to your amr-wind restart slice directory
3. Set `Actuator.T1.openfast_sim_mode = restart` for each turbine (i.e. T1 in this case)
4. Set `Actuator.T1.openfast_restart_file` as the openfast checkpoint file. Note that this must be a relative path from your amr-wind case root, and the ".chkp" must be removed from the filename
5. Set `Actuator.T1.openfast_start_time` to the restart time and double check that `Actuator.T1.openfast_stop_time` is ok

lawrenceccheung (approved):
Thanks @ndevelder, we definitely need this capability.

ndevelder:
And big thanks to @gantech for making it happen in the first place!

amr-wind/wind_energy/actuator/turbine/fast/FastIface.cpp
```diff
@@ -295,10 +301,55 @@ void FastIface::fast_replay_turbine(FastTurbine& fi)
...
+
+    // Determine the number of substeps for FAST per CFD timestep
+    fi.num_substeps = static_cast<int>(std::floor(fi.dt_cfd / fi.dt_fast));
+
+    AMREX_ALWAYS_ASSERT(fi.num_substeps > 0);
+    // Check that the time step sizes are consistent and FAST advances at an
+    // integral multiple of CFD timestep
+    double dt_err =
+        fi.dt_cfd / (static_cast<double>(fi.num_substeps) * fi.dt_fast) - 1.0;
+    if (dt_err > 1.0e-4) {
```
psakievich, line 348: `1e-4` seems kind of generous.  We use `1e-12` in nalu-wind see [here](https://github.com/Exawind/nalu-wind/blob/531b51826358b6f1fe306a5f71ade571e0addf36/src/aero/actuator/ActuatorBulkFAST.C#L69-L70).  IIRC this was due to precision issue @lawrenceccheung had early in his time at SNL where the ratio was close, but not quite exact and it led to really weird errors.

psakievich (approved):
LGTM. I would suggest changing your ratio tolerance, but I won't hold this up over it.

gantech (approved):
Thanks for getting this to the finish line.

PR #810, opened 2023-03-30
WIP boundary plane multilevel feature

michaeljbrazell:
still debugging this and could use some help if anyone has some spare time tomorrow


marchdf:
@michaeljbrazell what is needed to push this through? Anything I can help with?

michaeljbrazell:
This still needs debugging even for a single level. I also hacked in the high side by subtracting a small number. AMReX didn't allow interpolating on the high side face... just need to shift by an index inside AMReX

PR #812, opened 2023-04-04
LES Turbulence unit tests

mbkuhn:


unit_tests/turbulence/test_turbulence_LES.cpp
```diff
@@ -0,0 +1,272 @@
...
+
+    // Create turbulence model
+    sim().create_turbulence_model();
+    // Get turbulence model
+    auto& tmodel = sim().turbulence_model();
+
+    // Get coefficients
+    auto model_dict = tmodel.model_coeffs();
+
+    for (const std::pair<const std::string, const amrex::Real> n : model_dict) {
```
gdeskos, line 130: Not sure why there are two constants

unit_tests/turbulence/test_turbulence_LES.cpp
```diff
@@ -0,0 +1,272 @@
...
+        // Only a single model parameter, Cs
+        EXPECT_EQ(n.first, "Cs");
+        EXPECT_EQ(n.second, Cs);
+    }
+
+    // Constants for fields
+    const amrex::Real srate = 0.5;
```
gdeskos, line 137: Why is the strain rate constant? Does that mean we only test constant gradients?There is nothing wrong with it, I just wonder if we overlook things by doing that.

unit_tests/turbulence/test_turbulence_LES.cpp
mbkuhn: there's only one constant. n.first is the name of the constant, and n.second is the value of the constant.

unit_tests/turbulence/test_turbulence_LES.cpp
mbkuhn: yes, and I'm not exactly testing all the gradients within the strainrate. I can make it more complicated, but I wasn't sure if it was necessary to test more of fvm/strainrate.H. The smagorinsky functionality just calls the strainrate function, and that's how the gradients end up in the calculation.

gdeskos (approved):

PR #814, opened 2023-04-05
Remove projection

tonyinme:
Bug fix to remove velocity projection of the relative velocity


psakievich (changes requested):
@tonyinme let's remove the `tmat` variable completely from the FLLC since it isn't used anymore.

psakievich (approved):

PR #817, opened 2023-04-07
set up poiseuille reg test

mbkuhn:
shown to be second order (spatial), which verifies the behavior of no_slip_wall
image


mbkuhn:
For reproducibility, these are the timesteps used and times sampled

image

gdeskos (approved):

PR #823, opened 2023-04-14
Rayleigh damping per direction

mbkuhn:
* ability to turn off/on different directions of rayleigh forcing
* defaults to all on
* unit test adjusted to check this functionality
* helpful comments added to reg test input file


mchurchf (approved):
This looks great, Michael!  Good job and thanks for doing this.

gantech (approved):

PR #827, opened 2023-04-19
initialize a profile for TKE

mbkuhn:


gantech (approved):

mbkuhn:
BergEtAl2020_tkequote

@lawrenceccheung thank you for sharing. I'm simply trying to match the conditions in Berg et al. (2020) shown here. As it turns out, when I checked Beare et al, there should actually be a power of 3 with this expression. So I'll add that.

gdeskos:
> Hi @mbkuhn,
> 
> The question about initializing the SFS TKE field came up in the TCF project, particularly when we were considering what happens in a RANS model of an ABL. If you're interested, you can take a look at what we did here: https://github.com/lawrenceccheung/amr-wind/blob/cd26a0250d2e1058a323983a20bd3d6ea29be756/amr-wind/wind_energy/ABLFieldInit.cpp#L223-L256
> 
> @ndevelder might also be able to comment on what we did. This never made it into the main branch, but this part could be worth resurrecting.
> 
> Lawrence

This is great @lawrenceccheung , but again why is it sitting on a separate branch for so long instead of doing a PR and making it part of the main branch?

lawrenceccheung:
> 
> This is great @lawrenceccheung , but again why is it sitting on a separate branch for so long instead of making it a PR and then part of the main branch?

That is a great question @gdeskos!  There are some good issues regarding the results of that work which we should discuss offline...

PR #829, opened 2023-04-19
Add implementation for AMD subgrid stress model

moprak-nrel:


amr-wind/turbulence/LES/AMD.cpp
```diff
@@ -0,0 +1,400 @@
...
+template <typename Transport>
+AMD<Transport>::AMD(CFDSim& sim)
+    : TurbModelBase<Transport>(sim)
+    , m_vel(sim.repo().get_field("velocity"))
+    , m_temperature(sim.repo().get_field("temperature"))
+    , m_rho(sim.repo().get_field("density"))
+{
+    auto& phy_mgr = this->m_sim.physics_manager();
+    if (!phy_mgr.contains("ABL"))
+        amrex::Abort("AMD model only works with ABL physics");
```
marchdf, line 22: Is this true? Is it easy to lift this?

moprak-nrel: I have changed the implementation to be conditional on ABL physics instead of explicitly requiring it. However I'm not sure if having an `alphaeff` function will affect non ABL runs. Should I change that to be 0 explicitly without ABL physics?

marchdf: hmmmm I wouldn't want to break anything... so I think that might be safer.

marchdf (approved):

PR #831, opened 2023-04-20
Abl stats energy budget

mbkuhn:
4 terms to be output from the TKE equation:
1. Shear production
2. Buoyancy production
3. Dissipation
4. Diffusion

Dissipation was already calculated in the code; now it's saved so it can be output. Diffusion is difficult to compute directly because of the implicit solve, so it is calculated by subtracting the other terms from the equation.

### DIFF
--- a/amr-wind/equation_systems/tke/source_terms/KsgsM84Src.cpp
+++ b/amr-wind/equation_systems/tke/source_terms/KsgsM84Src.cpp
@@ -10,6 +10,7 @@ KsgsM84Src::KsgsM84Src(const CFDSim& sim)
     : m_turb_lscale(sim.repo().get_field("turb_lscale"))
     , m_shear_prod(sim.repo().get_field("shear_prod"))
     , m_buoy_prod(sim.repo().get_field("buoy_prod"))
+    , m_dissip(sim.repo().get_field("dissipation"))
     , m_tke(sim.repo().get_field("tke"))
 {
     AMREX_ALWAYS_ASSERT(sim.turbulence_model().model_name() == "OneEqKsgsM84");
@@ -30,6 +31,7 @@ void KsgsM84Src::operator()(
     const auto& tlscale_arr = (this->m_turb_lscale)(lev).array(mfi);
     const auto& shear_prod_arr = (this->m_shear_prod)(lev).array(mfi);
     const auto& buoy_prod_arr = (this->m_buoy_prod)(lev).array(mfi);
+    const auto& dissip_arr = (this->m_dissip)(lev).array(mfi);
     const auto& tke_arr = (this->m_tke)(lev).array(mfi);
     const amrex::Real Ceps = this->m_Ceps;
     const amrex::Real CepsGround = this->m_CepsGround;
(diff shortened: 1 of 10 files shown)


amr-wind/equation_systems/tke/source_terms/KsgsM84Src.cpp
```diff
@@ -59,13 +61,13 @@ void KsgsM84Src::operator()(
...
                 amrex::ParallelFor(
                     blo, [=] AMREX_GPU_DEVICE(int i, int j, int k) noexcept {
-                        amrex::Real ceps_local =
-                            (Ceps / 0.93) *
-                            (0.19 + (0.74 * tlscale_arr(i, j, k) / ds));
-                        src_term(i, j, k) += (ceps_local - CepsGround) *
-                                             std::sqrt(tke_arr(i, j, k)) *
-                                             tke_arr(i, j, k) /
-                                             tlscale_arr(i, j, k);
+                        // Remove previous dissipation term
```
gdeskos, line 64: I don't understand why you remove and add a new dissipation term ...

mbkuhn: If you look at the original code, that's what it's doing in lines 65-68. The original term which is added globally in lines 49-50 is -ceps_local * sqrt(tke) * tke / tlscale. When the wall model is on, the src term is modified in the wall-adjacent cell by first adding ceps_local * sqrt(tke) * tke / tlscale and then adding -CepsGround * sqrt(tke) * tke / tlscale. That's still what I'm doing in the new implementation, but I'm using the new functions I introduced (calc_dissip) and correcting what is stored as the dissipation term.

gantech (approved):
That was fast! Well done! I couldn't see if you checked whether the turbulence model KsgsM84 was chosen somewhere. This may not work for Smag model, but I say we fix this at a later date.

amr-wind/equation_systems/tke/source_terms/KsgsM84Src.cpp
gantech: I agree! 
Screen Shot 2023-04-21 at 2 24 53 PM

amr-wind/wind_energy/ABLStats.cpp
```diff
@@ -140,6 +141,56 @@ void ABLStats::calc_sfs_stress_avgs(
...
+            const auto& shear_prod_arr = shear_prod(lev).const_array(mfi);
+            const auto& dissipation_arr = dissipation(lev).const_array(mfi);
+            const auto& tke_arr = tke(lev).const_array(mfi);
+            const auto& tke_old_arr = tke_old(lev).const_array(mfi);
+            const auto& conv_arr = conv_term(lev).const_array(mfi);
+
+            amrex::ParallelFor(
+                bx, [=] AMREX_GPU_DEVICE(int i, int j, int k) noexcept {
+                    diffusion_arr(i, j, k) =
+                        (tke_arr(i, j, k) - tke_old_arr(i, j, k)) / dt -
```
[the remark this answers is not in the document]
mbkuhn: Replying to your above comment related to checking for the KsgsM84 - at this point, I just leave the correctness in usage up to the user. If the tke equation is not used (e.g. in the case of Smagorinksy), but the energy budget is requested, the code will abort.

PR #833, opened 2023-04-25
Added time a start time for the fllc

tonyinme:
This PR adds a new option in the fllc to start after a given amount of time. 
By default, the fllc will start at time zero, but this can now be changed in the input file.


psakievich (commented):
@tonyinme should we add this to the fixed wing case too? Would it be possible to make the execution just skip if the start time is greater than the current time?  Then all the changes would be isolated to the FLLC code and we wouldn't need to update the turbine and fixed wing ops separately.

tonyinme:
That's a good point @psakievich .
How can we best achieve this?
I am thinking of adding an if statement [here](https://github.com/Exawind/amr-wind/blob/22e18d86e0b89135b81a9086c2ed0af71026f301/amr-wind/wind_energy/actuator/FLLC.cpp#L8)?
`if (time.current_time() > tdata.fllc_start_time) return;` 

Would that work or should we do something else?

psakievich:
I would put it [here](https://github.com/Exawind/amr-wind/blob/22e18d86e0b89135b81a9086c2ed0af71026f301/amr-wind/wind_energy/actuator/FLLCOp.H#L20) and add a time value to the operator call if that works out. I haven't looked to closely so there might be some other complication I'm missing.  I think it is okay to initialize the FLLC data structure, you just don't want to start computing it is all.

psakievich (approved):
For completion of the conversation, @tonyinme and I discussed this and decided the interface is going to be adjusted either way by requiring propagation of the current time. So the design I suggested is not necessarily more optimal.

tonyinme:
The latest changes work with both fixed wing and openfast turbine.
And added documentation.
@psakievich , let me know if we need anything else.

tonyinme:
@psakievich , I thought this PR had been merged. Can we merge?

PR #836, opened 2023-04-27
Implementing a wall modeled LES channel flow

marchdf:
This PR adds:
- an analytical Smagorinsky channel test
- a constant shear stress wall model boundary condition 

TODO:
- [x] lift restrictions of wall normal direction, or at least guard against it
- [x] add regtest
- [x] document convergence run

FYI: @asalmgren @mbkuhn @moprak-nrel @amlattanzi

Co-authored-by: moprak-nrel <120606615+moprak-nrel@users.noreply.github.com>


mbkuhn:
> I hate to ask but... are you going to do this for `vorticity`, `laplacian`, `q-critierion`, and `gradient` as well?

This PR will only address the operators in ABLStats and turbulence modeling. Which means only gradient and strainrate. But the others can be addressed in a future PR.

moprak-nrel (approved):

amr-wind/boundary_conditions/wall_models/WallFunction.cpp
```diff
@@ -0,0 +1,125 @@
...
+                        varr(i, j, k - 1, 1) = 0.0;
+                    });
+            }
+
+            if (bx.bigEnd(idim) == domain.bigEnd(idim) &&
+                velocity.bc_type()[zhi] == BC::wall_model) {
+                amrex::ParallelFor(
+                    amrex::bdryHi(bx, idim),
+                    [=] AMREX_GPU_DEVICE(int i, int j, int k) noexcept {
+                        const amrex::Real mu = eta(i, j, k - 1);
```
mbkuhn, line 104: I realize this is how it's coded now, but I don't think the indices are correct. The bdryHi of the bx gives the highest k inside the domain, as far as I know. This would be the k at the highest cell center. Then mu should be at k, and varr should be set at k + 1. I may be wrong, but could you check this?

marchdf: `amrex::bdryHi(bx, idim)` gives `((0,0,32) (15,15,32) (0,0,1))` so I think this is fine.

mbkuhn (approved):

PR #840, opened 2023-05-01
Align starting index of plt and chk

rthedin:
Checkpoint starting index was not being taken into consideration when a restarted run was executed. It resulted in the first checkpoint index not being the starting index plus the checkpoint interval. With the recent OpenFAST restart capability (PR #807), that resulted in OpenFAST checkpoints that were _not_ aligned with AMR-Wind checkpoints, rendering the restart capability worthless in this scenario. Related: PR #488.

### DIFF
--- a/amr-wind/core/SimTime.cpp
+++ b/amr-wind/core/SimTime.cpp
@@ -234,6 +234,9 @@ void SimTime::set_restart_time(int tidx, amrex::Real time)
     m_new_time = time;
     m_cur_time = time;
     m_start_time = time;
+
+    m_chkpt_start_index = m_start_time_index;
+    m_plt_start_index   = m_start_time_index;
 }
 
 } // namespace amr_wind


marchdf:
@rthedin I am not sure I understand what this fixes? Can you explain a bit more?

rthedin:
Sure thing. It's easier if I give you an example.
I have a precursor with variable time step, so my final checkpoint was a "weird" number, in my case, `chk76826`. On my inflow-outflow case with turbines, I start at this checkpoint and ask for, say, save a checkpoint at every 12000 time steps. Because of the amount of OpenFAST sub-time steps within each AMR-Wind time step, I have OpenFAST checkpoints saved at every 60000 OpenFAST time steps.

What was happening is that we end up with checkpoints saved at 84000, 96000, 108000, etc, as per
https://github.com/Exawind/amr-wind/blob/257c13c1a634d841627a04f83f1923b2a8556ca5/amr-wind/core/SimTime.cpp#L208-L213
The variable `m_chkpt_start_index` is set to zero. The OpenFAST checkpoints are saved at 60000, 120000, 180000, etc.

The problem is that we would expect the AMR-Wind checkpoints to be at 88826, 100826, etc. The starting index (76826) is being ignored, and a multiple of the saving interval is being selected as per the code above. This PR only really updates the checkpoint and plt starting indices.
The OpenFAST checkpoints are aligned with the nonexistent AMR-Wind checkpoints 88826, 100826, etc, and NOT with the 84000, 96000, etc.

marchdf:
ah that makes much more sense. Thank you.

rthedin:
I had no idea about the `checkpoint_start` option, so I did not try it out. The changes I'm suggesting here works in the cases where `checkpoint_start` would need to be specified, but also works in other more common use cases.

As those two extra lines being where you suggested, the variable `m_start_time_index` is not actually initialized there for a restarted run. I tried there first and it was not doing what we would expect it to.

marchdf:
I think `checkpoint_start` could be useful if you only want to start saving data after a certain amount of time?

rthedin:
The proposed changes are not overwritten by what is in the input file; in fact, it's the opposite for restarted runs. In these cases, whatever is on the input file as `time.checkpoint_start` is ignored. I just ran a test case to confirm. Which now that I'm understanding better the variables and inputs, I realize it is not a good solution either. The way it is right now with these changes, it means that the user can control the `checkpoint_start` and `plot_start` if not a restart; otherwise the user has no way to control it and those inputs are overwritten.

I think this may require changes that are not as simple as the ones I've made. The `parse_parameters` is called _before_ the `set_restart_time`, so that `m_start_time_index`
https://github.com/Exawind/amr-wind/blob/c9e76e5a3fa00d45605e73b950ae081262ef70c3/amr-wind/core/SimTime.cpp#L232
hasn't been declared when `parse_parameters` is called.

What could be done is parse again `plot_start` and `checkpoint_start` at the `set_restart_tile` and re-set them as needed, but I understand that might not be ideal. Something along these lines?
```cpp
amrex::ParmParse pp("time");
pp.query("plot_start", m_plt_start_index);
pp.query("checkpoint_start", m_chkpt_start_index);

if (m_chkpt_start_index < m_start_time_index) {
	m_chkpt_start_index = m_start_time_index;
}

if (m_plt_start_index < m_start_time_index) {
	m_plt_start_index = m_start_time_index;
}
```

ashesh2512:
Okay, I understand the PR here - the request is to have the default behavior be such that the checkpoint and plot file are dumped based on the the restart file index instead of being a multiple of the plot or checkpoint interval - closing this PR, have opened an issue - #845

PR #842, opened 2023-05-02
Update SST and SST IDDES

marchdf:
Update the RANS and hybrid RANS-LES models:
- adding (optional) SUST variant to SST and IDDES, similar to what is done in Nalu-WInd
- adding a production limiter consistent with nalu-wind
- adding an output `rans_indicator` field
- upgraded IDDES to not be just an LES model but include the correct blending to RANS.

### DIFF
--- a/amr-wind/turbulence/RANS/KOmegaSST.cpp
+++ b/amr-wind/turbulence/RANS/KOmegaSST.cpp
@@ -198,7 +198,9 @@ void KOmegaSST<Transport>::update_turbulent_viscosity(
                     const amrex::Real diss_amb =
                         beta_star * rho_arr(i, j, k) * sdr_amb * tke_amb;
                     diss_arr(i, j, k) = -beta_star * rho_arr(i, j, k) *
-                                        tke_arr(i, j, k) * sdr_arr(i, j, k) + diss_amb;
+                                            tke_arr(i, j, k) *
+                                            sdr_arr(i, j, k) +
+                                        diss_amb;
 
                     tke_lhs_arr(i, j, k) = 0.5 * beta_star * rho_arr(i, j, k) *
                                            sdr_arr(i, j, k) * deltaT;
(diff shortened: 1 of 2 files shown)


sbidadi9:
@marchdf Can you make SUST an optional feature? For some cases, we want TKE to decay to a certain value at the rotor 

cc: @gantech @ashesh2512

sbidadi9:
@marchdf Can you make the SUST for IDDES an optional feature? For SST, these terms for k and omega are zero but are non-zero for IDDES.

@gantech @ashesh2512

marchdf:
@sbidadi9 I don't follow... if tke_amb and sdr_amb are zero (and they are by default) then `diss_amb` and `sdr_diss_amb` are zero and SUST is not active and you revert to original IDDES. What am I missing?

sbidadi9:
@marchdf Got it. Also, based on running NRELVI cases, the RANS mode does not contribute to the length scale. Hence, the blending terms can be commented out (again).

marchdf:
So I get that they don't make sense in that case. But they make sense to have in general so I am in favor of keeping them.

sbidadi9:
Can the l_RANS term blowup if the code goes unstable?

marchdf:
Possibly? But if the code goes unstable, all bets are off, right?

marchdf:
I haven't noticed that... have you seen that happen?

sbidadi9:
Okay. I have not tried it.

sbidadi9:
@marchdf Could you please try running one of the cases with and without l_RANS computation in AMR-Wind to see if it affects the stability of the solver?

marchdf:
Can this be merged?

sbidadi9:
I have a PR which fixes the SDR blowup issue (https://github.com/Exawind/amr-wind/pull/852). Once approved, we can merge the current one as well.

sbidadi9 (approved):

Issue #845, opened 2023-05-02
Dump checkpoint and plot file based on restart file index


ashesh2512:
the request is to have the default behavior be such that the checkpoint and plot file are dumped based on the the restart file index instead of being a multiple of the plot or checkpoint interval which is the current default behavior - need to make sure nalu-wind behaves the same too

PR #846, opened 2023-05-03
Use start time index as default value for checkpoint, plot, and regrid start upon restart

ashesh2512:
See #840 and #845

Any "more elegant" ideas are welcome. 

P.S. - This is currently inconsistent with Nalu-Wind behavior which acts on multiples of output/restart frequency.

### DIFF
--- a/amr-wind/core/SimTime.cpp
+++ b/amr-wind/core/SimTime.cpp
@@ -230,6 +230,15 @@ void SimTime::set_restart_time(int tidx, amrex::Real time)
 {
     m_time_index = tidx;
     m_start_time_index = tidx;
+    m_regrid_start_index = tidx;
+    m_plt_start_index = tidx;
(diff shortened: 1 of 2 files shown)


marchdf:
Could we flip the default? Make it so the default is the current behavior, consistent with nalu-wind, and allow the user to switch the default? Maybe with just one boolean flag (eg. `use_start_time_index_for_io`, but better named).

gantech:
I feel that the new behavior enabled by this PR is desirable and should be the default. I have discussed with several people including those at GE who have had to scrap large runs because they expected the behavior enabled by this PR, but then had to realize it wasn't. 

On a completely separate effort, we should change nalu-wind's behavior to be consistent with the new behavior enabled by this PR.

marchdf:
gotcha. We should make sure that nalu-wind does this at the same time.

marchdf (approved):

PR #849, opened 2023-05-24
Add verification (channel, burggraf) to documentation

mbkuhn:


mbkuhn:
I included the channel and burggraf flow verification cases, but I didn't add the vortex-dipole case. I could add that in a separate, new validation section if you feel it's necessary.

marchdf (approved):

Issue #784, opened 2023-02-08
Upgrade ADM-Joukowski model to handle rated conditions


lawrenceccheung:
Putting this on the to-do list in case anybody has time.  The current ADM Joukowski model assumes that the turbine is always operating under design load conditions and the constant circulation assumption applies.  However, under off-design conditions the load distributions no longer match the analytical loads. 

Consequently, when the model is running above rated conditions, it will overpredict the power as in this example below:
![image](https://user-images.githubusercontent.com/15526007/217408610-54900f95-0d8b-43a3-a0b5-a7792c397504.png)
(see the [turbine model details](https://github.com/lawrenceccheung/AWAKEN_summit_setup/tree/main/turbines/GE1.79-100)) 

This is a known problem and Sorensen has a suggested fix in the publication here: 
https://wes.copernicus.org/preprints/wes-2022-108/
Basically there is an additional $S_0$ term that should be added to the $u_\theta$ azimuthal velocity distribution:

$$\frac{u_\theta}{U_0} = (\frac{q_0}{x}+S_0 x)g(x) F(x) $$

Hopefully it is a straightforward implementation.

Lawrence

PR #851, opened 2023-05-30
Upgrade Joukowski model to handle rated conditions

lawrenceccheung:
Addresses issue in #784.

I updated the Joukowski model to handle rated conditions, following the formulation in Sorensen (2022): 
https://wes.copernicus.org/preprints/wes-2022-108/

Note: this pull request is in draft state, please do not merge until we run tests to confirm that the Joukowski model behaves properly at rated Ct.

**EDIT**: I added some results, let's review it and see if it's OK to use.

Lawrence

### DIFF
--- a/amr-wind/wind_energy/actuator/disk/Joukowsky.H
+++ b/amr-wind/wind_energy/actuator/disk/Joukowsky.H
@@ -22,6 +22,8 @@ struct JoukowskyData : public DiskBaseData
     // --- Sorenson 2020 equation 10 constants ----
     amrex::Real root_correction_coefficient{2.335};
     amrex::Real root_correction_exponent{4.0};
+    // --- Sorenson 2022 equation 26 constants ----
+    amrex::Real Ct_rated{0.0};
     // --------------------------------------------
     int num_blades{3};
     vs::Vector disk_force{0.0, 0.0, 0.0};
--- a/amr-wind/wind_energy/actuator/disk/Joukowsky.H
+++ b/amr-wind/wind_energy/actuator/disk/Joukowsky.H
@@ -24,6 +24,8 @@ struct JoukowskyData : public DiskBaseData
     amrex::Real root_correction_exponent{4.0};
     // --- Sorenson 2022 equation 26 constants ----
     amrex::Real Ct_rated{0.0};
+    amrex::Real S0_alpha1{-0.20};
+    amrex::Real S0_alpha2{-0.05};
     // --------------------------------------------
     int num_blades{3};
     vs::Vector disk_force{0.0, 0.0, 0.0};
--- a/amr-wind/wind_energy/actuator/disk/Joukowsky.H
+++ b/amr-wind/wind_energy/actuator/disk/Joukowsky.H
@@ -19,11 +19,17 @@ struct JoukowskyData : public DiskBaseData
     amrex::Real vortex_core_size;
     amrex::Real current_cp;
     amrex::Real current_power;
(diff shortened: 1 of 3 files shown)


lawrenceccheung:
I was able to test out the new Joukowski model on a couple of turbines, and able to get them to behave better at rated wind speeds.

For the 2.4-107 model:
image

For the 1.79-100
image

However, please note a couple of things:
- I'm interpreting the $C_{T,r}$ variable (`Ct_rated`) to mean the value of Ct when it's operating in region 2, and not the value of Ct when the turbine hits rated power.   That way the solid body term vanishes in region 2 and power production is unaffected in that part of the power curve:

$S_0 = \alpha(\frac{C_{T,r}-C_T}{C_{T,r}})^\beta$

- I'm using a different value of $\alpha$ compared to Sorensen's paper.  While I don't expect the same parameters to hold across all turbine models, it turns out I had to use a _negative_ $\alpha$ in order to limit the power above rated.  I'm using $\alpha=-0.21$ for $C_T > C_{T,r}$, whereas Sorensen used $\alpha=0.05$.  I'm not sure if there's a typo in the paper or a mistake somewhere in the code.  Or could it also be related to the fact that we switched signs on the rotor rotation direction (https://github.com/Exawind/amr-wind/pull/848)?

Lawrence

lawrenceccheung:
@psakievich It occurred to me that we could also put in a generator efficiency and directly report electrical power generation.   Let me do that in a separate PR (maybe long with some documentation) after this one is approved and merged.

Lawrence

amr-wind/wind_energy/actuator/disk/Joukowsky.H
```diff
@@ -22,6 +22,10 @@ struct JoukowskyData : public DiskBaseData
     // --- Sorenson 2020 equation 10 constants ----
     amrex::Real root_correction_coefficient{2.335};
     amrex::Real root_correction_exponent{4.0};
+    // --- Sorenson 2022 equation 26 constants ----
```
psakievich, line 25: @lawrenceccheung let's add a comment here about the change in sign, and using 0.2 vs 0.8 for the C_t < C_t,r case.

amr-wind/wind_energy/actuator/disk/Joukowsky_ops.H
```diff
@@ -180,20 +204,34 @@ struct ComputeForceOp<Joukowsky, ActSrcDisk>
...
+        const amrex::Real term1 = std::pow(a2 * lambda + a3 * S0, 2.0);
+        const amrex::Real term2 =
+            a1 * (0.5 * Ct - 2.0 * a4 * lambda * S0 - a5 * S0 * S0);
+        const amrex::Real term3 = a2 * lambda + a3 * S0;
 
         const amrex::Real numerator = std::sqrt(term1 + term2) - term3;
 
         const amrex::Real denominator =
-            amrex::max<amrex::Real>(4.0 * a1, machine_eps);
+            amrex::max<amrex::Real>(1.0 * a1, machine_eps);
```
psakievich, line 234: A little pedantic, but we can just remove the coefficient here now.

psakievich (approved):
A couple of minor things to clean up but nothing to stop the merge if required.

amr-wind/wind_energy/actuator/disk/Joukowsky.H
lawrenceccheung: Thanks @psakievich, I added some comments explaining those values in the latest commit.

PR #853, opened 2023-06-01
Modification to make hybrid solver work for prescribed flows

mbkuhn:
put prescribe conditional within dedicated function


mbkuhn:
goes with https://github.com/Exawind/exawind-driver/pull/48

marchdf (approved):

Issue #850, opened 2023-05-25
Compiling with latest OpenFAST version


lawrenceccheung:
I'm getting an error if I compile amr-wind with the latest version of OpenFAST (v3.5+):
```
/projects/wind_uq/lcheung/AMRWindBuilds/hfm.20230525/amr-wind/amr-wind/wind_energy/actuator/turbine/fast/FastIface.cpp: In instantiation of 'void exw_fast::{anonymous}::fast_func(const FType&&, Args ...) [with FType = void(int*, double*, const char*, int*, int*, int*, int*, float*, float*, int*, int*, float*, int*, double*, int*, int*, int*, exw_fast::OpFM_InputType*, exw_fast::OpFM_OutputType*, exw_fast::SC_DX_InputType*, exw_fast::SC_DX_OutputType*, int*, char*); Args = {int*, double*, char*, int*, int*, int*, int*, float*, float*, int*, int*, float*, int*, double*, int*, int*, exw_fast::OpFM_InputType*, exw_fast::OpFM_OutputType*, exw_fast::SC_DX_InputType*, exw_fast::SC_DX_OutputType*}]':
/projects/wind_uq/lcheung/AMRWindBuilds/hfm.20230525/amr-wind/amr-wind/wind_energy/actuator/turbine/fast/FastIface.cpp:216:77:   required from here
/projects/wind_uq/lcheung/AMRWindBuilds/hfm.20230525/amr-wind/amr-wind/wind_energy/actuator/turbine/fast/FastIface.cpp:22:9: error: cannot convert 'exw_fast::OpFM_InputType*' to 'int*' in argument passing
   22 |     func(std::forward<Args>(args)..., &ierr, err_msg);
      |     ~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
make[2]: *** [CMakeFiles/amrwind_obj.dir/amr-wind/wind_energy/actuator/turbine/fast/FastIface.cpp.o] Error 1
```
This occurs with both gcc and intel compilers, at least on Sandia HPC machines.  However, if I switch to OpenFAST v3.4 (https://github.com/OpenFAST/openfast/commit/e8ec53f9c7f9d3f6a13bfb61dba12a0ca04d8a2f), this problem goes away.

It will take some time to bisect the problem and find out the problem, but perhaps one thing to look at is this commit:
[...]

PR #858, opened 2023-06-16
Changes to OpenFAST interface to compile with OpenFAST 3.5.0

shenpai35:
Pass the uniform distribution option to account for changes in this OpenFAST pull request https://github.com/OpenFAST/openfast/pull/1493

### DIFF
--- a/amr-wind/wind_energy/actuator/turbine/fast/FastIface.cpp
+++ b/amr-wind/wind_energy/actuator/turbine/fast/FastIface.cpp
@@ -206,6 +206,7 @@ void FastIface::fast_init_turbine(FastTurbine& fi)
     int abort_lev;
     char inp_file[fast_strlen()];
     copy_filename(fi.input_file, inp_file);
+    int chord_cluster_type = 0;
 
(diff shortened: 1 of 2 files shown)


psakievich (changes requested):
@gantech have you tried compiling `openfast@3.4` with these changes? Unfortunately I suspect this change will break the ability to compile with anything older than `3.5` since the function signature changed.  We will need to do if/def's like we did for the `SC_DX` change to make sure the code compiles with all versions.

lawrenceccheung:
Is it worth considering pinning amr-wind to run with specific versions of OpenFAST, and then upgrading versions of OpenFAST after we've checked out there aren't any API changes/behavioral changes?  @psakievich -- we could possibly do this with spack-manager, although people compiling manually still might face issues like https://github.com/Exawind/amr-wind/issues/850.

Lawrence

gantech:
Man.. an intermediate class between OpenFAST and CFD solvers can't get here soon enough! Ok.. let's close this for now and reconfigure this to have specific version flags in CMake for now?

PR #862, opened 2023-06-21
Fix for ABL inflow default BC

marchdf:
For non-boundary IO planes, ABLFillInflow defaults to a no-op, leaving the BC data as is. This PR changes so that ABLFIllInflow inherits from BCDirichlet and fills 0 into non-inflow BCs.

### DIFF
--- a/amr-wind/wind_energy/ABLFillInflow.H
+++ b/amr-wind/wind_energy/ABLFillInflow.H
@@ -15,7 +15,7 @@ namespace amr_wind {
  *
  *  \sa ABLBoundaryPlane
  */
-class ABLFillInflow : public FieldFillPatchOps<FieldBCNoOp>
+class ABLFillInflow : public FieldFillPatchOps<FieldBCDirichlet>
 {
 public:
     ABLFillInflow(
--- a/amr-wind/wind_energy/ABLFillInflow.cpp
+++ b/amr-wind/wind_energy/ABLFillInflow.cpp
@@ -7,7 +7,7 @@ ABLFillInflow::ABLFillInflow(
     const amrex::AmrCore& mesh,
     const SimTime& time,
     const ABLBoundaryPlane& bndry_plane)
-    : FieldFillPatchOps<FieldBCNoOp>(
+    : FieldFillPatchOps<FieldBCDirichlet>(
           field, mesh, time, FieldInterpolator::CellConsLinear)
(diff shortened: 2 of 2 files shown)


marchdf:
Ok I think I've convinced myself that this is the right fix. 

The first thing to recognize is that, by default, typically, we create a fillpatch op for velocity that is BCDirichlet: https://github.com/Exawind/amr-wind/blob/main/amr-wind/equation_systems/PDEHelpers.H#L68

Except that, in the current version of the code, ABLFIllInflow inherits from `FieldFillPatchOps<FieldBCNoOp>`, which means that the fillpatch operations have a cphysbc or fphysbc that don't do anything. and, therefore, leave NaNs in the ghost cells (in debug mode. In non-debug mode, those are just garbage uninitialized values).

So in the case of `abl_bndry_input_amr` reg test, we have a fine level hitting the bottom BC, which is a wall. When that level gets fillpatched, it has NaNs in the ghost cells on zlo (again, only when using ABLFIllInflow, because that case noops _all_ BCs). The fix of this PR is to actually provide fillpatch with cphysbc and fphysbc functors that actually make sense in that they put _something_ in the ghost cells. That _something_ is by default 0 when using BCDirichlet. 

This will introduce diffs. For example, the `abl_bndry_input_amr` reg test shows diffs where you would expect them (at the CF interface, at the wall). This is x-velocity:
Screen Shot 2023-06-22 at 09 49 15


This is a bug that doesn't just affect GPUs. You can easily find it with a CPU debug build and trapping NaNs. As a matter of fact, it's been on the dashboard for a long time (https://my.cdash.org/test/80682734) and we've taken half-hearted stabs at fixing it in the past but never made it a priority. Lesson learned I guess? 

Paul has confirmed that this PR fixes the issues he was seeing on Frontier.

ashesh2512 (approved):

mbkuhn (approved):

marchdf:
This seals it for me. I ran a precursor sim and then restarted from that and used that inflow boundary planes. I did that for the main branch and this PR. Though the fields show diffs in fcompare (which I expect), they are visually almost identical. No weirdnesses are apparent in the refined region and the inflow data is coming in the same.  This comparison is done 5000s post-restart.

Screen Shot 2023-06-22 at 13 36 04

PaulMullowney:
This is great news!

PR #867, opened 2023-06-26
Add a central convection scheme

moprak-nrel:
Creating a central scheme for convection

Todos:
- [x] Testing

### DIFF
--- a/amr-wind/convection/Godunov.H
+++ b/amr-wind/convection/Godunov.H
@@ -10,7 +10,17 @@
 
 namespace godunov {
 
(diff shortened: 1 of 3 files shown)


gdeskos:
Prakash @moprak-nrel this only addresses a small part of the flux discretization. Upwinding is subsequently applied multiple times before calculating the final advection term. I am not saying that you should not try this out, I am just saying that in order to achieve a pure second-order, central-difference scheme you may need to change other parts of the code too!

moprak-nrel:
Thanks @gdeskos, do you happen to know the LOCs where this upwinding is occurring? Or is this just from 2nd order Godunov?

PR #868, opened 2023-06-29
Actuators: sample mac velocity instead of cell-centered velocity

mbkuhn:
goal of this is to sample velocity at n-1/2, advance timestep by a dt, apply forces at n+1/2

### DIFF
--- a/amr-wind/boundary_conditions/wall_models/LogLaw.H
+++ b/amr-wind/boundary_conditions/wall_models/LogLaw.H
@@ -1,0 +1,59 @@
+#ifndef LogLaw_H
+#define LogLaw_H
+
+#include "AMReX_AmrCore.H"
+#include <AMReX.H>
+namespace amr_wind {
+struct LogLaw
+{
+    /*
+     * A simple wall model that sets the wall-shear stress
+     * based on computing u_tau given the horizontal velocity
+     * magnitude at a zref. This is akin to an explicit non-linear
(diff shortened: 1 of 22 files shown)


tonyinme:
Thanks @mbkuhn ! 
@gdeskos and @mchurchf , this is the latest development/test for the ALM based on our conversations.

tonyinme:
Hi all, just wanted to add that this PR has improved the advanced ALM results significantly. 
The conversations with the team have suggested that this should become the standard way of implementing ALM body forces (velocity sampled at n-1/2 and forces implemented at n+1/2).

mbkuhn:
Because we haven't yet tested this for other actuator types (e.g. disk), I made the new sampling approach an option that is turned on by an input argument, Actuator.sample_vel_nmhalf = true

tonyinme (commented):
These changes look great.
I think it is best to set the mac velocity as a default, because it is actually a bug fix. 
We might want/need to remove the legacy code eventually.

lawrenceccheung:
My apologies I should have added this earlier, can this also be the default for `TurbineFastDisk`?  We believe it could also be important for OpenFAST ADM cases.

Lawrence

mbkuhn:
At this point, I think it has the default behavior as desired.

tonyinme:
Hey @mbkuhn , that's excellent!
You can add the option to the documentation, and merge.

PR #872, opened 2023-07-11
Remove 90 deg Latitude Constraint from Hurricane Forcing

mchurchf:
Removing the constraint that latitude must be 90 degrees for the hurricane forcing.


gdeskos (commented):
Why did you remove HurricaneForcing?

gdeskos (approved):

PR #873, opened 2023-07-12
Limiting option for IDDES-SDR production term

sbidadi9:
A user-defined clipping parameter option(epsilon) is introduced into the omega production term:
image

It can be set as follows:
`KOmegaSSTIDDES_coeffs.sdr_prod_clip_factor = 10.0`

Advantages:
- Using values less than 10.0 (default) provides additional stability to the solver by lowering the threshold of production term. 
- Based on numerical experiments, the clipping needs to be activated for deep stall turbine simulations combined with highly refined meshes. 
- Activating the clipping did not have an effect on the magnitude of kinetic energy production and overall solution.

Recommendation:
- For runs with convergence issues, the recommendation is to repeatedly lower the factor by an order of magnitude until SDR residuals converge.

### DIFF
--- a/amr-wind/turbulence/RANS/KOmegaSST.H
+++ b/amr-wind/turbulence/RANS/KOmegaSST.H
@@ -94,8 +94,6 @@ protected:
     amrex::Real m_buoyancy_factor = 0.0;
     amrex::Real m_sigma_t{0.85};
     amrex::Vector<amrex::Real> m_gravity{{0.0, 0.0, -9.81}};
-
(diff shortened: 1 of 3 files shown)


ashesh2512 (approved):

ashesh2512:
@sbidadi9 This looks good. Could you please update the branch to be up to date with the main branch?

sbidadi9:
> @sbidadi9 This looks good. Could you please update the branch to be up to date with the main branch?

@ashesh2512 Branch has been updated and merged.

PR #874, opened 2023-07-12
using mac vel within wall function.

mbkuhn:
* enables use of n+1/2 velocity when it's called for the viscous terms

### DIFF
--- a/amr-wind/wind_energy/ABLWallFunction.cpp
+++ b/amr-wind/wind_energy/ABLWallFunction.cpp
@@ -211,7 +211,7 @@ void ABLVelWallFunc::wall_model(
                         const amrex::Real uu =
                             0.5 * (umac_arr(i, j, k) + umac_arr(i + 1, j, k));
                         const amrex::Real vv =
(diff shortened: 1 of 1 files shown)


amr-wind/wind_energy/ABLWallFunction.cpp
```diff
@@ -210,9 +210,6 @@ void ABLVelWallFunc::wall_model(
                         // const amrex::Real vv = vold_arr(i, j, k, 1);
                         const amrex::Real uu =
                             0.5 * (umac_arr(i, j, k) + umac_arr(i + 1, j, k));
-                        // Is this reasonable, to interpolate using the wall
-                        // value? Or should I extrapolate using the two points
-                        // above the wall?
                         const amrex::Real vv =
                             0.5 * (vmac_arr(i, j, k) + umac_arr(i, j + 1, k));
```
gdeskos, line 214: Please make sure that you don't mix vmac_arr with umac_arr !

mbkuhn: good catch!

PR #881, opened 2023-07-19
Time intervals for plt and chkpt output

mbkuhn:
* instead of outputting at specified intervals of timesteps, this enables outputting at specified intervals of time
* this is an enabling feature for adaptive dt simulations and is key to making normal-looking videos when dt varies significantly with time (such as with breaking waves)
* the implementation includes the option to enforce dt to adhere to the plotting or checkpointing time intervals specified; otherwise the program will output the first plotfile or checkpoint after the time interval is passed
* also allows the maximum growth of dt (in adaptive dt simulations) to be an input parameter
* includes unit tests and abort statements for combinations of parameters that do not make sense


amr-wind/core/SimTime.cpp
```diff
@@ -201,15 +257,23 @@ bool SimTime::do_regrid() const
 bool SimTime::write_plot_file() const
 {
     return (
-        (m_plt_interval > 0) &&
-        ((m_time_index - m_plt_start_index) % m_plt_interval == 0));
+        ((m_plt_interval > 0) &&
+         ((m_time_index - m_plt_start_index) % m_plt_interval == 0)) ||
```
ashesh2512, line 261: Is it worthwhile adding a check to make sure that both conditions are not being activated in the same simulation? Are there scenarios that we foresee otherwise?

amr-wind/core/SimTime.cpp
```diff
@@ -34,6 +39,38 @@ void SimTime::parse_parameters()
...
+            "timestep- and time-based plotting should not be used together; "
+            "please only specify one.");
+    }
+
+    if (m_chkpt_interval > 0 && m_chkpt_t_interval > 0.0) {
+        amrex::Abort(
+            "checkpoint_interval and checkpoint_time_interval are both "
+            "specified. timestep- and time-based checkpointing should not be "
+            "used together; please only specify one.");
+    }
```
mbkuhn, line 55: @ashesh2512 these two abort statements make sure that timestep- and time-based intervals are not being used together.

amr-wind/core/SimTime.H
```diff
@@ -127,6 +129,18 @@ public:
...
+    AMREX_FORCE_INLINE
+    amrex::Real get_enforced_dt_for_output(
+        const amrex::Real dt,
+        const amrex::Real cur_time,
+        const amrex::Real interval)
+    {
+        // Tolerance for being slightly under is .1% of requested interval
+        return std::min(
+            dt,
+            (std::floor(cur_time / interval + 1e-3) + 1) * interval - cur_time);
```
marchdf, line 141: do you want to make this a tunable tolerance? just in case?

ashesh2512 (approved):

mbkuhn:
> @mbkuhn have you looked into `amr.plot_per`? https://github.com/AMReX-Codes/amrex/blob/development/Src/Amr/AMReX_Amr.H#L122 That might be more relevant to an AmrLevel than AmrCore but it might be worth looking at how it is done there and follow similar thing?

Thanks for sharing this example! My setup of the dt enforcement is very similar to the amrex code. For the plotfile and checkpoint outputs, I mimicked part of a code from my PhD research group: https://github.com/desjardi/NGA2/blob/master/src/libraries/event_class.f90#L61 . Both are pretty similar conceptually.

PR #882, opened 2023-07-19
Changed velocity sampling to be consistent with standard ALM definition

tonyinme:
These changes improve the velocity sampling for the ALM in fixed wing applciations.
The previous definition rotated the velocity vector using quaternions.
This new implementation defines a local coordinate system and projects the velocity onto the local coordinate system.
The new implementation follows the implementation of SOWFA.


amr-wind/wind_energy/actuator/wing/wing_ops.H
```diff
@@ -118,14 +118,25 @@ struct ComputeForceOp<
...
-            wvel.y() = 0.0;
 
-            const auto vmag = vs::mag(wvel);
-            const auto aoa = std::atan2(wvel.z(), wvel.x());
+            // Build the local reference frame
+            vs::Vector wspan = wdata.end - wdata.start;
+            wspan = wspan.unit();
+
+            // Use the global coord to orient the blade
+            // This works only for inflow in the x direction
```
psakievich, line 127: If it will only work for a single direction then we need a check in place to ensure that is the only way it can be run.

psakievich (changes requested):

amr-wind/wind_energy/actuator/wing/wing_ops.H
tonyinme: Thanks for noticing that @psakievich !
This was an old comment and the new formulation works for more directions.
I still need to put some thought to make sure that the wing can work in any direction without a problem. 
Will fix this tomorrow.

tonyinme:
@psakievich , it looks like this has been ready for a while and escaped our radar.
Let me know if this looks good and we can merge it.

psakievich:
@tonyinme let's get the CI passing and the branch updated. Then we can merge if you are happy with it.

psakievich (approved):

PR #883, opened 2023-07-20
Fix some warnings

marchdf:


jrood-nrel (approved):
Thanks for fixing these.

PR #885, opened 2023-07-25
Checkpoint tools

mbkuhn:
tools to
* coarsen checkpoint file by a factor of 2
* convert checkpoint file to CSV

### DIFF
--- a/tools/utilities/CheckpointToCSV/CheckpointFileUtil.H
+++ b/tools/utilities/CheckpointToCSV/CheckpointFileUtil.H
@@ -1,0 +1,114 @@
+#ifndef CHECKPOINTFILEUTIL_H_
+#define CHECKPOINTFILEUTIL_H_
+#include <AMReX_Config.H>
+
+#include <AMReX_Geometry.H>
+#include <AMReX_MultiFab.H>
+#include <AMReX_VisMF.H>
+#include <string>
+#include <memory>
+
+using namespace amrex;
+
+class CheckpointFileDataImpl
+{
+public:
+    CheckpointFileDataImpl(std::string const& chkptfile_name);
+
+    [[nodiscard]] int spaceDim() const noexcept { return m_spacedim; }
+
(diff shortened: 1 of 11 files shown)


tools/utilities/CheckpointToCSV/CheckpointFileUtil.H
```diff
@@ -0,0 +1,107 @@
...
+
+    [[nodiscard]] int finestLevel() const noexcept { return m_finest_level; }
+
+    [[nodiscard]] const BoxArray& boxArray(int level) const noexcept
+    {
+        return m_ba[level];
+    }
+
+    [[nodiscard]] const DistributionMapping&
+    DistributionMap(int level) const noexcept
```
marchdf, line 32: totally minor: you seem to be following camelCase for all the other member functions but this one.

tools/utilities/coarsen-chkpt/CoarsenCheckpt.H
```diff
@@ -0,0 +1,37 @@
...
+namespace tools {
+
+class CoarsenCheckpt : public incflo
+{
+public:
+    CoarsenCheckpt();
+
+    virtual ~CoarsenCheckpt() = default;
+
+    void read_chkpt_add_baselevel();
```
marchdf, line 16: and these are snake case ;)

mbkuhn: I bet it does look funny with the different cases! But this PR has a few tools in it. CoarsenCheckpt is separate from CheckpointToCSV. Each is different because of the code I started from.

tools/utilities/CheckpointToCSV/CheckpointFileUtil.H
mbkuhn: true! this is straight from "amrex/Src/Base/AMReX_PlotFileDataImpl.H" though

mbkuhn:
@marchdf are you good with these changes?

marchdf (approved):

PR #890, opened 2023-07-27
Bump clang-format

marchdf:


jrood-nrel (approved):
Should probably fix the name of the commit before changing.

PR #896, opened 2023-08-08
Picard both

mbkuhn:
just for comparison now


ashesh2512:
@mbkuhn Have you tried this on the convecting taylor vortex? Or the flow past sphere? For the former, I would be curious if you approach second order accuracy with an increase in the number of solution exchange iterations with an increase in the number of cells.

PR #897, opened 2023-08-09
Correcting Coriolis and geostrophic wind forcing to allow for any latitude

gdeskos:
- Changes in CoriolisForcing.cpp and GeostrophicForcing.cpp to allow for any latitude (other than 90 deg)
- Updating the calculation of the Coriolis factor to be f=2*Omega*sin(lat)
- Omega is now calculated using a rotational time period of 86164.091 seconds which corresponds to 23hrs, 56 minutes and 4 seconds, instead of 24hrs
- Added a Boolean to determine whether the full 3D or just 2D version of the Coriolis force will be used
- Unit-tests were updated and passing

### DIFF
--- a/amr-wind/equation_systems/icns/source_terms/CoriolisForcing.H
+++ b/amr-wind/equation_systems/icns/source_terms/CoriolisForcing.H
@@ -50,7 +50,7 @@ private:
     //! `2.0 * \Omega
     amrex::Real m_coriolis_factor{0.0};
 
-    bool m_is_horizontal{true};
+    bool m_is_horizontal{false};
 };
 
 } // namespace amr_wind::pde::icns
--- a/amr-wind/equation_systems/icns/source_terms/CoriolisForcing.H
+++ b/amr-wind/equation_systems/icns/source_terms/CoriolisForcing.H
@@ -44,8 +44,13 @@ private:
     amrex::Real m_sinphi{0.0};
     amrex::Real m_cosphi{0.0};
 
-    //! `2.0 * \Omega`
+    //! \Omega --> Earth’s rotation angular speed
+    amrex::Real m_omega{7.3e-5};
+
+    //! `2.0 * \Omega
     amrex::Real m_coriolis_factor{0.0};
+
+    bool m_is_2d{false};
 };
 
 } // namespace amr_wind::pde::icns
--- a/amr-wind/equation_systems/icns/source_terms/CoriolisForcing.H
+++ b/amr-wind/equation_systems/icns/source_terms/CoriolisForcing.H
@@ -50,7 +50,7 @@ private:
     //! `2.0 * \Omega
     amrex::Real m_coriolis_factor{0.0};
 
-    bool m_is_2d{false};
+    bool m_is_horizontal{true};
 };
 
 } // namespace amr_wind::pde::icns
--- a/amr-wind/equation_systems/icns/source_terms/CoriolisForcing.cpp
+++ b/amr-wind/equation_systems/icns/source_terms/CoriolisForcing.cpp
@@ -41,7 +41,7 @@ CoriolisForcing::CoriolisForcing(const CFDSim& sim)
     amrex::Real rot_time_period = 86164.091;
     pp.query("rotational_time_period", rot_time_period);
 
-    m_omega = 2.0 * utils::pi() / rot_time_period;
+    m_omega = utils::two_pi() / rot_time_period;
     m_coriolis_factor = 2.0 * m_omega;
 
     pp.queryarr("east_vector", m_east, 0, AMREX_SPACEDIM);
--- a/amr-wind/equation_systems/icns/source_terms/CoriolisForcing.cpp
+++ b/amr-wind/equation_systems/icns/source_terms/CoriolisForcing.cpp
@@ -41,7 +41,7 @@ CoriolisForcing::CoriolisForcing(const CFDSim& sim)
     amrex::Real rot_time_period = 86164.091;
     pp.query("rotational_time_period", rot_time_period);
 
-    m_omega = 2.0 * utils::two_pi() / rot_time_period;
+    m_omega = 2.0 * utils::pi() / rot_time_period;
     m_coriolis_factor = 2.0 * m_omega;
 
     pp.queryarr("east_vector", m_east, 0, AMREX_SPACEDIM);
--- a/amr-wind/equation_systems/icns/source_terms/CoriolisForcing.cpp
+++ b/amr-wind/equation_systems/icns/source_terms/CoriolisForcing.cpp
@@ -36,16 +36,21 @@ CoriolisForcing::CoriolisForcing(const CFDSim& sim)
     m_sinphi = std::sin(m_latitude);
     m_cosphi = std::cos(m_latitude);
 
-    // Read the rotational time period (in seconds)
-    amrex::Real rot_time_period = 86400.0;
+    // Read the rotational time period (in seconds) -- This is 23hrs and 56
+    // minutes and 4.091 seconds
+    amrex::Real rot_time_period = 86164.091;
     pp.query("rotational_time_period", rot_time_period);
-    m_coriolis_factor = 2.0 * utils::two_pi() / rot_time_period;
+
+    m_omega = 2.0 * utils::two_pi() / rot_time_period;
+    m_coriolis_factor = 2.0 * m_omega;
 
     pp.queryarr("east_vector", m_east, 0, AMREX_SPACEDIM);
     pp.queryarr("north_vector", m_north, 0, AMREX_SPACEDIM);
     utils::vec_normalize(m_east.data());
     utils::vec_normalize(m_north.data());
     utils::cross_prod(m_east.data(), m_north.data(), m_up.data());
+
+    pp.query("is_two_dimensional", m_is_2d);
 }
 
 CoriolisForcing::~CoriolisForcing() = default;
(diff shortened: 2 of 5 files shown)


mbkuhn (dismissed):

amr-wind/equation_systems/icns/source_terms/CoriolisForcing.H
```diff
@@ -44,8 +44,13 @@ private:
...
     amrex::Real m_cosphi{0.0};
 
-    //! `2.0 * \Omega`
+    //! \Omega --> Earth’s rotation angular speed
+    amrex::Real m_omega{7.3e-5};
+
+    //! `2.0 * \Omega
     amrex::Real m_coriolis_factor{0.0};
+
+    bool m_is_2d{false};
```
mchurchf, line 53: I would make it default to 2D like all other codes (other than Nalu).

amr-wind/equation_systems/icns/source_terms/CoriolisForcing.cpp
```diff
@@ -36,16 +36,21 @@ CoriolisForcing::CoriolisForcing(const CFDSim& sim)
     m_sinphi = std::sin(m_latitude);
     m_cosphi = std::cos(m_latitude);
 
-    // Read the rotational time period (in seconds)
-    amrex::Real rot_time_period = 86400.0;
+    // Read the rotational time period (in seconds) -- This is 23hrs and 56
+    // minutes and 4.091 seconds
```
mchurchf, line 40: I've been thinking about this a lot the past couple months.  24 hours is a solar day, and 23 hrs, 56 min, 4 s is a sidereal day, so with reference to fixed stars.  I think you are right in using it instead of the solar day.  What I am not completely sure about is that the Earth is both rotating and traveling about a curved orbit about the Sun.  Is the sidereal day the thing to use?  I think so...  Where can we find the definite answer.

amr-wind/equation_systems/icns/source_terms/CoriolisForcing.cpp
```diff
@@ -36,16 +36,21 @@ CoriolisForcing::CoriolisForcing(const CFDSim& sim)
...
+    amrex::Real rot_time_period = 86164.091;
     pp.query("rotational_time_period", rot_time_period);
-    m_coriolis_factor = 2.0 * utils::two_pi() / rot_time_period;
+
+    m_omega = 2.0 * utils::two_pi() / rot_time_period;
```
mchurchf, line 44: I don't agree here.  Omega should be 2 pi radians divided by the rotation period.  I don't understand the premultiplying by 2.

mbkuhn:
I think the mindset behind these changes is to be as accurate to reality as possible, so that by default the code is run to mimic real-world physics, but less realistic common practices are still available in the code for the sake of code-to-code comparisons. In several project, like AWAKEN for example, the code is being used in a predictive capacity, and that is how the code is primarily being perceived.

amr-wind/equation_systems/icns/source_terms/CoriolisForcing.cpp
mbkuhn: good catch, Matt. I agree that the calculation of omega should just be 2pi / rot_time_period. The new code can add clarity by showing m_omega directly but it shouldn't be numerically different.

unit_tests/wind_energy/test_abl_src.cpp
```diff
@@ -199,7 +199,8 @@ TEST_F(ABLMeshTest, geostrophic_forcing)
         geostrophic_forcing(lev, mfi, bx, amr_wind::FieldState::New, src_arr);
     });
 
-    constexpr amrex::Real corfac = 2.0 * amr_wind::utils::two_pi() / 86400.0;
+    constexpr amrex::Real corfac =
+        4.0 * amr_wind::utils::two_pi() / 86164.091 * 0.80901699437;
```
mbkuhn, line 203: the factor of 4, instead of 2, would need to be reverted back when the m_omega calculation is fixed

mchurchf:
> I think the mindset behind these changes is to be as accurate to reality as possible, so that by default the code is run to mimic real-world physics, but less realistic common practices are still available in the code for the sake of code-to-code comparisons. In several project, like AWAKEN for example, the code is being used in a predictive capacity, and that is how the code is primarily being perceived.

I am not sure that the full formulation is more accurate to reality.  It might be if our domain were curved like the surface of the Earth, but given it is flat, we might be making it even less accurate to reality by bringing in a term that doesn't make sense for an uncurved domain.  But if someone can show me on paper why this is a more accurate way to do it, then great.

amr-wind/equation_systems/icns/source_terms/CoriolisForcing.cpp
gdeskos: Haha, that's a good catch indeed. I hostly thought this was `utils::pi()` and I completely overlooked the two_ in front of it

unit_tests/wind_energy/test_abl_src.cpp
gdeskos: On it

amr-wind/equation_systems/icns/source_terms/CoriolisForcing.H
gdeskos: If we do that, we will need to update the reg tests

mchurchf (approved):

lawrenceccheung:
> > I think the mindset behind these changes is to be as accurate to reality as possible, so that by default the code is run to mimic real-world physics, but less realistic common practices are still available in the code for the sake of code-to-code comparisons. In several project, like AWAKEN for example, the code is being used in a predictive capacity, and that is how the code is primarily being perceived.
> 
> I am not sure that the full formulation is more accurate to reality. It might be if our domain were curved like the surface of the Earth, but given it is flat, we might be making it even less accurate to reality by bringing in a term that doesn't make sense for an uncurved domain. But if someone can show me on paper why this is a more accurate way to do it, then great. That z-term accounts for how horizontal velocity is transferred to vertical velocity and vice versa, but our flat domain doesn't allow this and I am concerned that the vertical term ends up just interacting with dp/dz in an unknown way.

One quick note about the sticking to reality versus modeling assumptions and the choices we make.  The difference between the solar day versus sidereal day is about (1-86164/86400)*100 = 0.27%.  However, note that we're also making the assumption that the Coriolis force is a _constant_ force throughout the domain.  Depending on where you are in the world, there could be a 0.27% difference in the Coriolis force in as short as ~20km due to a change in latitude between the N and S end of the domain.  So some errors might be swamped by other modeling assumptions.

Lawrence

amr-wind/equation_systems/icns/source_terms/CoriolisForcing.H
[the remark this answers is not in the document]
gdeskos: Changed to "is_horizontal"

amr-wind/equation_systems/icns/source_terms/GeostrophicForcing.cpp
```diff
@@ -22,28 +22,28 @@ namespace amr_wind::pde::icns {
...
+    // minutes and 4.091 seconds
+    amrex::Real rot_time_period = 86164.091;
+    ppc.query("rotational_time_period", rot_time_period);
 
-    {
-        // Read the geostrophic wind speed vector (in m/s)
-        amrex::ParmParse pp("GeostrophicForcing");
-        pp.getarr("geostrophic_wind", m_target_vel);
-    }
+    amrex::Real omega = 2.0 * utils::two_pi() / rot_time_period;
```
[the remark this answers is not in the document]
gdeskos: this is outdated

amr-wind/equation_systems/icns/source_terms/GeostrophicForcing.cpp
```diff
@@ -22,28 +22,28 @@ namespace amr_wind::pde::icns {
...
+
+    coriolis_factor = 2.0 * omega * sinphi;
+    amrex::Print() << "Geostrophic forcing: Coriolis factor = "
+                   << coriolis_factor << std::endl;
+
+    // Read the geostrophic wind speed vector (in m/s)
+    amrex::ParmParse ppg("GeostrophicForcing");
+    ppg.getarr("geostrophic_wind", m_target_vel);
 
     m_g_forcing = {
```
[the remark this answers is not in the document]
gdeskos: added fat to geostrophic forcing

unit_tests/wind_energy/test_abl_src.cpp
```diff
@@ -347,7 +348,7 @@ TEST_F(ABLMeshTest, hurricane_forcing)
 TEST_F(ABLMeshTest, coriolis_const_vel)
 {
     constexpr amrex::Real tol = 1.0e-12;
-    constexpr amrex::Real corfac = 2.0 * amr_wind::utils::two_pi() / 86400.0;
+    constexpr amrex::Real corfac = 4.0 * amr_wind::utils::two_pi() / 86164.091;
```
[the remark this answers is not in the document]
gdeskos: this is outdated

unit_tests/wind_energy/test_abl_src.cpp
```diff
@@ -429,7 +430,7 @@ TEST_F(ABLMeshTest, coriolis_height_variation)
     // ABL unit test mesh has 64 cells in z
     constexpr int kdim = 63;
     constexpr amrex::Real tol = 1.0e-12;
-    constexpr amrex::Real corfac = 2.0 * amr_wind::utils::two_pi() / 86400.0;
+    constexpr amrex::Real corfac = 4.0 * amr_wind::utils::two_pi() / 86164.091;
```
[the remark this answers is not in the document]
gdeskos: outdated

mbkuhn (approved):

PR #901, opened 2023-08-29
overset: calc pressure gradient before proceeding

mbkuhn:
- update AMReX-Hydro for new function
- use new function to get nalu pressure into gp

### DIFF
--- a/amr-wind/projection/incflo_apply_nodal_projection.cpp
+++ b/amr-wind/projection/incflo_apply_nodal_projection.cpp
@@ -435,3 +435,119 @@ void incflo::ApplyProjection(
         }
     }
 }
+
+void incflo::UpdateGradP(
+    Vector<MultiFab const*> density, Real time, Real scaling_factor)
+{
+    BL_PROFILE("amr-wind::incflo::UpdateGradP");
+
+    bool variable_density =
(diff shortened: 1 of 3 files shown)


amr-wind/projection/incflo_apply_nodal_projection.cpp
```diff
@@ -435,3 +435,119 @@ void incflo::ApplyProjection(
         }
     }
 }
+
+void incflo::UpdateGradP(
```
ashesh2512, line 439: Do you think there is any room to refactor this to extract the common code needed by the nodal projection step here also needed by incflo_apply_nodal_projection? I haven't found much redundancy.

ashesh2512 (approved):

psakievich (approved):

amr-wind/projection/incflo_apply_nodal_projection.cpp
mbkuhn: I attempted to do this, but it wasn't working. Issues with properly setting up sigma and boundary velocities somehow (compiled fine, some of the reg tests failed, some numerically and others by seg faults). The changes are in my branch "calcGradP_addlcode". Going to merge this as-is.

PR #909, opened 2023-09-25
Add dependabot

marchdf:
Automatically create PRs when certain submodules change. We can then decide whether to merge. More automation possible using instructions here: https://docs.github.com/en/code-security/dependabot/working-with-dependabot/automating-dependabot-with-github-actions

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+
+updates:
+  - package-ecosystem: gitsubmodule
+    schedule:
+        interval: "weekly"
+    directory: /
+    allow:
+      - dependency-name: "submods/amrex"
+      - dependency-name: "submods/AMReX-Hydro"


jrood-nrel (approved):

sayerhs:
@marchdf Is the intent to create a PR for ever commit that AMReX has? You might want to add a check on the release tags for AMReX and hydro submodules, otherwise this is going to create a lot of unnecessary noise.

jrood-nrel:
https://github.com/Exawind/amr-wind/blob/fae858477bb8023a62369e8753ac774d992f6183/.github/dependabot.yml#L6

marchdf:
Yeah this will just check weekly and then update to latest. So the noise is a weekly PR. If you dont do anything with that PR, the next PR will close the old one and open a new one.

sayerhs:
Ah I see. I didn't catch the fact that the previous PR will automatically be closed upon a new PR being created. Thanks for clarifying that.

PR #931, opened 2023-11-10
Delay output of plot or checkpoint files

mbkuhn:


marchdf (approved):

marchdf:
The excessive padding warning can sometimes come from the ordering of the struc/class's member variables. Something like this can trigger it:
```
int foo
string bar
int foo1
string bar1
int foo2
string bar2
```
whereas doing all ints then all strings will not.

PR #935, opened 2023-11-29
Implement a "coded" boundary condition

ewquon:
To facilitate the simulation of arbitrary inlet conditions. Code is based on the implementation of ABLModulatedPowerLaw.

To use:
* compile user-defined velocity and temperature field functions (e.g., [Rankine.cpp.txt](https://github.com/Exawind/amr-wind/files/13504889/Rankine.cpp.txt) in a standalone shared library (in this case, "Rankine.dylib" with `clang++ -v Rankine.cpp -o Rankine.dylib -shared -fPIC`)
* add `codedInlet.lib = "Rankine.dylib"` to your input file -- this will be loaded at runtime
* all `mass_inflow` boundaries will be set by the standalone code

### DIFF
--- a/amr-wind/wind_energy/ABL.H
+++ b/amr-wind/wind_energy/ABL.H
@@ -12,7 +12,7 @@
 #include "amr-wind/utilities/FieldPlaneAveraging.H"
 #include "amr-wind/wind_energy/ABLStatsBase.H"
 #include "amr-wind/wind_energy/ABLModulatedPowerLaw.H"
-#include "amr-wind/wind_energy/ABLCodedInlet.H"
+#include "amr-wind/wind_energy/ABLUserDefined.H"
 
 /**
  *  \defgroup wind Wind energy modeling
@@ -91,7 +91,7 @@ public:
 
     const ABLBoundaryPlane& bndry_plane() const { return *m_bndry_plane; }
     const ABLModulatedPowerLaw& abl_mpl() const { return *m_abl_mpl; }
-    const ABLCodedInlet& abl_coded_inlet() const { return *m_abl_coded_inlet; }
+    const ABLUserDefined& abl_udf() const { return *m_abl_udf; }
 
     //! Return the ABL statistics calculator
     const ABLStatsBase& abl_statistics() const { return *m_stats; }
@@ -117,8 +117,8 @@ private:
     //! ABL modulated power law instance
     std::unique_ptr<ABLModulatedPowerLaw> m_abl_mpl;
 
-    //! ABL coded inlet instance
-    std::unique_ptr<ABLCodedInlet> m_abl_coded_inlet;
+    //! ABL user-defined function instance
+    std::unique_ptr<ABLUserDefined> m_abl_udf;
 
     //! Reference to ABL forcing term if present
     mutable pde::icns::ABLForcing* m_abl_forcing{nullptr};
--- a/amr-wind/wind_energy/ABL.H
+++ b/amr-wind/wind_energy/ABL.H
@@ -12,6 +12,7 @@
 #include "amr-wind/utilities/FieldPlaneAveraging.H"
 #include "amr-wind/wind_energy/ABLStatsBase.H"
 #include "amr-wind/wind_energy/ABLModulatedPowerLaw.H"
+#include "amr-wind/wind_energy/ABLCodedInlet.H"
 
 /**
  *  \defgroup wind Wind energy modeling
@@ -90,6 +91,7 @@ public:
 
     const ABLBoundaryPlane& bndry_plane() const { return *m_bndry_plane; }
     const ABLModulatedPowerLaw& abl_mpl() const { return *m_abl_mpl; }
+    const ABLCodedInlet& abl_coded_inlet() const { return *m_abl_coded_inlet; }
 
     //! Return the ABL statistics calculator
     const ABLStatsBase& abl_statistics() const { return *m_stats; }
@@ -114,6 +116,9 @@ private:
 
     //! ABL modulated power law instance
     std::unique_ptr<ABLModulatedPowerLaw> m_abl_mpl;
+
+    //! ABL coded inlet instance
+    std::unique_ptr<ABLCodedInlet> m_abl_coded_inlet;
 
     //! Reference to ABL forcing term if present
     mutable pde::icns::ABLForcing* m_abl_forcing{nullptr};
--- a/amr-wind/wind_energy/ABL.H
+++ b/amr-wind/wind_energy/ABL.H
@@ -35,11 +35,18 @@ class incflo;
 
 namespace amr_wind {
 
+class ABLMesoscaleInput;
+
 namespace pde::icns {
 class ABLForcing;
 class ABLMeanBoussinesq;
+class ABLMesoForcingMom;
 class HurricaneForcing;
 } // namespace pde::icns
+
+namespace pde::temperature {
+class ABLMesoForcingTemp;
+}
 
 /** Atmospheric Boundary Layer physics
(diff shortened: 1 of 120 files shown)


ewquon:
Example input/output for above case
[abl_coded_inflow.inp.txt](https://github.com/Exawind/amr-wind/files/13504954/abl_coded_inflow.inp.txt)

https://github.com/Exawind/amr-wind/assets/18267059/94c03b5a-c93c-44fb-b516-322c9468d9a6

ashesh2512:
@jrood-nrel Any thoughts on why only HIP and SYCL are causing grief here? For the latter, it appears that an AMReX function is deprecated?

amr-wind/wind_energy/ABLCodedInlet.cpp
```diff
@@ -0,0 +1,200 @@
...
+            const auto& arr = mfab[mfi].array();
+
+            amrex::ParallelFor(
+                bx, [=, coded = m_user_coded] AMREX_GPU_DEVICE(
+                        int i, int j, int k) noexcept {
+                    const amrex::Real x = problo[0] + (i + 0.5) * dx[0];
+                    const amrex::Real y = problo[1] + (j + 0.5) * dx[1];
+                    const amrex::Real z = problo[2] + (k + 0.5) * dx[2];
+
+                    coded.get_temp(time, x, y, z, arr(i, j, k));
```
ashesh2512, line 194: So apparently SYCL kernel cannot call through a function pointer. @marchdf Do you know of workarounds here?

ashesh2512: I will add though, I don't understand the design here. Using a function pointer makes little sense to me, considering that it cannot represent both the host and device version.

ewquon: The function pointers were new for me. But it was based on examples I found for implementing runtime library linking. Perhaps there's another way?

ewquon: If there isn't a cross-platform solution, I could just create a cmake flag for this that's disabled by default...

amr-wind/wind_energy/ABL.H
```diff
@@ -90,6 +91,7 @@ public:
 
     const ABLBoundaryPlane& bndry_plane() const { return *m_bndry_plane; }
     const ABLModulatedPowerLaw& abl_mpl() const { return *m_abl_mpl; }
+    const ABLCodedInlet& abl_coded_inlet() const { return *m_abl_coded_inlet; }
```
marchdf, line 94: instead of "CodedInlet" I would favor renaming to UDF or UserDefined

amr-wind/wind_energy/ABLCodedInlet.cpp
```diff
@@ -0,0 +1,200 @@
...
+            userfun_lib = dlopen(m_user_lib.c_str(), RTLD_NOW);
+            m_user_coded.get_vel =
+                reinterpret_cast<Vfun_ptr>(dlsym(userfun_lib, "velocity"));
+            m_user_coded.get_temp =
+                reinterpret_cast<Tfun_ptr>(dlsym(userfun_lib, "temperature"));
+
+            // attempt to use user-coded functions
+            double Vtmp[3];
+            m_user_coded.get_vel(0, 0, 0, 0, Vtmp);
+            amrex::Print() << "ABLCodedInlet: Loaded xvelocity_field function "
```
marchdf, line 41: this looks like a debug statement that should be removed?

amr-wind/wind_energy/ABLCodedInlet.cpp
marchdf: hmmm cross-platform is going to be a problem I think... This is a hard thing to accomplish. In general, if you are going to compile a file to be included as a dl, why not just edit the code? We could make the use/definition of that kind of function easier for the user?

ewquon: @marchdf I see your point and that was going to be my original approach. But I was curious to see if this approach would work because it (arguably) has more usability and flexibility:
* The user can start with a pre-compiled amr-wind executable (that is, e.g., set up to be performant for the target system) and develop custom field functions w/o having to compile amr-wind on their own. (Side note: From tiny profiler, the `set_velocity` calls from this solution path amounts to 0.03%)
* The user will not have to understand the nuances of AMReX or GPU programming.
* This allows general inflows to be specified and is comparable to the OpenFOAM's codedFixedValue BC.

amr-wind/wind_energy/ABL.H
ewquon: "UDF" makes sense, but there are already UDFs in other parts of the code. I'll rename to UserDefined.

amr-wind/wind_energy/ABLCodedInlet.cpp
marchdf: Yeah I see the utility of this. Esp since OpenFoam does something like it. But (if I recall?) I think you supply text to openfoam and it converts it to code? There's also the option of driving this through py-amrex and then we can provide an easy python UDF...

ewquon: OpenFOAM has a JIT compiler setup, so this solution comes up a little bit short in comparison. I'm not sure how to implement py-amrex -- do you?

marchdf: Not off the top of my head. But you would need a py-amrex to drive the sim and then hooks in amr-wind to have the right callbacks. It would be work...

jrood-nrel:
I will drop my opinion in here. If this can't accommodate GPUs, I don't think it should be merged. I can't argue whether or not this is useful, but we put a lot of effort into making AMR-Wind portable on GPUs. I think there is more benefit to being portable than creating features only for CPUs. I also don't think implementing a `dlopen` in this code is something we will want to support.

ewquon:
@jrood-nrel I think the issue is only with SYCL... with HIP, it appears to just be an issue with a compile flag, and it already works with CUDA (see https://github.com/Exawind/amr-wind/actions/runs/7039362632/job/19158186631?pr=935). I'd say that we wouldn't necessarily be on the hook for maintaining a dev feature that's not enabled by default. And, for what it's worth, there is already 1 user lined up (@mchurchf).

mchurchf:
I do not understand why GPU-portability must ALWAYS trump usability.  In the end we need to apply this code to a huge variety of problems.  For example, I am having to apply this code to hurricane flows.  That requires flexility.  I think we can find a way to make it work and bring a flexible feature to this code that many users would benefit from.

jrood-nrel:
I'm not adamant about stopping this because I think peopling using the applications we develop is paramount. I just think other approaches should be explored before accepting `dlopen` into the code. If `dlopen` ends up to be the right option, the user library should probably be made a target in CMake that uses a consistent compilation, and maybe that is all we need in the end. However, I think getting user code implemented into the framework will take more effort than this and I think we are just arguing for usability from different perspectives. We need more people to weigh in on it.

michaelasprague:
Lots of discussion here, but I will echo @jrood-nrel , if "this can't accommodate GPUs, I don't think it should be merged."  A fundamental goal of the ExaWind codes is that they are performance portable, with the "portable" being paramount.  We need to be able to run these codes on Frontier, Aurora, Kestrel, Perlmutter, etc.

marchdf:
Talked with Jon and we have a bunch of ideas of how to make this work but I think a meeting would be nice to make sure we understand the goal and can best help out.

marchdf:
We talked and I think I can tweak this to make this do what everyone wants. I will be taking this on in the next week or so.

PR #937, opened 2023-12-04
ALM update

mbkuhn:
This implementation of ALM suits the Godunov-based framework of AMR-Wind, unlike the original ALM implementation, and it is more accurate than the previous "fix" from PR #868. This new update has been tested in realistic application cases, such as the NREL 5MW and IEA 15MW turbines, and it has been shown to address the problems linked to both previous ALM implementations in AMR-Wind. Simple 2D test cases have also confirmed the new approach as conforming to theory in the steady-motion case, with convergent behavior with decreasing dx and consistent load prediction with varying dt. This PR also features a new "moving wing" regression test to include actuator motion in the regression suite. Finally, this PR restores the original functionality of the actuator disk (ADM) from before PR #868.


amr-wind/wind_energy/actuator/wing/fixed_wing_ops.H
```diff
@@ -45,6 +45,26 @@ struct ReadInputsOp<FixedWing, ActSrcLine>
         pp.queryarr("chord", wdata.chord_inp);
         bool use_fllc = false;
         pp.query("fllc", use_fllc);
+
+        std::string motion_type{"none"};
+        pp.query("motion_type", motion_type);
+        {
+            if (amrex::toLower(motion_type) == "none") {
+                wdata.motion_type = 0;
```
tonyinme, line 53: Is there are reason to set these as integers as opposed to strings that pass the specified value directly?

amr-wind/wind_energy/actuator/wing/fixed_wing_ops.H
```diff
@@ -68,16 +88,64 @@ struct ReadInputsOp<FixedWing, ActSrcLine>
...
-        info.bound_box = amrex::RealBox(
-            amrex::min(p1.x(), p2.x()) - search_radius,
-            amrex::min(p1.y(), p2.y()) - search_radius,
-            amrex::min(p1.z(), p2.z()) - search_radius,
-            amrex::max(p1.x(), p2.x()) + search_radius,
-            amrex::max(p1.y(), p2.y()) + search_radius,
-            amrex::max(p1.z(), p2.z()) + search_radius
-        );
-        // clang-format on
+        // Set up bounding box depending on motion type
```
tonyinme, line 91: I guess the reason for the integers is the switch statement here.
Was there a reason to use the switch statements instead of if statements?
If you leave the switch statements (they are fine) can you please add a comment to each case to explain what they are, so that it makes it easier to read the code (non, linear, sine, etc).

amr-wind/wind_energy/actuator/wing/wing_ops.cpp
```diff
@@ -168,4 +168,63 @@ void write_netcdf(
...
+    int motion,
+    amrex::Real period,
+    vs::Vector svec)
+{
+    // Get displacement of points from n to n+1
+    // Also, if translation velocity changes, update it
+    vs::Vector disp{0.0, 0.0, 0.0};
+    switch (motion) {
+    case (0):
+        // "none"
```
tonyinme, line 196: nice, I see the comments here!

docs/sphinx/user/inputs_Actuator.rst
```diff
@@ -179,6 +164,36 @@ Example for ``FixedWingLine``::
    
    This is how often to write actuator output. The default is ``10``.
 
+.. input_param:: Actuator.FixedWingLine.motion_type
+
+   **type:** String, optional
+
+   Slightly contrary to its name, the FixedWingLine actuator allows for motion,
```
tonyinme, line 171: Fixed wing is the correct terminology even if the wing is moving.
I think it is best to remove "Slightly contrary to its name"

tonyinme (commented):
Good job @mbkuhn ! 
Thanks for implementing these new capabilities and the tests.
I have some minor feedback.

amr-wind/wind_energy/actuator/ActSrcLineOp.H
```diff
@@ -92,7 +111,10 @@ void ActSrcOp<ActTrait, ActSrcLine>::operator()(
 
         amrex::Real src_force[AMREX_SPACEDIM]{0.0, 0.0, 0.0};
         for (int ip = 0; ip < npts; ++ip) {
-            const auto dist = cc - pos[ip];
+            // Put force at n+1/2 location for Godunov
+            constexpr amrex::Real wt = 0.5;
```
mbkuhn, line 115: wt = 1.0 is equivalent to the original implementation of ALM in AMR-Wind

amr-wind/wind_energy/actuator/wing/wing_ops.cpp
```diff
@@ -168,4 +168,56 @@ void write_netcdf(
...
+
+void new_wing_position_velocity(
+    VecList& points,
+    vs::Vector& vtr,
+    const int npts,
+    amrex::Real tn,
+    amrex::Real tnp1,
+    const std::string motion,
+    amrex::Real period,
+    vs::Vector svec)
```
marchdf, line 189: feels like a bunch of these args can be `const` so the next developer knows what is in/out.

amr-wind/wind_energy/actuator/wing/fixed_wing_ops.H
```diff
@@ -45,6 +45,24 @@ struct ReadInputsOp<FixedWing, ActSrcLine>
         pp.queryarr("chord", wdata.chord_inp);
         bool use_fllc = false;
         pp.query("fllc", use_fllc);
+
+        pp.query("motion_type", wdata.motion_type);
+        {
+            if (amrex::toLower(wdata.motion_type) == "none") {
```
marchdf, line 51: if this isn't doing anything, then it can be removed?

marchdf (approved):

PR #940, opened 2023-12-12
Add a CustomVelocity/Scalar UDF and a Rankine UDF

marchdf:
In reference to the discussion in #935 

This PR:
- creates an empty `CustomVelocity` UDF for a user to populate velocity easily without needing to know all the other hooks necessary. The user just needs to know about 2 files.
- creates an empty `CustomScalar` UDF for a user to populate any scalar easily without needing to know all the other hooks necessary. The user just needs to know about 2 files.
- creates a Rankine Velocity UDF (bonus!)

Still to do:
- [x] document the use of `CustomVelocity` and `CustomScalar` in rst format
- [x] Add a similar functionality for scalars

Closes #935


marchdf:
@ewquon I think I finished what I wanted to do here. Can you let me know if this helps your use case and if anything else is missing?

ewquon:
This accomplishes what I set out to do. My only comment is that the user will have to make edits in four places instead of one, but I do like that it allows for inputs from other parameter lists to be pulled in (e.g., reusing `incflo.velocity`). Let's add @mchurchf as a reviewer as he'll be a user and see if he has any useful feedback.

ewquon:
I talked to @mchurchf this morning and he has reviewed these changes and thinks they are fine and meet his needs.

mbkuhn (approved):

Issue #857, opened 2023-06-13
"Invalid particle" error when adding refinements to simulation


jfrederik-nrel:
Hi there, I am running AMR-Wind with OpenFAST, and getting an error that seems related to the refinements I do around the OpenFAST turbine. The error I'm getting is:

`Step: 14403 dt: 0.05 Time: 7200.1 to 7200.15
CFL: 0 (conv: 0 diff: 0 src: 0 )
terminate called after throwing an instance of 'std::runtime_error'
  what():  ParticleContainer::locateParticle(): invalid particle.
MPT ERROR: Rank 12(g:12) received signal SIGABRT/SIGIOT(6).
	Process ID: 196504, Host: r2i2n29, Program: /home/jfrederi/projects/amr-wind/build/amr_wind
	MPT Version: HPE MPT 2.23  11/19/20 04:21:56`

This error seems related (identical?) to the issue reported here: https://github.com/Exawind/amr-wind/issues/415 However, this is an old issue and it doesn't offer a solution, as it's supposed to be resolved in newer versions of AMR-Wind. 

I'm using two refinement boxes around my turbine. When I remove the second level, the simulation runs. I've also tried using cylinders instead of boxes, but this results in the same error. My simulation input file is attached below. My AMR-Wind header is shown below:

```
                              AMR-Wind (https://github.com/exawind/amr-wind)

  AMR-Wind version :: e0ffbb6-DIRTY
  AMR-Wind Git SHA :: e0ffbb62eb60d70d74b02ed2a2587551face2aa2-DIRTY
  AMReX version    :: 23.01-26-ga232be4e275b

  Exec. time       :: Mon Jun 12 17:21:09 2023
  Build time       :: Feb 15 2023 16:21:48
  C++ compiler     :: GNU 8.4.0

  MPI              :: ON    (Num. ranks = 64)
  GPU              :: OFF
  OpenMP           :: OFF

  Enabled third-party libraries: 
    NetCDF    4.7.3
    OpenFAST
```

If anybody can help me resolve this issue, it would be much appreciated!

[simulation.zip](https://github.com/Exawind/amr-wind/files/11739623/simulation.zip)

jfrederik-nrel:
[...]

Issue #515, opened 2021-12-15
KSGS issues on restart with refinement


neilmatula:
The OneEqKsgs84 turbulence model seems to sometimes have issues when restarting and adding refinement at the same time.
The residual reported for all equations becomes zero after the first timestep.
A small case that demonstrates the issue is defined by the two attached input files.
To reproduce the issue, run precursor.i, followed by turbine.i.
For reference, this was done on eagle using 64 processors.
The issue does not appear if there is no restart, or if there is only one level of refinement (max_level=1), or if Smagorinsky is used.
[issue.zip](https://github.com/Exawind/amr-wind/files/7721517/issue.zip)

gantech:
`#¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨#
#        ADAPTIVE MESH REFINEMENT       #
#.......................................#
amr.n_cell              = 128 128 32    # Grid cells at coarsest AMRlevel
amr.max_level           = 2           # Max AMR level in hierarchy 
#¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨¨#
#              GEOMETRY                 #
#.......................................#
geometry.prob_lo        =   0.       0.     0.  # Lo corner coordinates
geometry.prob_hi        =   5000.  5000.  1000.  # Hi corner coordinates
geometry.is_periodic    =   1   1   0   # Periodicity x y z (0/1)
incflo.delp             =   0.  0.  0.  # Prescribed (cyclic) pressure gradient`

Does this uniformly refine the mesh across the entire domain?

neilmatula:
No, it should refine according to the refinement blocks blocks at the very bottom, in the mesh refinement section.

michaeljbrazell:
this might help: https://github.com/Exawind/amr-wind/pull/492

tonyinme:
This [function call](https://github.com/Exawind/amr-wind/blob/c1531ecc7ccccef0956cb72e3461535375dbe4f6/amr-wind/equation_systems/tke/source_terms/KsgsM84Src.cpp#L59) throws an error in the debug build.

I am not entirely sure what this does, but this fix supresses the error: ` amrex::Box blo = amrex::adjCellLo(bx, dir, 1) & bx;`
[...]

PR #942, opened 2023-12-15
Addressing stability issues with introducing refinement levels on a restart

mbkuhn:
- some turbulence models feature quantities (with their own PDEs) that must remain positive, such as tke and sdr
- if the values go negative, NaNs will result, due to the square root of a negative number, etc.
- these models have "post-solve" operations that clip the values to guarantee their boundedness during a simulation
- but upon restart, when new levels are introduced, these values are not guaranteed to remain bounded when being interpolated to finer levels and this can be exacerbated at the wall, especially for walls BCs with extrapolation for scalars, such as slip_wall
- this PR puts the post-solve operations for all PDEs into their initialization (active on a restart) and into the post_regrid actions, which could have the same behavior as introducing finer levels to a restart
- this PR should eliminate issues with having refinement zones extend all the way to the lower boundary in turbine simulations that follow a coarser precursor simulation, such as #515 and #857.

### DIFF
--- a/amr-wind/equation_systems/PDE.H
+++ b/amr-wind/equation_systems/PDE.H
@@ -71,6 +71,9 @@ public:
             m_fields, m_sim.has_overset(), variable_density,
             m_sim.has_mesh_mapping()));
         m_src_op.init_source_terms(m_sim);
+
+        // clip (if needed) and fillpatch
+        m_post_solve_op(m_time.current_time());
     }
 
     //! Perform update actions after a regrid is performed
@@ -90,6 +93,9 @@ public:
         m_adv_op.reset(new AdvectionOp<PDE, Scheme>(
             m_fields, m_sim.has_overset(), variable_density,
             m_sim.has_mesh_mapping()));
+
+        // clip (if needed) and fillpatch
+        m_post_solve_op(m_time.current_time());
     }
 
     //! Return the object holding the fields necessary for solving this PDE
--- a/amr-wind/equation_systems/PDE.H
+++ b/amr-wind/equation_systems/PDE.H
@@ -72,7 +72,7 @@ public:
             m_sim.has_mesh_mapping()));
         m_src_op.init_source_terms(m_sim);
 
-        // clip (if needed) and fillpatch
+        // Post-solve operations should also apply after initialization
         m_post_solve_op(m_time.current_time());
     }
 
@@ -94,7 +94,7 @@ public:
             m_fields, m_sim.has_overset(), variable_density,
             m_sim.has_mesh_mapping()));
 
-        // clip (if needed) and fillpatch
+        // Post-solve operations should also apply after a regrid
         m_post_solve_op(m_time.current_time());
     }
 
--- a/amr-wind/incflo.cpp
+++ b/amr-wind/incflo.cpp
@@ -107,7 +107,7 @@ void incflo::init_amr_wind_modules()
         eqn->initialize();
     }
 
-    // m_sim.pde_manager().fillpatch_state_fields(m_time.current_time());
+    m_sim.pde_manager().fillpatch_state_fields(m_time.current_time());
     m_sim.post_manager().post_init_actions();
 }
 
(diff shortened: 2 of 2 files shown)


amr-wind/equation_systems/PDE.H
```diff
@@ -71,6 +71,9 @@ public:
             m_fields, m_sim.has_overset(), variable_density,
             m_sim.has_mesh_mapping()));
         m_src_op.init_source_terms(m_sim);
+
+        // clip (if needed) and fillpatch
```
marchdf, line 75: I would remove this comment, since a lot of different stuff can happen in the post_solve_op

amr-wind/incflo.cpp
```diff
@@ -107,7 +107,8 @@ void incflo::init_amr_wind_modules()
         eqn->initialize();
     }
 
-    m_sim.pde_manager().fillpatch_state_fields(m_time.current_time());
+    m_sim.pde_manager().fillpatch_state_fields(
+        m_time.current_time()); // may be unnecessary
```
mbkuhn, line 111: I added comments here, because I think that the changes in PDE.H make this call unnecessary. The default PostSolveOp as specified in PDEOps.H is a fillpatch of the pde field. This would imply that any custom PostSolveOp should at least contain the fillpatch step (and some also have clipping and calculation of other variables). Now that the PostSolveOp is included in the initialize() and post_regrid() steps for every PDE, there shouldn't be a need to loop through each PDE and fillpatch its variable, which is what fillpatch_state_fields does. The additional nuance of setting the state of the variable is not used in the code, it always uses the default of the New field state, aka the current value.

mbkuhn:
From discussion with @rybchuk, the previous recommendation to address the issue of unbounded tke was to manually set the interpolation type using TKE.interpolation = PiecewiseConstant. However, as a standard practice, I think that clipping is a better option than using a less accurate interpolation because the values are clipped every timestep anyway. Furthermore, there are other quantities that need to stay bounded, that this interpolation modification isn't set up for.

mbkuhn:
The reason why the code gets NaNs from negative tke values is because there are multiple terms calculated using sqrt(tke). Although it is important to make the tke values bounded, should we also ensure that we can't get NaNs from unbounded tke? Should we replace the sqrt(tke) operations with sqrt(max(0,tke)) to ensure that the code can keep going?

marchdf:
> Should we replace the sqrt(tke) operations with sqrt(max(0,tke)) to ensure that the code can keep going?

I think not. Clearly this exposed some behavior that was unexpected (as in we didn't realize the clipping wasn't happening). Though of course we do that kind of thing all the time (another example is avoiding divisions by zero with an epsilon). I still think I would rather see the bad behavior, think about why it's happening and then come up with the right fix (which may be something like sqrt(max(0, val))). 

>  I think that clipping is a better option than using a less accurate interpolation

Agreed. Particularly since the clipping (when it is being done) is typically localized (e.g. to the wall) and the changing the interpolation scheme could have a wider effect.

amr-wind/incflo.cpp
marchdf: Good point. Though of course it is always scary to remove a fillpatch... the consequences could be dire. But I think that under these conditions, you can be confident that it happens. We could comment that out with some version of your text included to indicate that we don't need the fillpatch because it happens in post_solve. Just in case this does come back to bite us. And then some time down the road when we are confident it didn't bite us, we can remove the comment. Thoughts?

mbkuhn: sounds good! I think we may not be able to comment it out because of CI, and also these instances are the only ones where this function is used, which would mean we should remove the fillpatch_state_fields function. But I think we need to comment somewhere to mention the fillpatch being included through another means. I can check the reg tests and the test case that led to this bug fix. I'll also confirm from the code that every pde has a post_solve step with fillpatch.

marchdf (approved):

amr-wind/incflo.cpp
mbkuhn: turns out that the fillpatch_state_fields function is still necessary. Most of it is not, but the part where a density fillpatch is performed for constant density simulations needs to be there. In my opinion, that should be inside the icns PDE and not in the overall PDE system, because the density field is created when icns is created. Plus, much of this involves the m_constant_density boolean, which tends to be forgotten in input files for multiphase problems because it isn't required (no consequences for the simulation, but leads to unnecessary copying). fixing that in a nice way is a big enough scope for another PR.

I did confirm with fcompare that the constant density fillpatch is the only non-redundant part of fillpatch_state_fields.

marchdf: Ok lmk if you need help on that.

PR #944, opened 2023-12-20
New mass outflow boundary condition

gdeskos:
- New mass outflow boundary condition to be implemented
- The new BC will take into account the difference between mass inflow and mass outflow (through the local advection velocity)


amr-wind/boundary_conditions/MassOutflowBC.H
```diff
@@ -0,0 +1,35 @@
+#ifndef MASSINFLOWBC_H
+#define MASSINFLOWBC_H
```
mbkuhn, line 2: typo - you need MASSOUTFLOWBC_H

PR #474, opened 2021-08-25
passive tracer tagging

michaeljbrazell:
- adding as a draft for now, please give comments. I'm not thrilled with the naming of this, anyone have suggestions?
- I tested this on a uniform disk and it seems to work, I'll attach the input file

PR #949, opened 2024-01-04
Add a passive scalar equation

marchdf:
Pulling in just the passive scalar PDE from #474. This seems to be needed by ERF (@asalmgren  and @baperry2). I also switched the ScalarAdvection physics class to use the new passive scalar PDE instead of hijacking the temperature PDE.

### DIFF
--- a/amr-wind/equation_systems/CMakeLists.txt
+++ b/amr-wind/equation_systems/CMakeLists.txt
@@ -10,4 +10,4 @@ add_subdirectory(tke)
 add_subdirectory(sdr)
 add_subdirectory(levelset)
 add_subdirectory(vof)
-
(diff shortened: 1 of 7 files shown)


mbkuhn:
the code looks good! Could you add a regression test that uses this equation? can be very basic for now

marchdf:
Because I changed the ScalarAdvection case to use this new PDE both `scalar_advection_*` reg tests should be exercising this capability. However, I should add some documentation.

mbkuhn (approved):

rybchuk:
I know this branch is closed, but I just want to say that I'm excited to see this capability in AMR-Wind!

PR #953, opened 2024-01-09
ABL + waves: forcing term modifications

mbkuhn:


mbkuhn:
Starting with the source terms. No regression test yet because there are more PRs to come, but I want to keep the PR size more manageable. Have confirmed that the changes do not affect the existing reg test results.

marchdf (approved):

marchdf:
Nice work! Did you want to add documentation?

mbkuhn:
Documentation is a good idea, but in case this approach does not actually work in practice, I'm going to wait until more of the pieces are put together.

PR #954, opened 2024-01-09
Radar and Spinner Lidar Sampling 

ndevelder:
Large commit for new sampling that mimics Spinner Lidar and Radar


amr-wind/utilities/sampling/ConeSampler.cpp
```diff
@@ -0,0 +1,189 @@
+#include "amr-wind/utilities/sampling/ConeSampler.H"
+#include "amr-wind/CFDSim.H"
+#include "amr-wind/utilities/tensor_ops.H"
+
+#include "AMReX_ParmParse.H"
+
+#define PI 3.14159265
```
mbkuhn, line 7: shouldn't need to define PI here. pi is available from AMReX by using M_PI, and this is the standard in other portions of the code. you should be able to remove this line and replace PI with M_PI.

amr-wind/utilities/sampling/DTUSpinnerSampler.H
```diff
@@ -54,9 +74,51 @@ public:
...
 private:
+    // Number of subsamples in this timestep
     int m_ns{1};
-    amrex::Real m_dt_s{0.01};
+
+    // Max number of subsamples in all timesteps
+    int m_ntotal{1};
+
+    int m_update_count{0};
+    amrex::Real radtodeg{180.0 / M_PI};
```
mbkuhn, line 84: looks like some of these should be const

amr-wind/utilities/sampling/SamplingContainer.cpp
```diff
@@ -95,6 +95,8 @@ void SamplingContainer::initialize_particles(
 {
     BL_PROFILE("amr-wind::SamplingContainer::initialize");
 
+    amrex::Print() << "Start Initialize Particles" << std::endl;
```
mbkuhn, line 98: There are a lot of new print statements added in this PR. Are all of these desired for the final version? Some of them seem more targeted for development and debugging. If they are intended to go into main, I recommend that they be turned on via some verbosity level.

mbkuhn:
This PR looks like a lot of work and will be a valuable addition to the code. It really needs some unit tests prior to merging, addressing each new sampling type, especially due to the size of the PR. unit_tests/utilities/test_sampling.cpp and unit_tests/wind_energy/actuator/test_actuator_sampling.cpp should provide a starting point. Address the unit tests, the remaining CI, and the minor stuff in my line comments, and we should be able to merge this.

ndevelder:
Thanks for all the comments Michael! Haha, sorry for being the example of what not to do as far as testing...I do think there's a lot we could do there and I'll get on it asap! And yeah, definitely too large of a PR...it started as multiple projects but they ended up depending on each other and I didn't see a good way to split. Definitely open to suggestions on this front. I'll try and write up a short summary/guide to the PR so that other reviewers can get the overview without wading through so much code.

mbkuhn:
It's all good, keeping PRs small is not an easy achievement. The unit testing being in place will make that less of a worry, though.

ndevelder:
Ok, so I'm moving this out of draft as soon as all the checks go through. I think I've addressed everything (albeit in the most minimal way possible). There are a few details to note:

-  I removed ConeSampler since it is untested and not part of any work package...it was a waypoint on the path to getting the radar sampler in.
- The spinner and radar sampler unit tests only check the expected number of sampling locations, but could have a lot more work done to them. I'm hoping we can file this under future nice things to have. We've tested both samplers outside of the unit and reg test frameworks and have thought the results are reasonable. 
- After thinking about it some more, I don't think there is a place for actuator testing here...the sampler "attaches" to one openfast instance and reads the orientation matrix coming from openfast...but has no actual interaction with actuator points. And it seems like we're still working on a simple/seamless integration of openfast models in the testing infrastructure? If there are any ideas on something to test related to this I'm happy to consider, but I think we need openfast testing integration first? 

I'd also like to add a reg test for each sampler in a subsequent PR. I can open an issue and assign to myself if that seems worth it.

amr-wind/utilities/sampling/SamplingUtils.cpp
```diff
@@ -0,0 +1,187 @@
...
+vs::Tensor scale(vs::Tensor v, double a)
+{
+    vs::Tensor vnew;
+    for (int j = 0; j < 9; ++j) {
+        vnew[j] = a * v[j];
+    }
+    return vnew;
+}
+
+void spherical_cap_quadrature(
```
mbkuhn, line 83: There is a lot of code added in SamplingUtils, and these standalone functions seem like good candidates for unit testing. Don't need tests for every single one, but ideally if there is a basic problem with a known solution for a few of the complicated ones, it would help solidify this part of the code and rule out errors here if there is debugging required in the future.

mbkuhn:
I think this PR is really close. 2 things come to mind: unit tests for SamplingUtils, which I mentioned above, and the preexisting sampling methodology. Because this PR reworked some of the functions in Sampling.cpp, is there any concern that the changes have affected the ordinary sampling there? Unit tests exist for some of that already, I'm just wanting to confirm that this aspect has been looked at.

ndevelder:
Jumping back into this today! A couple points related to your comments:

- Happy to look at sampling utils and construct a test or two
- We've done testing and production runs with this PR for the RAAW project (multiple samplers in addition to the spinner) and Dan didn't see any problems. I do think messing with the buffer variable comes with some risk, but we haven't been able to break it yet.
- Other than that, it was mostly tidying, line-of-sight calculation for the radar, and allowing for data reduction in the sampler (with the default for most samplers that num points = num output points)

mbkuhn:
Excellent. @ndevelder I'll approve and merge after some unit tests for Sampling Utils functions are added.

ndevelder:
I got some of the simpler SamplingUtils tests in, but I don't have an easy way to get an independent result for the spherical cap quadrature stuff. That was effectively a direct copy from Nalu-Wind implemented by Robert K. I suppose I could try to put in something random and get a result, and just use that result to make sure it doesn't change in the future?

mbkuhn:
Thanks for doing this, Nate! That is fine with me, if it's not too much of a lift. Just to safeguard the current version.

mbkuhn:
Great work, Nate. Thanks for cooperating with my requests and improving the standards of the code.

mbkuhn (approved):

PR #955, opened 2024-01-11
Add documentation for mesoscale forcing capabilities

ewquon:
This PR documents the mesoscale-to-microscale coupling strategies that were introduced by https://github.com/Exawind/amr-wind/pull/652. These strategies provide mesoscale forcing through source terms based on:
1. Tendencies (or mesoscale budget components) 
2. Profile assimilation — generalized to include "direct" and "indirect" forcing schemes, applied to the full simulation domain or a portion thereof.

### DIFF
--- a/docs/sphinx/theory/theory.rst
+++ b/docs/sphinx/theory/theory.rst
@@ -132,7 +132,7 @@ Often for simulations involving walls, (e.g., channel flows, complex terrains et
 .. toctree::
    :glob:
    :maxdepth: 2
-   
+
    mapping.rst
 
 Multiphase flow modelling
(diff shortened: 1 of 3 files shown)


marchdf:
Thank you for doing this! One suggestion I might have is to refer to this documentation in the "theory manual" part of the documentation:https://exawind.github.io/amr-wind/theory/theory.html#source-terms We have a section there on GravityForcing in the "source terms" section. Just a short section/sentence with a cross reference of sort would probably be plenty.

ewquon:
Good suggestion! I'll add some text shortly.

marchdf (approved):

PR #956, opened 2024-01-11
Calculate planar average of tke only if it is available

ewquon:
Bug is triggered if using Smagorinsky and amrwind is built with netcdf.


mbkuhn:
nicely done, Eliot!

mbkuhn (approved):

PR #957, opened 2024-01-12
Further generalize partial profile assimilation approach, add reg tests

ewquon:


ewquon:
@gdeskos, @mbkuhn, and @mchurchf — this is the code version used to generate the examples that I will be providing you. Note that these changes are only needed for the most general "abl_meso.inp" case. The other 3 examples do not require this.

mbkuhn:
Before merging this PR, I'd like some form of confirmation that the reg tests can run after someone installs it on another machine. If it's true that these features are dependent on a specific version of NetCDF (which the discussion between @ewquon and @gdeskos suggested), then that needs to be made clear or modified.

mbkuhn:
Going to merge this now and add improvements later!

mbkuhn (approved):

PR #965, opened 2024-01-24
MultiPhase ABL (periodic)

mbkuhn:

PR #970, opened 2024-02-07
Enable multilevel inflow/outflow BC with native format

marchdf:
This was already done for netcdf, now doing it for the native format.


marchdf:
[Results of CPU test](https://github.com/Exawind/amr-wind/files/14215170/amr-wind-cpu-tests.o14054498.txt). The only test failing is `hbl_godunov` and it is failing because of something changed in #965. @mbkuhn is checking to see that that is ok.

mbkuhn (approved):

mbkuhn:
Were you able to confirm that it's "working"? The reg test is fine going forward, but because there's no unit test, have you checked that it's actually writing and reading the additional level?

marchdf:
Yup it's all working. The reg tests basically test that and it works. A long time ago we talked about unit tests but ultimately gave up on the idea. It's a weird capability because it interacts with the filesystem so much and require basically a fully functional sim that the "best" way to test was the reg tests with test dependencies.

[Results of GPU tests](https://github.com/Exawind/amr-wind/files/14223351/amr-wind-gpu-tests.o14054601.txt)

mbkuhn:
Yep, sounds great! Thanks for filling me in.

ewquon:
Fantastic! Thanks for putting this in, @marchdf. We were just talking a couple days ago about how this is critically needed for another project.

Issue #750, opened 2022-12-09
Discontinuities in sampled velocities passed to OpenFAST on first time-step


tonyinme:
When running a simulation using OpenFAST and ALM, there is a problem with the velocities being passed to openfast.
This is a one-blade turbine simulation (fixed wing) that we are using to debug.
These [velocities](https://github.com/Exawind/amr-wind/blob/a409c5af9cabc27b934dca7d63f4632b47a5701f/amr-wind/wind_energy/actuator/turbine/fast/turbine_fast_ops.H#L390) have a problem shown in the file below.
On the very first time-step, these velocities should all match the initial condition (1 everywhere).

This issue not seem to depend on number of cores and can be reproduced with only 1 core. 

![image](https://user-images.githubusercontent.com/11055361/206808538-7517ff5e-de25-4d20-8058-5d0cae32376b.png)

tonyinme:
Following up on our conversation with @psakievich @lawrenceccheung @michaelasprague , here are the input files for a case with less number of actuator points (60) and coarser epsilon (epsilon/R=.1). The same issue is still present even when using less actuator points.

[amr_wind_nrel5mw_fixed_wing_coarse.zip](https://github.com/Exawind/amr-wind/files/10997075/amr_wind_nrel5mw_fixed_wing_coarse.zip)

![image](https://user-images.githubusercontent.com/11055361/225790604-9bd3dfa0-792c-42f2-bcc7-f98d5394f7cf.png)

tonyinme:
Hi @marchdf , yes, this is still an issue. It only happens for the first time-step, so it is not stopping our simulations from running or affecting the results after some time.

mbkuhn:
> Hi @marchdf , yes, this is still an issue. It only happens for the first time-step, so it is not stopping our simulations from running or affecting the results after some time.

Do you know if this is an issue upon restart?

tonyinme:
That's a good question. I haven't checked restarting. I see it in the
beginning of a simulation.

tonyinme:
@mbkuhn , I tried running a restart simulation with the test above and I can confirm that this only shows up in the very first time-step of a new simulation. When restarting the simulation, we do not see the same problem.

PR #974, opened 2024-02-13
Fix for correct velocity sampling upon initialization (actuator methods)

mbkuhn:
fillpatch velocity field prior to sampling
addresses #750

### DIFF
--- a/amr-wind/wind_energy/actuator/Actuator.cpp
+++ b/amr-wind/wind_energy/actuator/Actuator.cpp
@@ -66,6 +66,10 @@ void Actuator::post_init_actions()
         act->init_actuator_source();
     }
 
+    // Ensure velocity fills ghost cells prior to sampling
+    auto& vel = m_sim.repo().get_field("velocity");
+    vel.fillpatch(m_sim.time().current_time());
+
     setup_container();
     update_positions();
     update_velocities();


mbkuhn:
@tonyinme feel free to confirm that this fixes the issue. I got it to work for me.

mbkuhn:
@marchdf I think this brings up a bigger issue, which is why I'm keeping this as a draft pull request. In many cases, the "initialize_fields" functions for different physics cases do not fill the ghost cells. This particular bug #750 reveals this issue because the particles were sampling some real values in the domain (1.0) and some non-initialized ghost cells (0.0) leading to sampled velocities between 0 and 1 (mostly 0.5). 

One option would be to make sure that initialize_fields functions go into the ghost cells. A fill patch operation doesn't really make sense there because those are level-based functions. Instead, I think it makes the most sense for the PDEs to call a fillpatch on their fields as a post_init step, but that is currently only taking place as a post_solve step. Does anything fill the ghost cells prior to the first advance?

tonyinme:
Thank you @mbkuhn ! I can confirm that the issue is no longer there with this new fix.

mbkuhn:
@marchdf we have much less of an issue to worry about, actually. I already fixed the lack of a fillpatch prior to advancing with this line: https://github.com/Exawind/amr-wind/blob/main/amr-wind/equation_systems/PDE.H#L76 . However, what we are running into now is the fact that the post_init actions of physics classes happen before the initialize() calls for PDEs. https://github.com/Exawind/amr-wind/blob/main/amr-wind/incflo.cpp#L105 . We could switch the ordering here and see what happens to the reg tests.

marchdf:
@tonyinme  can you check this PR again? @mbkuhn made a couple changes.

mbkuhn:
> @tonyinme can you check this PR again? @mbkuhn made a couple changes.

@tonyinme Tony, it's alright, I checked it myself.

marchdf (approved):

tonyinme:
Sounds good, I checked and the fix works well with the newer version!

PR #975, opened 2024-02-13
Initialize PDEs before physics post_init

mbkuhn:


mbkuhn:
reg tests and code analysis revealed this is a bad idea

PR #979, opened 2024-02-20
Donelan Heat Flux Calculation to Work in HEAT_FLUX Mode

mchurchf:
Adding a switch to the Donelan wall temperature flux model so that if the user specifies heat flux (HEAT_FLUX mode in MOData), Donelan just uses it rather than tries to recompute it with the 0.0012 * wspd_mean * (theta_surface - theta_mean) formula.  That formula should only be used if the user specifies a surface temperature (SURFACE_TEMPERATURE mode in MOData).


gdeskos (approved):

gdeskos:
@mbkuhn, since I have no merging rights anymore, would you be able to review and merge ? We need this asap!

mbkuhn (approved):

PR #980, opened 2024-02-21
Hurricane Temperature forcing

gdeskos:
Implementing a hurricane forcing term for the temperature equation

- This PR includes a unit-test. Current unit-tests passes because it calculates everything to zero. Need to revisit the golds.

### DIFF
--- a/amr-wind/equation_systems/temperature/source_terms/CMakeLists.txt
+++ b/amr-wind/equation_systems/temperature/source_terms/CMakeLists.txt
@@ -1,3 +1,4 @@
 target_sources(${amr_wind_lib_name} PRIVATE
 ABLMesoForcingTemp.cpp BodyForce.cpp
+HurricaneTempForcing.cpp
   )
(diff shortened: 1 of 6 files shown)


mchurchf:
Yorgos, thanks for implementing this.  My main comment is do we really need this forcing to be something that is specified separate from `hurricaneForcing`.  I'd just make this temperature forcing come on always if hurricane forcing is specified, but it can default that dT/dr = 0, so if the user doesn't input it, it has no effect.  Also, dT/dr is a constant in this implementation.  Is that true of dV/dR?  To be more realistic, we should be able to specify dV/dR and dT/dR as functions of z, if not also functions of time (but that could come as a new pull request soon).

mchurchf (commented):
See my comment in the comments section.

gdeskos:
1. Agree that this is ugly, but I am not sure if we can initialize the hurricane forcing for the temperature from the ICNS one. You see the way source terms are setup at the moment they require to separate them by the equation they are applied. If dT/dr is zero then no need to add a source term to the potential temperature equation. 
2.dT/dr is a constant for simplicity. It should be (like dV/dR) a function of z. This is easy to change. Do you have an idea of how dT/dr varies vertically (based on the data that Miguel shared with us?)

mchurchf (approved):
I approve!  Thank you, Yorgos.

mchurchf (approved):

mbkuhn (approved):

PR #983, opened 2024-02-26
2D Pitching Fixed Wing

ET14:
Add capabilities to read a pitch angle time series from user provided text file for the fixed wing actuator case.
The implementation is intended to work also for "2D" cases and still misses the implementation of a 2D Gaussian.


mbkuhn:
@ET14 great job getting this set up! I've added a few commits for the sake of passing the CI tests and correcting a print statement. If there are more changes you plan to make, please perform "git pull" before making the changes. I can also add the 2D Gaussian -- we need that for other uses as well.

amr-wind/wind_energy/actuator/wing/ActuatorWing.H
```diff
@@ -37,6 +37,15 @@ struct WingBaseData
     //! Pitch angle for the wing
     amrex::Real pitch{0.0};
 
+    //! File name for pitch actuation table
+    std::string pitch_timetable;
```
tonyinme, line 41: Let's add an identifier to the variable name, this would be best: "pitch_timetable_file"

amr-wind/wind_energy/actuator/wing/wing_ops.H
```diff
@@ -164,8 +165,20 @@ struct ComputeForceOp<
...
+
+            // Calculate the aoa based on the pitch actuation table if supplied
+            // (interpolate pitch angle at step n) or the initial pitch angle
+            amrex::Real aoa;
+            if (!wdata.pitch_timetable.empty()) {
+                const amrex::Real current_pitch = amr_wind::interp::linear(
+                    wdata.time_table, wdata.pitch_table, time.current_time());
+                amrex::Print() << "Wing Ops: Current wing pitch is "
+                               << current_pitch << " degrees.\n";
+                aoa = std::atan2(windvector[2], windvector[0]) + current_pitch;
```
tonyinme, line 177: Can you modify wdata.pitch in this step and then use the same formula that's already in the code to calculate the aoa for any case?
We would not need an else statement in that case.

tonyinme (commented):
It's looking good. Here are some minor comments.

tonyinme:
> @ET14 great job getting this set up! I've added a few commits for the sake of passing the CI tests and correcting a print statement. If there are more changes you plan to make, please perform "git pull" before making the changes. I can also add the 2D Gaussian -- we need that for other uses as well.

@mbkuhn what do you think is the best way to implement the 2D Gaussian? We would need some information to orient the Gaussian in the right direction.

amr-wind/wind_energy/actuator/wing/ActuatorWing.H
ET14: Done

amr-wind/wind_energy/actuator/wing/wing_ops.H
ET14: Done, that is indeed more concise

amr-wind/wind_energy/actuator/wing/wing_ops.H
```diff
@@ -164,8 +165,19 @@ struct ComputeForceOp<
             windvector[2] = (grid.vel[ip] - wdata.vel_tr) & blade_z;
 
             const auto vmag = vs::mag(windvector);
+
+            // Calculate the aoa based on the pitch actuation table if supplied
+            // (interpolate pitch angle at step n) or the initial pitch angle
+            amrex::Real aoa;
+            if (!wdata.pitch_timetable_file.empty()) {
+                wdata.pitch = amr_wind::interp::linear(
```
ET14, line 173: @mbkuhn At the moment the pitch angle is interpolated to the current time. Is this consistent with how the velocities, the forces and the actuator point locations are handled (between step n and n+1)? Or should one interpolate to n + 1/2?

mbkuhn: It should be interpolated to n+1/2, the average between current_time and new_time. Good question!

amr-wind/wind_energy/actuator/ActSrcLineOp.H
```diff
@@ -116,7 +116,17 @@ void ActSrcOp<ActTrait, ActSrcLine>::operator()(
             const auto pos_ip = wt * pos[ip] + (1.0 - wt) * opos[ip];
             const auto dist = cc - pos_ip;
             const auto dist_local = tmat[ip] & dist;
-            const auto gauss_fac = utils::gaussian3d(dist_local, eps[ip]);
```
ET14, line 119: @mbkuhn and @tonyinme I am not sure if the 'dist_local' and 'eps' vectors are consistent. When using the x-z plane for a 2D airfoil (spanwise direction is the y direction) and different epsilon values for the spanwise&thickness directions the thickness value appeared in the y direction rather than the z direction (and the spanwise value in the z direction rather than the y direction).
For the 2D Gaussian I flipped the order now, but we should probably check also the 3D Gaussian. Maybe I am wrong and I miss some detail here.

tonyinme: I see, yes, that makes sense. 
When I tested this a few years ago, I remember the chord direction being aligned correctly, but I did not pay attention to the other 2 directions. So it makes sense that they are inverted but I missed it.

amr-wind/wind_energy/actuator/ActSrcLineOp.H
```diff
@@ -116,7 +116,17 @@ void ActSrcOp<ActTrait, ActSrcLine>::operator()(
...
+            // If the spanwise eps is zero use the two-dimensional Gaussian
+            // The order of epsilon is (chord, thickness, span)
+            amrex::Real gauss_fac;
+            if (eps[ip].z() < 1e-12) {
```
marchdf, line 122: is there a significance to 1e-12? If not, lets use std numeric limits for small numbers

amr-wind/wind_energy/actuator/actuator_utils.H
```diff
@@ -54,6 +54,33 @@ gaussian3d(const vs::Vector& dist, const vs::Vector& eps)
...
+gaussian2d(const vs::Vector& dist, const vs::Vector& eps)
+{
+    const vs::Vector rr{
+        dist.x() / eps.x(), dist.y() / eps.y(), 0.0};
+    const amrex::Real rr_sqr = vs::mag_sqr(rr);
+
+    if (rr_sqr < 16.0) {
+        constexpr amrex::Real fac = 0.31830988618379067;
+        const amrex::Real eps_fac = eps.x() * eps.y();
+        return (fac / eps_fac) *
```
marchdf, line 77: avoid an early return and store in variable

amr-wind/wind_energy/actuator/actuator_utils.H
```diff
@@ -54,6 +54,33 @@ gaussian3d(const vs::Vector& dist, const vs::Vector& eps)
...
+ */
+AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE amrex::Real
+gaussian2d(const vs::Vector& dist, const vs::Vector& eps)
+{
+    const vs::Vector rr{
+        dist.x() / eps.x(), dist.y() / eps.y(), 0.0};
+    const amrex::Real rr_sqr = vs::mag_sqr(rr);
+
+    if (rr_sqr < 16.0) {
+        constexpr amrex::Real fac = 0.31830988618379067;
```
marchdf, line 75: This float feels important. Is it always fixed?

ET14:
@tonyinme do we want to generalize the options to 
- disable certain force components
- calculate the actuator force with the freestream velocity magnitude
- use the 2D Gaussian in "3D simulations" where another axis than the y-direction is the spanwise direction?
In its current form the implementation works for our specific problem, but is definitely not ready to be merged.

mbkuhn:
@ET14 @tonyinme I will try to get this polished up for merging soon. I'll be making direct edits to the branch, but feel free to make comments on my work.

amr-wind/wind_energy/actuator/wing/fixed_wing_ops.H
```diff
@@ -161,15 +212,17 @@ struct InitDataOp<FixedWing, ActSrcLine>
                 meta.span_locs, meta.chord_inp, wx, meta.chord);
             // clang-format off
             meta.epsilon_chord = {
-                meta.epsilon_chord.x(), meta.epsilon_chord.z(),
-                meta.epsilon_chord.y()};
+                meta.epsilon_chord.x(), meta.epsilon_chord.y(),
+                meta.epsilon_chord.z()};
```
mbkuhn, line 216: @ET14 I think this is why you were seeing that mismatch in the epsilon components. The code was switching the place of the y and z components when it relied on epsilon_chord. I have to check this, though.

mbkuhn: I was wrong about this change, but I was close. This reordering operation is necessary, but it was not being performed for eps_inp. Fixed.

mbkuhn:
> @tonyinme do we want to generalize the options to
> 
>     * disable certain force components
> 
>     * calculate the actuator force with the freestream velocity magnitude
> 
>     * use the 2D Gaussian in "3D simulations" where another axis than the y-direction is the spanwise direction?
>       In its current form the implementation works for our specific problem, but is definitely not ready to be merged.

I generalized the first two, and I mostly generalized the third one. The 2D implementation assumes that the dimension not being used is the spanwise dimension, but it does not have to be y.

mbkuhn:
@ET14 I believe I have generalized the capabilities from your work. I put a new reg test in test/test_files/act_pitching_wing_2D that should serve as an example for the input arguments. Please try out that case and see if it does what you need it to. It would be great if you could add a pitch_timetable file to that test case, too. I will probably add a unit test or two before merging, but I think it's ready otherwise. By the way, the "get_either" and "query_either" changes are for something Tony requested -- they weren't needed to get this to work.

ET14:
@mbkuhn Thank you for restructuring this, looks great! I will test all the flags again and also add a pitch input table.

amr-wind/wind_energy/actuator/ActSrcLineOp.H
mbkuhn: @ET14 this was a good catch. For the sake of the vector and tensor operations in the code, the local coordinates are (chord, span, thickness), which is the way dist_local and eps are laid out when used in ActSrcLineOp, but the convention for specifying epsilon is (chord, thickness, span). Therefore, the thickness and span components need to switch places after being input and before being used. The problem was that for the fixed wing, this switch was not taking place for eps_inp, which explains the mismatch you were seeing. Corrected now.

ET14: @mbkuhn Another thought related to this. I think the transformation tmat is only calculated here (https://github.com/Exawind/amr-wind/blob/02feaf5699dcc75a124a6361178b1217c49f8d76/amr-wind/wind_energy/actuator/wing/wing_ops.cpp#L43) during the init phase with the initial pitch angle. However, now the pitch angle is time dependent. Should we update tmat at each time step?

mbkuhn: Ah yes, great catch. By the looks of it, the pitch is only used at initialization for this. This should be clear if you set epsilon to be different in chordwise and thickness directions and pitch the wing -- the Gaussian won't pitch with it, even though the code should still be calculating the angle of attack correctly. the grid.orientation variable needs to be updated when the pitch is updated in the ComputeForceOp operator within wing_ops.H.

amr-wind/wind_energy/actuator/wing/ActuatorWing.H
```diff
@@ -37,6 +37,24 @@ struct WingBaseData
     //! Pitch angle for the wing
     amrex::Real pitch{0.0};
 
+    //! Switch for turning on 2D Gaussian (off in spanwise)
+    bool gauss_2D{false};
+
+    //! Switch for turning off force components of actuator
+    vs::Vector fcoord_flags{1.0, 1.0, 1.0};
```
marchdf, line 44: can we make this member var slightly more verbose? `force_coord_flags` ? And it might make more sense to make it a `Vector<int>`?

amr-wind/wind_energy/actuator/wing/fixed_wing_ops.H
```diff
@@ -36,16 +36,47 @@ struct ReadInputsOp<FixedWing, ActSrcLine>
...
+            pp.get("pitch", wdata.pitch);
+        }
         pp.get("airfoil_table", wdata.airfoil_file);
         pp.query("airfoil_type", wdata.airfoil_type);
         pp.queryarr("span_locs", wdata.span_locs);
         pp.queryarr("chord", wdata.chord_inp);
         bool use_fllc = false;
         pp.query("fllc", use_fllc);
 
+        amrex::Vector<int> fcf, dcf;
```
marchdf, line 52: I don't immediately see where fcf and dcf are being used. If they are being used, let's move this to where they are being used. If not, let's remove.

amr-wind/wind_energy/actuator/wing/fixed_wing_ops.H
```diff
@@ -85,17 +116,53 @@ struct ReadInputsOp<FixedWing, ActSrcLine>
...
+        amrex::Real maxpy = amrex::max(p1.y(), p2.y());
+        amrex::Real minpz = amrex::min(p1.z(), p2.z());
+        amrex::Real maxpz = amrex::max(p1.z(), p2.z());
+        // Bounding box limits for 2D actuator
+        if (wdata.gauss_2D) {
+            // Assume largest component of span needs entire domain, but the
+            // others do not need the bounding box to grow
+            const auto wspan = wdata.end - wdata.start;
+            // Flag for checking periodicity in spanwise direction
+            int warn_per = -1;
```
marchdf, line 132: feels like this should be a bool since we don't care what the value is. And then the checks below become much simpler.

amr-wind/wind_energy/actuator/wing/fixed_wing_ops.H
```diff
@@ -85,17 +116,53 @@ struct ReadInputsOp<FixedWing, ActSrcLine>
...
+            auto per = data.sim().mesh().Geom()[0].periodicity();
+            if (std::abs(wspan.x()) >=
+                std::max(std::abs(wspan.y()), std::abs(wspan.z()))) {
+                // Biggest component is in x, grow bounding box
+                minpx = data.sim().mesh().Geom(0).ProbLoArray()[0];
+                maxpx = data.sim().mesh().Geom(0).ProbHiArray()[0];
+                // Check periodicity
```
marchdf, line 139: remove this and the comments like it below. The code is obvious.

amr-wind/wind_energy/actuator/wing/wing_ops.H
```diff
@@ -172,7 +190,35 @@ struct ComputeForceOp<
...
 
             // Assume unit chord
-            const auto qval = 0.5 * vmag * vmag * chord[ip] * dx[ip];
+            // If the spanwise eps is zero use the two-dimensional Gaussian
+            // The order of epsilon is (chord, thickness, span)
+            // For the 2D Gaussian the force is normalized on addition with the
+            // wing span in order to conserve the total applied force
+
+            // Calculate factor for qval: 0.5 * Uinf * Uinf * dx
+            auto qfac = 0.5 * vmag * vmag;
```
marchdf, line 199: make a ternary expression with the the if condition on line 200.

amr-wind/wind_energy/actuator/wing/wing_ops.H
```diff
@@ -186,6 +232,9 @@ struct ComputeForceOp<
             // Compute force on fluid from this section of wing
             grid.force[ip] = -(lift_dir * lift + drag * drag_dir);
 
+            // In global coords, zero disabled force components
+            grid.force[ip] = grid.force[ip] * fcoord_flags;
```
marchdf, line 236: could be a `*=`

test/test_files/act_pitching_wing_2D/sineWave.txt
```diff
@@ -0,0 +1,401 @@
+0.000000 4.000000
```
marchdf, line 1: Silly but the style of naming for this file doesn't match the others in this directory. Make it `sine_wave.txt` ?

marchdf (changes requested):

amr-wind/wind_energy/actuator/wing/fixed_wing_ops.H
mbkuhn: good catch

amr-wind/wind_energy/actuator/wing/fixed_wing_ops.H
mbkuhn: done. I meant to use this to change the warning depending on the direction but it was a bad approach anyway

amr-wind/wind_energy/actuator/wing/fixed_wing_ops.H
mbkuhn: done

amr-wind/wind_energy/actuator/wing/wing_ops.H
mbkuhn: done

amr-wind/wind_energy/actuator/wing/wing_ops.H
mbkuhn: avoided this initially because the vs::Vector type didn't have it as an operator. added it now

test/test_files/act_pitching_wing_2D/sineWave.txt
mbkuhn: done

amr-wind/wind_energy/actuator/ActSrcLineOp.H
mbkuhn: almost forgot to go back and fix this. the transformation matrix now updates every compute force step, and the tilting of the force distribution is obvious when an anisotropic epsilon distribution is specified.

amr-wind/wind_energy/actuator/wing/wing_ops.H
```diff
@@ -141,22 +137,43 @@ struct ComputeForceOp<
             grid.pos, wdata.vel_tr, npts, time.current_time(), time.new_time(),
             wdata.motion_type, wdata.s_period, wdata.s_vector);
 
+        // Build the local reference frame
+        vs::Vector wspan = wdata.end - wdata.start;
```
marchdf, line 141: const?

amr-wind/wind_energy/actuator/wing/fixed_wing_ops.H
```diff
@@ -85,17 +107,53 @@ struct ReadInputsOp<FixedWing, ActSrcLine>
...
+        amrex::Real minpz = amrex::min(p1.z(), p2.z());
+        amrex::Real maxpz = amrex::max(p1.z(), p2.z());
+        // Bounding box limits for 2D actuator
+        if (wdata.gauss_2D) {
+            // Assume largest component of span needs entire domain, but the
+            // others do not need the bounding box to grow
+            const auto wspan = wdata.end - wdata.start;
+            // Flag for checking periodicity in spanwise direction
+            bool warn_per{false};
+            std::string sdir{" "};
```
marchdf, line 124: you shouldn't need to init like this. `std::string sdir` is enough.

amr-wind/wind_energy/actuator/wing/wing_ops.H
```diff
@@ -141,22 +137,43 @@ struct ComputeForceOp<
...
+
+        // Use the global coord to orient the blade
+        // The default is for inflow in the x direction
+        auto blade_x = wdata.blade_x.unit();
+        auto blade_y = wspan.unit();
+        auto blade_z = (blade_x ^ blade_y).unit();
+        // Ensure orthogonality for new reference frame
+        blade_x = (blade_y ^ blade_z).unit();
```
marchdf, line 149: should we just `auto blade_x = (blade_y ^ blade_z).unit();` and remove lin 145? Should all these be const?

marchdf (changes requested):
Nice work! I left some minor suggestions.

amr-wind/wind_energy/actuator/wing/wing_ops.H
mbkuhn: nvm, you can substitute line 145 and make them all const

mbkuhn:
Comments addressed

marchdf (approved):

PR #987, opened 2024-03-04
Turbulent viscosity Smagorinsky in multiphase

tonyinme:
When conducting wave simulations in AMR-Wind involving sea surface interactions, the viscosity near the interaction region can exhibit unusually high values. This phenomenon arises from the Volume of Fluid (VOF) method, where viscosity is multiplied by density. Given that water density is approximately 1000 times that of air, this multiplication significantly amplifies the turbulent viscosity. Simulations of ABL + waves with the Smagorinsky model do not work because the high values of viscosity require too small a time-step, and the simulation becomes unstable and crashes. 

The main challenge comes from separating the two phases for density and viscosity but using the gradients of the entire flow field.  In the case of volume fractions between 0 and 1, the gradients of the flow will be dominated by the fastest moving flow (the air). The high gradients will create an artificially high turbulent viscosity that will get multiplied by the artificially high turbulent viscosity. For this reason, it makes more sense to only use the turbulent viscosity for air whenever the volume fraction is less than one. This will ensure that the turbulent viscosity will not be artificially higher when there is water in the cell.

This code is one of the ideas I have to mitigate this problem. 
It involves finding the minimum density in the flow and assigning it whenever the volume fraction is less than one. 

I am looking to start a conversation around solving this issue and we can come up with other approaches as well.

### DIFF
--- a/amr-wind/turbulence/LES/Smagorinsky.cpp
+++ b/amr-wind/turbulence/LES/Smagorinsky.cpp
@@ -34,7 +34,7 @@ void Smagorinsky<Transport>::update_turbulent_viscosity(
     const amrex::Real Cs_sqr = this->m_Cs * this->m_Cs;
 
     const bool is_vof = this->m_sim.repo().field_exists("vof");
-    amrex::Real rho_min=0.;
+    amrex::Real rho_min = 1e9;
 
     const auto& vof = this->m_sim.repo().get_field("vof");
     if (is_vof) {
--- a/amr-wind/turbulence/LES/Smagorinsky.cpp
+++ b/amr-wind/turbulence/LES/Smagorinsky.cpp
@@ -7,6 +7,8 @@
 #include "AMReX_MultiFab.H"
 #include "AMReX_ParmParse.H"
 #include "amr-wind/equation_systems/vof/volume_fractions.H"
+#include "amr-wind/fvm/filter_harmonic.H"
+#include "amr-wind/fvm/gradient.H"
 
 namespace amr_wind {
 namespace turbulence {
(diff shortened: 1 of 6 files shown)


mbkuhn:
> The high gradients will create an artificially high turbulent viscosity that will get multiplied by the artificially high turbulent viscosity. 

For the record, this should say "The high gradients will create an artificially high turbulent [kinematic] viscosity (due to the presence of air in current and nearby cells) that will get multiplied by the artificially high density (due to the presence of water in the current cell)."

mbkuhn:
Even though this approach is nice because it doesn't require references to other parts of the code, it seems like overkill to do a global operation, like the minimum, especially because it does it every time. Here are some other thoughts that may work via a different implementation.

1. reference MultiPhase physics to get the air density upon initialization (sim is available when initializing). Then use this as the minimum density as you have it.
2. reference MultiPhase physics to get the air density and water density. Then use a harmonic average instead of an arithmetic one to calculate the density that will be multiplied. 1/rho = vof/rho_liquid + (1-vof)/rho_gas.
3. instead of using the local density, average the density over the strainrate stencil (requires no additional references). This could be with an arithmetic average or a harmonic one (though we may need to do a harmonic average to get the desired effect). This could be turned on via a switch detecting the vof variable, though it would have no effect on constant density simulations.

Thoughts?

amr-wind/turbulence/LES/Smagorinsky.cpp
```diff
@@ -52,9 +63,20 @@ void Smagorinsky<Transport>::update_turbulent_viscosity(
...
+            const auto& vof_arr = vof(lev).const_array(mfi);
+
             amrex::ParallelFor(
                 bx, [=] AMREX_GPU_DEVICE(int i, int j, int k) noexcept {
-                    const amrex::Real rho = rho_arr(i, j, k);
+                    amrex::Real rho;
+
+                    const bool is_air = vof_arr(i, j, k) < 1.;
+                    //
+                    if (is_vof) {
```
mbkuhn, line 74: I think I know now why this didn't work! you're using the minimum density whenever the vof variable is present (is_vof). You meant to do that only when there is air in the current cell (is_air).

tonyinme: This has been updated in the latest version.

tonyinme:
We have tried a few different strategies to mitigate the high turbulent viscosities. 
We tried the harmonic averaging, but that still had a few cells with too high of a viscosity. 

The latest version uses only the air density whenever there is air (vor<1) and the gradient of vof is 0. 
This approach ensures that the gradients of velocity in the cell will not be affected by the higher gradients from nearby cells. 

@mbkuhn , let me know what further improvements we can make to this approach.

hgopalan:
You can also check for the sum of shear production and  buoyancy production due to density. If the sum is negative you can set turbulent viscosity to zero. RANS models add this extra production term to drive turbulent viscosity to zero near the interface. 

WRF uses a similar approach with the Smagorinsky model and uses the Brunt–Vasala frequency to turn off turbulent viscosity for certain stable conditions. 

Buoyancy production reference: https://www.sciencedirect.com/science/article/pii/S0378383917302107

PR #1019, opened 2024-04-12
Ocean Waves Improvements, esp. Stokes Waves

mbkuhn:
## Summary

Bug fix for stokes waves, adding feature of wave length calculation, generalizing and de-hardcoding ocean waves code

## Pull request type

Bugfix + Feature

Please check the type of change your PR introduces:

## Checklist

The following should accompany this PR:

- [ x] new unit-test(s)
- [ x] new regression test(s) pass

- the unit tests
  - [ ] on GPU 
  - [x] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU 

## Additional background

With small water depths, the stokes waves velocities seemed reasonable before the bug fix, but with a depth of 10 m, the velocities in x were on the order of 1e20, due to the mix-up of variables m and n.

Issue Number: N/A

### DIFF
--- a/amr-wind/ocean_waves/relaxation_zones/RelaxationZones.H
+++ b/amr-wind/ocean_waves/relaxation_zones/RelaxationZones.H
@@ -18,6 +18,8 @@ struct RelaxZonesBaseData
 
     amrex::Real water_depth{0.5};
 
+    amrex::Real g{9.81};
+
     // Wave generation/absorption parameters
     amrex::Real gen_length{4.0};
 
--- a/amr-wind/ocean_waves/relaxation_zones/RelaxationZones.H
+++ b/amr-wind/ocean_waves/relaxation_zones/RelaxationZones.H
@@ -18,7 +18,8 @@ struct RelaxZonesBaseData
 
     amrex::Real water_depth{0.5};
 
-    amrex::Real g{9.81};
+    // Gravitational constant - wave types that use gravity store value here
+    amrex::Real g;
 
     // Wave generation/absorption parameters
     amrex::Real gen_length{4.0};
--- a/amr-wind/ocean_waves/relaxation_zones/stokes_waves_K.H
+++ b/amr-wind/ocean_waves/relaxation_zones/stokes_waves_K.H
@@ -35,10 +35,9 @@ AMREX_FORCE_INLINE amrex::Real stokes_wave_length(
     if (k < tol || iter >= iter_max || std::isnan(k)) {
         // Return negative wavelength if faulty result
         return -1;
-    } else {
-        // Return wavelength calculated from wavenumber
-        return 2.0 * M_PI / k;
     }
+    // Return wavelength calculated from wavenumber
+    return 2.0 * M_PI / k;
 }
 
 AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE void stokes_coefficients(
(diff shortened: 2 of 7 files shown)


amr-wind/ocean_waves/relaxation_zones/RelaxationZones.H
```diff
@@ -18,6 +18,8 @@ struct RelaxZonesBaseData
 
     amrex::Real water_depth{0.5};
 
+    amrex::Real g{9.81};
```
marchdf, line 21: do we need another variable that carries gravity? Can we hijack it from somewhere else?

amr-wind/ocean_waves/relaxation_zones/stokes_waves_K.H
```diff
@@ -4,8 +4,44 @@
...
+    const amrex::Real omega = 2.0 * M_PI / T;
+    const amrex::Real lambda0 = (g * T * T) / 2.0 / M_PI;
+
+    // Begin Newton-Raphson iterations
+    constexpr amrex::Real tol = 1e-10;
+    constexpr int iter_max = 20;
+    int iter = 0;
+    amrex::Real k = omega / std::sqrt(g);
+    amrex::Real f = g * k * std::tanh(k * d) - omega * omega;
+    while (f > 1e-10 && iter < iter_max) {
```
marchdf, line 26: was this 1e-10 meant to use `tol` above? In general should we have a more uniform way of specifying tolerances. Feels like we have this and `eps` in a bunch of different places. Not necessarily a "now" thing...

amr-wind/ocean_waves/relaxation_zones/stokes_waves_K.H
```diff
@@ -284,65 +324,65 @@ AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE void stokes_waves(
     // Compute cosines
     const amrex::Real cc11 = my_cosh_cos(
```
marchdf, line 325: can we remove the `my_` ?

amr-wind/ocean_waves/relaxation_zones/stokes_waves_K.H
mbkuhn: thanks for noticing that. I agree that it would be good to have a more standardized reference for tolerances, but often the level of tolerance required depends on the context. So I would think that ad-hoc tolerances would still be used.

amr-wind/ocean_waves/relaxation_zones/RelaxationZones.H
mbkuhn: interested in what you think of my approach. Right now, gravity doesn't belong to any specific part of the code. It is read in with the parser prefix "incflo", but basically each part of the code that uses it parses it directly. I'm trying to get away from that, so in my own development efforts related to waves I have created a gravity variable that belongs to the MultiPhase physics class, and other parts of the code that need gravity get it from that class. However, in the RelaxationZones structure, the functions that need gravity don't have access to the sim variable and therefore don't have access to physics_manager or the MultiPhase class. This g variable is to store the value so it is accessible in the functions where it's needed, but the value is obtained through the MultiPhase class.

mbkuhn:
@gdeskos would you be able to take a glance at this PR, mainly the function that calculates wavelength?

amr-wind/ocean_waves/relaxation_zones/RelaxationZones.H
marchdf: yeah that seems like a fine solution

marchdf (approved):

gdeskos (commented):
@mbkuhn good job fixing the bugs in the original stokes waves code! The code inside the function that calculates the wavelength looks good. I assume you are solving equation 13 of the Fenton 1985 paper.

PR #1027, opened 2024-04-23
Disable ABL wall function for a zhi BC

### LINKED ISSUE #803: ABL physics wall_model at zhi is using the wrong zref
Currently for the MO-law there seems to be just one zref, which is either input file specified or the first cell height from the bottom. When the wall model option is specified for zhi, it seems to use the same bottom zref at the top too, which is incorrect.

marchdf:
## Summary

Disable the Moeng wall model on zhi.

## Pull request type

Please check the type of change your PR introduces:

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

## Checklist

- the unit tests
  - [x] on GPU 
  - [x] on CPU 
- the regression tests
  - [x] on GPU 
  - [x] on CPU 

[gpu_tests.txt](https://github.com/Exawind/amr-wind/files/15080534/gpu_tests.txt)
[cpu_tests.txt](https://github.com/Exawind/amr-wind/files/15080535/cpu_tests.txt)

## Additional background

Closes #803

### DIFF
--- a/amr-wind/wind_energy/ABLWallFunction.cpp
+++ b/amr-wind/wind_energy/ABLWallFunction.cpp
@@ -175,6 +175,11 @@ void ABLVelWallFunc::wall_model(
         return;
     }
 
+    if ((velocity.bc_type()[zhi] == BC::wall_model) &&
+        (m_wall_shear_stress_type == "moeng")) {
+        amrex::Abort("The Moeng wall model is not applicable for a zhi BC");
+    }
+
     for (int lev = 0; lev < nlevels; ++lev) {
         const auto& geom = repo.mesh().Geom(lev);
         const auto& domain = geom.Domain();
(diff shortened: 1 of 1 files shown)


moprak-nrel:
We may need to disable this for more than just the `moeng` shear stress, anything that uses quantities from  `MOData` will be incorrect at `zhi` since it will be using the quantities computed at `zlo` like: `utau`, `wspd_mean`, etc. I suspect the right fix might be to disable wall models for `zhi` entirely in `ABLWallFunction`. Example in the `Schumann` wall model: https://github.com/Exawind/amr-wind/blob/38dfd817bace4e6d864d9c0b97ab01e9ff2b7e47/amr-wind/wind_energy/ShearStress.H#L92-L103

amr-wind/wind_energy/ABLWallFunction.cpp
```diff
@@ -170,11 +170,18 @@ void ABLVelWallFunc::wall_model(
...
-    if ((velocity.bc_type()[zlo] != BC::wall_model) &&
-        (velocity.bc_type()[zhi] != BC::wall_model)) {
+    if (velocity.bc_type()[zhi] == BC::wall_model) {
+        amrex::Abort("ABL wall models are not applicable to a zhi BC");
+    }
+    if (velocity.bc_type()[zlo] != BC::wall_model) {
         return;
     }
 
+    if ((velocity.bc_type()[zhi] == BC::wall_model) &&
```
moprak-nrel, line 180: this branch is not needed anymore since there's a more general `Abort` above

moprak-nrel (changes requested):

amr-wind/wind_energy/ABLWallFunction.cpp
marchdf: good catch

moprak-nrel (approved):

PR #1035, opened 2024-04-30
default ability of Field Norms to mask redundant grids

mbkuhn:
## Summary

The field norm utility previously summed over all cells indiscriminately. These changes allow the field norm utility to sum over only the finest cells locally and make this the default behavior. The old form can still be used via an input flag.

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [x] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [x] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU


amr-wind/utilities/sampling/FieldNorms.cpp
```diff
@@ -43,23 +45,45 @@ amrex::Real FieldNorms::L2_norm(amr_wind::Field& field, const int comp)
     const auto& geom = field.repo().mesh().Geom();
     constexpr int nghost = 0;
 
+    const auto& mesh = field.repo().mesh();
+
     for (int lev = 0; lev <= finest_level; lev++) {
```
marchdf, line 50: Feels like we shouldn't roll our own: https://github.com/AMReX-Codes/amrex/blob/development/Src/Base/AMReX_MultiFab.H#L226. If those don't have masked versions (I think only norm0) does, maybe it makes more sense to make a PR to amrex?

mbkuhn: It's unclear to me if norm2 does, it seems like it doesn't, but there is a mention of an overlapmask. https://github.com/AMReX-Codes/amrex/blob/development/Src/Base/AMReX_MultiFab.cpp#L1069
norm0 appears to rely on norminf from the FabArray definition itself.

If it's this hard to tell what the functionality is, I think it would make sense to have it in AMR-Wind.

moprak-nrel (approved):

PR #1037, opened 2024-05-02
VOF operations changes

mbkuhn:
## Summary

* deprecate vof-only sharpening
* add replace mask
* make VOF advection into an operator applied from n to n+1

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [x] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [x] on CPU 
- the regression tests
  - [ ] on GPU 
  - [x] on CPU 

## Additional background

Trying to keep the PRs small for incorporating the multiphase hybrid changes into main.


mbkuhn:
@moprak-nrel or @gdeskos, could you give your approval to this? Most of the changes are only to do with multiphase hybrid work, which only affects me. The one aspect of the functionality that still applies during normal standalone operation is that the VOF advection step doesn't work with only the new state of the variable, it starts with the old, applies operations to the nph state as a working copy, then updates the new state. I ran the reg tests and they all passed.

gdeskos (approved):

Issue #1038, opened 2024-05-03
FLLC Reintroduced bug


tonyinme:
There was a bug introduced [here](https://github.com/Exawind/amr-wind/blame/bdddf133e41a9b7b4c8ce28f1ea1bebec47678f5/amr-wind/wind_energy/actuator/turbine/fast/turbine_fast_ops.H#L246).

This looks like it was an issue that had been fixed [here](https://github.com/Exawind/amr-wind/pull/894/files) but was reintroduced as part of another PR.

mbkuhn:
I'm pretty sure that this bug was fixed by #1043. @tonyinme @marchdf can we confirm?

PR #1039, opened 2024-05-03
Get number of points along the blade

tonyinme:
## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [ ] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU 

## Additional background

This was a former bugfix that was reintroduced. 
This is a re-implementation of the bug fix. 

[Issue](https://github.com/Exawind/amr-wind/issues/1038)

### DIFF
--- a/amr-wind/wind_energy/actuator/turbine/fast/turbine_fast_ops.H
+++ b/amr-wind/wind_energy/actuator/turbine/fast/turbine_fast_ops.H
@@ -243,7 +243,7 @@ struct InitDataOp<TurbineFast, SrcTrait>
         amrex::ParallelDescriptor::Comm_dup(
             amrex::ParallelDescriptor::Communicator(), tdata.tcomm);
 
-        int sz_info[4]{0, 0, 0, 0};
+        int sz_info[5]{0, 0, 0, 0, 0};
         if (info.is_root_proc) {
             tdata.fast->init_turbine(tdata.fast_data.tid_local);
 
@@ -252,6 +252,7 @@ struct InitDataOp<TurbineFast, SrcTrait>
             sz_info[1] = tf.to_cfd.fx_Len;
             sz_info[2] = tf.from_cfd.u_Len;
             sz_info[3] = tf.num_pts_tower;
+            sz_info[4] = tf.num_blade_elem;
         }
 
         // Broadcast data to everyone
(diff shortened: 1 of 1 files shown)


amr-wind/wind_energy/actuator/turbine/fast/turbine_fast_ops.H
```diff
@@ -243,7 +243,7 @@ struct InitDataOp<TurbineFast, SrcTrait>
         amrex::ParallelDescriptor::Comm_dup(
             amrex::ParallelDescriptor::Communicator(), tdata.tcomm);
 
-        int sz_info[4]{0, 0, 0, 0};
+        int sz_info[5]{0, 0, 0, 0, 0};
```
marchdf, line 246: I would be inclined to make this an `amrex::Vector<int> sz_info(5, 0)`. Also, is there an assert you can add somewhere to check that it was correctly populated? Maybe below on line 255? I would hate for this to happen again.

marchdf: Different type, like a pointer? You would get that with `sz_info.data()`. Then your assert would just be `sz_info.size() == 5`

tonyinme: The problem was not in the assert. It is in another fast function where it gets called. 
This is the error
`/home/lmartine/amr-wind-tony/amr-wind/wind_energy/actuator/turbine/fast/turbine_fast_ops.H:264:52: error: no matching function for call to 'Bcast(amrex::Vector<int>&, int, const int&, MPI_Comm&)'
             sz_info, 3, info.root_proc, tdata.tcomm);`

We need to leave it as `int sz_info[5]{0, 0, 0, 0, 0};`

marchdf: Lets drop this, I have an upcoming PR that will fix all these.

marchdf:
There is actually another bug here. I am moving these fixes to another PR and adding the other fix I found.

marchdf:
@tonyinme can you check to make sure I moved the fixes right? It will be #1043 and it will be in `amr-wind/wind_energy/actuator/turbine/fast/turbine_fast_ops.H`. You can also see what I mean by using `.begin()` to get around that type error you were seeing. And the bugfix I was talking about was the MPI broadcast had the wrong size for `sz_info`.

tonyinme:
> @tonyinme can you check to make sure I moved the fixes right? It will be #1043 and it will be in `amr-wind/wind_energy/actuator/turbine/fast/turbine_fast_ops.H`. You can also see what I mean by using `.begin()` to get around that type error you were seeing. And the bugfix I was talking about was the MPI broadcast had the wrong size for `sz_info`.


Yes, I went over it and the implementation looks good.

PR #1045, opened 2024-05-07
Fllc non uniform

tonyinme:
## Summary

Non-uniform distribution of points for the FLLC to reduce computational cost.
This code will create a non-uniform distribution of actuator points based on the condition that epsilon divided by actuator spacing is less that a specified value. 
This is ready for review. 

## Pull request type

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [ ] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU 

## Additional background

Issue Number:


amr-wind/wind_energy/actuator/FLLC.H
```diff
@@ -33,6 +33,13 @@ struct FLLCData
     bool different_sizes;
     RealList span_distance_vel;
     RealList span_distance_force;
+
+    // non-uniform variables
+    bool nonu{true};    // non-uniform flag
```
psakievich, line 38: If you need two comments to explain what a boolean variable is then the name is not good. Abbreviating is not doing anyone any favors here. I would just call in `nonuniform` and delete the comments.

amr-wind/wind_energy/actuator/FLLC.H
```diff
@@ -33,6 +33,13 @@ struct FLLCData
...
+    amrex::Real eps_dr; // the ratio of epsilon to actuator width
+    RealList r_n;       // non-uniform radius
+    RealList dr_n;       // non-uniform radius
+    VecList lift_n;     // non-uniform lift
```
psakievich, line 42:
```suggestion
    RealList nonuniform_r;
    RealList nonuniform_dr; 
    VecList nonuniform_lift; 
```
Same thing here. This is much clearer when we get to the actual math. Consider your question to me on slack vs my reply
``` c++
// question
eps =  interp::linear(r_, data.span_distance_vel, data.optimal_epsilon); (edited) 

// reply
eps_nonuniform = interp::linear(r_uniform, eps_optimal, r_nonuniform) 
```

The first one it is hard to interpret and check because of the names. The second is much cleaner b/c of the names.

amr-wind/wind_energy/actuator/FLLC.cpp
```diff
@@ -38,6 +39,45 @@ void FLLCInit(
     data.dr[0] = vs::mag(view.pos[1] - view.pos[0]) / 2.;
     data.dr[npts - 1] = vs::mag(view.pos[npts - 1] - view.pos[npts - 2]) / 2.;
 
+    namespace interp = ::amr_wind::interp;
+
+    // Non-uniform radial distribution variables
+    size_t idx = 0;
+    amrex::Real dr;
```
psakievich, line 46: Same thing here. Please give more descriptive names of the variables. It is hard to review `dr` `dr1` and `dr2` because I have to keep in my mind what they all mean. If you name the variables based on what they actually are you will need less comments and it will be easier to understand.

Comments are helpful, but if you have to annotate every line of code to explain what is going on, then it means the code is confusing, you have to read twice as much and do more work. 

1) Read comments
2) read code
3) check the code does what the comments say
4) confirm that is the right behavior 

versus

1) read the code
2) read the comments when they are there to clarify the code
2) confirm the behavior is right

This compounds over time and as the code base grows.

psakievich (changes requested):
Nice start @tonyinme! I see the concept, and it looks good. I would like to see the code cleaned up more before really looking deeper. See comments.

amr-wind/wind_energy/actuator/FLLCOp.H
```diff
@@ -45,6 +45,7 @@
         auto& u_les = fllc.les_velocity;
         auto& u_opt = fllc.optimal_velocity;
         auto& G = fllc.lift;
+        auto& nG = fllc.lift_n; // non-uniform G
```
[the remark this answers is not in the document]
tonyinme: Fixed in latest version.

amr-wind/wind_energy/actuator/FLLC.H
tonyinme: Agreed, fixed in latest version.

amr-wind/wind_energy/actuator/FLLC.H
tonyinme: Agreed, fixed in latest version.

amr-wind/wind_energy/actuator/FLLC.cpp
tonyinme: I modified most variable names to account for this. 
Some variables are temporary and I was not sure what to name them. 
Let me know if you prefer another naming convention for specific variables like dr1 and dr2.

tonyinme:
Thanks @psakievich for your comments. 
I have addressed most of them in the latest version. 
This new version is ready for review.

mbkuhn:
Hey @tonyinme, this doesn't look very ready for review with all of the commented/uncommented print statements littered throughout. Are you wanting it to be checked conceptually before cleaning it up? Also, if you apply the clang formatting, that will enable the PR to go through the rest of the CI checks.

If you're still actively testing it, too, that's fine; just let us know.

tonyinme:
@mbkuhn this should be ready to review. 
I think I forgot to push the latest version but it is now all in.

amr-wind/wind_energy/actuator/FLLCOp.H
```diff
@@ -191,12 +193,27 @@ struct FLLCOp
...
                 std::max(vs::mag(vel), vs::DTraits<amrex::Real>::eps());
             const auto vmag2 = vmag * vmag;
-
+            fllc.vel_rel[ip] = vel;
             const auto fv = force & vel;
 
             G[ip] = (force - vel * fv / vmag2) / dr;
         }
 
+        // Get the non-uniform distribution of force
```
psakievich, line 202: I think this needs conditionals around it if the non-uniform spacing is optional.  It seems like the non-uniform points wouldn't be populated.

amr-wind/wind_energy/actuator/FLLC.cpp
```diff
@@ -49,6 +112,8 @@ void FLLCParse(const utils::ActParser& pp, FLLCData& data)
     std::string typeString = "variable_chord";
     pp.query("fllc_type", typeString);
     data.correction_type = FLLCTypeMap.at(typeString);
+    pp.query("fllc_nonuniform", data.nonuniform);
+    pp.query("fllc_eps_dr", data.eps_dr);
```
psakievich, line 116: A bit nit picky,but this input variable does not match with the others in terms of abbreviation.  For users it is nice to keep the same level and types of abbreviations so `fllc_epsilon_dr`. But I am also not really a fan of this name either.  This is a ratio right?  I have to keep looking in the code to see how it is used to understand what it is.  I think a longer, more descriptive name would be preferred.

psakievich (commented):
It is not clear to me that there is a strong case for making this an option if the behavior is so bad without it.  I would consider removing the boolean flag altogether.

amr-wind/wind_energy/actuator/FLLC.cpp
```diff
@@ -38,6 +42,70 @@ void FLLCInit(
...
+    // Non-uniform radial distribution calc
+    data.nonuniform_r.push_back(
+        data.r[0]
+); // Initialize the first location
+//amrex::OutStream() << "Error NOT Here nonu back " << data.nonuniform_r.back()  << std::endl;
+//amrex::OutStream() << "Error NOT Here back " << data.r.back()  << std::endl;
+    while (data.nonuniform_r.back() < data.r.back()) {
+        amrex::Real r_ = data.nonuniform_r.back(); // The latest radial position
+        // Interpolate the value of epsilon to the current radial location
+        eps1 = interp::linear(data.r, data.optimal_epsilon, r_);
```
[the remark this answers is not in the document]
tonyinme: This should be interpolating the value at the last radial location `r_` in every iteration. 
This needs to be done inside the loop because the value keeps changing. 
Every iteration in this loop will append a new radial location. That is how the nonuniform grid is constructed. 
The linear interpolation function is overloaded and in this case it should return only one value.

amr-wind/wind_energy/actuator/FLLC.H
```diff
@@ -33,6 +34,16 @@ struct FLLCData
     bool different_sizes;
     RealList span_distance_vel;
     RealList span_distance_force;
+
+    // non-uniform variables
+    bool nonuniform{true};      // non-uniform flag
```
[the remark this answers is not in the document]
tonyinme: I think it is best to leave the original/default behavior. 
It would help in testing/debugging and it is consistent with the original derivations in the manuscript. 
If the ALM data structure had the non-uniform distribution, then we would not need to use this non-uniform option and it could default to the old behavior.

amr-wind/wind_energy/actuator/FLLC.cpp
tonyinme: Yes, I was trying to leave the same format for user inputs so that they know that this is a parameter for the fllc. 
`eps_dr` is the ratio `epsilon/dr` that is used to generate the non-uniform spacing. 
Would `fllc_epsilon_dr_ratio` be a better name?

amr-wind/wind_energy/actuator/FLLCOp.H
tonyinme: Added conditional in latest commit.

amr-wind/wind_energy/actuator/FLLC.cpp
psakievich: Yeah I like that better

psakievich (approved):

mbkuhn:
@tonyinme I looked at the merge conflict, and it seems pretty simple to fix. I think the conflict came from an effort to use amrex arrays. Also, I don't see "assert(" elsewhere in the code, so I think the amrex version is probably preferable. I'll make comments directly for those changes.

amr-wind/wind_energy/actuator/FLLC.cpp
```diff
@@ -38,6 +43,64 @@ void FLLCInit(
...
+            dr1 = eps1 / data.eps_dr; // spacing needed to match condition
+
+            // Interpolate the value of epsilon to the next radial location
+            eps2 = interp::linear(data.r, data.optimal_epsilon, r_ + dr1);
+            dr2 = eps2 / data.eps_dr;
+
+            // This will ensure that the spacing will always meet the
+            // requirement eps/dr
+            dr = std::min(dr1, dr2);
+            assert(dr > 0.);
```
mbkuhn, line 77: amrex assert function & it's better to compare floats to a tiny number
```suggestion
            AMREX_ALWAYS_ASSERT(dr > std::numeric_limits<amrex::Real>::epsilon());
```

amr-wind/wind_energy/actuator/turbine/fast/turbine_fast_ops.H
```diff
@@ -243,7 +243,10 @@ struct InitDataOp<TurbineFast, SrcTrait>
         amrex::ParallelDescriptor::Comm_dup(
             amrex::ParallelDescriptor::Communicator(), tdata.tcomm);
 
-        int sz_info[4]{0, 0, 0, 0};
+        int sz_info[5]{0, 0, 0, 0, 0};
+        // Check if the size of sz_info is 5
+        assert(sizeof(sz_info) / sizeof(sz_info[0]) == 5);
```
mbkuhn, line 248:
```suggestion
        AMREX_ALWAYS_ASSERT(sz_info.size()) == 5);
```

amr-wind/wind_energy/actuator/turbine/fast/turbine_fast_ops.H
```diff
@@ -243,7 +243,10 @@ struct InitDataOp<TurbineFast, SrcTrait>
         amrex::ParallelDescriptor::Comm_dup(
             amrex::ParallelDescriptor::Communicator(), tdata.tcomm);
 
-        int sz_info[4]{0, 0, 0, 0};
+        int sz_info[5]{0, 0, 0, 0, 0};
```
mbkuhn, line 246:
```suggestion
        amrex::Array<int, 5> sz_info = {0, 0, 0, 0, 0};
```

amr-wind/wind_energy/actuator/turbine/fast/turbine_fast_ops.H
marchdf: TBH I would `git checkout main -- amr-wind/wind_energy/actuator/turbine/fast/turbine_fast_ops.H` having made sure to pull the latest main before. I don't think we should see any diffs in this file.

tonyinme: I agree with @marchdf , this file should not be changed by this PR. The reason this shows up as a conflict is because I started working on it with the fix I had on my local branch before @marchdf 's PR was merged. And because I needed this for an older version of OpenFAST.

mbkuhn: Marc's method is best, but I think if you get the merge conflict out of the way, you'll be able to update the branch through GitHub and that'll clear the diffs. I can go ahead and do that

amr-wind/wind_energy/actuator/FLLC.cpp
```diff
@@ -38,6 +43,64 @@ void FLLCInit(
...
+            dr1 = eps1 / data.eps_dr; // spacing needed to match condition
+
+            // Interpolate the value of epsilon to the next radial location
+            eps2 = interp::linear(data.r, data.optimal_epsilon, r_ + dr1);
+            dr2 = eps2 / data.eps_dr;
+
+            // This will ensure that the spacing will always meet the
+            // requirement eps/dr
+            dr = std::min(dr1, dr2);
+            AMREX_ALWAYS_ASSERT(dr > std::numeric_limits<amrex::Real>::epsilon());
```
mbkuhn, line 77: My change wasn't formatted!
```suggestion
            AMREX_ALWAYS_ASSERT(
                dr > std::numeric_limits<amrex::Real>::epsilon());
```

amr-wind/wind_energy/actuator/FLLC.cpp
```diff
@@ -38,6 +43,65 @@ void FLLCInit(
...
+            dr = std::min(dr1, dr2);
+            AMREX_ALWAYS_ASSERT(
+                dr > std::numeric_limits<amrex::Real>::epsilon());
+
+            // Append value to the array
+            // Ensure that the value is smaller than the tip
+            data.nonuniform_r.push_back(std::min(r_ + dr, data.r.back()));
+        }
+
+        int npts_n = data.nonuniform_r.size();
```
mbkuhn, line 85:
```suggestion
        int npts_n = static_cast<int>(data.nonuniform_r.size());
```

mbkuhn:
There's still a failing check because the FLLC unit tests don't pass in debug mode. I'm not sure exactly what the solution is, but the expectations may have changed (needing a modification to the unit test), or there needs to be an improvement to the FLLC code itself.

tonyinme:
> There's still a failing check because the FLLC unit tests don't pass in debug mode. I'm not sure exactly what the solution is, but the expectations may have changed (needing a modification to the unit test), or there needs to be an improvement to the FLLC code itself.

I see, I think the problem comes from the nonuniform flag being true. 

There are 2 options. 
1. Leaving the default of the `nonuniform` flag as false
2. Passing that flag to the unit-test as false

psakievich:
> I see, I think the problem comes from the nonuniform flag being true.
> 
> There are 2 options.
> 
> 1. Leaving the default of the `nonuniform` flag as false
> 2. Passing that flag to the unit-test as false

I don't think either of those are acceptable options. The failure is an index out of range error which means there is a memory leak with the non-uniform flag. 

```
[----------] 1 test from TestFLLCData
[ RUN      ] TestFLLCData.data_initializes_with_cviews
  what():  Assertion `i < (this->std::vector<T, Allocator>::size())' failed, file "/home/runner/work/amr-wind/amr-wind/submods/amrex/Src/Base/AMReX_Vector.H", line 35
 ```

The proper thing to do is to fix the unit test to make it work if you expect the nonuniform flag to be a production feature, or pass it as false then write a unit-test that works with the non-uniform flag.

You should probably do the second (or parameterize the test to test both conditions) b/c we want to know that both code paths work.

tonyinme:
The unit test has now been fixed and a new unit test was added for the nonuniform version. 
Thanks @psakievich for helping with the unit test.

mbkuhn (approved):

mbkuhn:
I think this makes sense because there is a new default behavior. However, before we rebless, we should probably confirm that the old pathway, which is still in the code, introduces no diffs. would you be able to test it with Actuator.FixedWingLine.fllc_nonuniform = false added to the input file?

tonyinme:
Ah, yes, this is expected. If we want to default to the original test value, we can set this [flag](https://exawind.github.io/amr-wind/user/inputs_Actuator.html#input_param-Actuator.FixedWingLine.fllc_nonuniform) to false in the test.

psakievich:
I think a rebless is warranted over changing the test. The default behavior is not really intended to be used in production research

marchdf:
Ok cool, yes tests pass with `Actuator.FixedWingLine.fllc_nonuniform = false` so I will rebless. Thanks!

PR #1047, opened 2024-05-09
overset refactoring (UpdateGradP)

mbkuhn:
## Summary

This is part of a bigger effort for bringing features for overset multiphase into main. Also, this is a clearer way of labeling things that AMR-Wind needs to do when coupled to the overset solver.
* remove references to overset from incflo_advance
* put them in incflo instead
* UpdateGradP is only used in overset simulations, define it in OversetOps class
* move generic projection functions to header file, remove from incflo class

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [x] on CPU 
- the regression tests
  - [ ] on GPU 
  - [x] on CPU 

## Additional background

Should be tested with regression tests


mbkuhn:
this should also be tested with the regression tests from exawind-driver

mbkuhn:
Confirmed that the regression tests (exawind-driver and amr-wind) both pass with the changes from this PR.

amr-wind/overset/OversetOps.cpp
```diff
@@ -0,0 +1,169 @@
...
+    amr_wind::Field const* mesh_detJ =
+        mesh_mapping ? &((*m_sim_ptr)
+                             .repo()
+                             .get_mesh_mapping_detJ(amr_wind::FieldLoc::CELL))
+                     : nullptr;
+    const auto* ref_density =
+        is_anelastic ? &((*m_sim_ptr).repo().get_field("reference_density"))
+                     : nullptr;
+
+    // ASA does not think we actually need to define sigma here. It
```
mbkuhn, line 52: this PR would be a good opportunity to check if the sigma calculation is needed and remove it

mbkuhn: confirmed it is unnecessary and removed it

amr-wind/incflo.cpp
```diff
@@ -310,6 +311,9 @@ void incflo::Evolve()
 
 void incflo::do_advance()
 {
+    if (m_sim.has_overset()) {
+        m_ovst_ops.pre_advance_actions();
```
marchdf, line 315: we have a bunch of places with `pre_advance_work`. Should this be renamed to that? And should it actually be moved to `pre_advance_stage2` where we do all the pre_advance stuff?

amr-wind/overset/OversetOps.H
```diff
@@ -0,0 +1,26 @@
...
+
+class OversetOps
+{
+public:
+    OversetOps() = default;
+
+    void initialize(CFDSim& sim);
+    void pre_advance_actions();
+
+    // Needs to be public for cuda
```
marchdf, line 17: remove this comment.

amr-wind/projection/incflo_apply_nodal_projection.cpp
```diff
@@ -4,33 +4,39 @@
...
+        abl.abl_mpl().set_velocity(lev, time, vel_fld, vel_mfab);
     }
 }
 
 Array<amrex::LinOpBCType, AMREX_SPACEDIM>
-incflo::get_projection_bc(Orientation::Side side) const noexcept
+amr_wind::nodal_projection::get_projection_bc(
+    Orientation::Side side,
+    amr_wind::Field& pressure,
+    Array<int, AMREX_SPACEDIM> is_periodic)
```
marchdf, line 34: pass by reference `const ... &`

marchdf (changes requested):

amr-wind/incflo.cpp
mbkuhn: Renaming it sounds good. The difficulty with putting it in either pre_advance_stage1 or pre_advance_stage2 is that these are actions that need to take place after the exchange_solution step is performed by exawind-driver. Both the pre advance stages are called prior to exchange_solution. If we wanted to, having another pre_advance_stage would be another way to organize it, but that seems like too much, especially because I don't think there's a nalu-wind counterpart for that.

marchdf: ah that's a good reason

marchdf (approved):

Issue #1008, opened 2024-03-28
Segfaults on initial MAC projection


gyalla:
I am encountering segfault issues with AMR-Wind on the initial MAC projection step, which aborts with the following message:

```
System                     Iters      Initial residual        Final residual
  ----------------------------------------------------------------------------

L-inf norm MAC vels: before MAC projection
..............................................................................
Max u:          10.71283853 |  Location (x,y,z):       3765,     1822.5,      312.5
Min u:         -10.03829912 |  Location (x,y,z):          0,          0,          0
Max v:          3.259859803 |  Location (x,y,z):          0,          0,          0
Min v:         -3.510260129 |  Location (x,y,z):     4332.5,        660,      552.5
Max w:          2.517395188 |  Location (x,y,z):    2253.75,     916.25,      102.5
Min w:         -1.063436419 |  Location (x,y,z):      912.5,     1112.5,         20
..............................................................................

  MAC_projection                 6           4.431624589       1.873963523e-06
[ec749:54683] *** Process received signal ***
[ec749:54683] Signal: Segmentation fault (11)
[ec749:54683] Signal code: Address not mapped (1)
[ec749:54683] Failing at address: 0x1d74b000
[ec749:54683] [ 0] /usr/lib64/libpthread.so.0(+0xf630)[0x2aaab2791630]
[ec749:54683] [ 1] amr_wind[0x505a9a]
[ec749:54683] [ 2] amr_wind[0x50631b]
[ec749:54683] [ 3] amr_wind[0x50fcf1]
[ec749:54683] [ 4] amr_wind[0x70a664]
[ec749:54683] [ 5] amr_wind[0x448f33]
[ec749:54683] [ 6] amr_wind[0x44ba9a]
[ec749:54683] [ 7] amr_wind[0x44e9c0]
[ec749:54683] [ 8] amr_wind[0x43e699]
[ec749:54683] [ 9] /usr/lib64/libc.so.6(__libc_start_main+0xf5)[0x2aaab29c0555]
[ec749:54683] [10] amr_wind[0x447cdb]
[ec749:54683] *** End of error message ***
```

The issue is node dependent, and I have compiled the results from multiple tests that myself and others (including @lawrenceccheung and @jfrederik-nrel) have run in the following table: 

```
Machine | Processor count | Result
------------------------------------
Frontier |   736 GPU      | Works
Frontier |   1792 CPU     | Failed
Eagle    |   512 CPU      | Failed
[...]

PR #1050, opened 2024-05-10
Diagnostics bug fix + de-hardcoding refRatio in makeFineMask calls

mbkuhn:
## Summary

Fixing the use of the mask array in PrintMaxMACVelLocations to avoid going out-of-bounds
Correcting places where the refinement ratio was hard-coded in makeFineMask calls

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [x] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [x] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU 

Issue Number: #941 #1008

### DIFF
--- a/amr-wind/physics/BurggrafFlow.cpp
+++ b/amr-wind/physics/BurggrafFlow.cpp
@@ -121,7 +121,7 @@ amrex::Real BurggrafFlow::compute_error(const Field& field)
         if (lev < nlevels - 1) {
             level_mask = makeFineMask(
                 m_mesh.boxArray(lev), m_mesh.DistributionMap(lev),
-                m_mesh.boxArray(lev + 1), amrex::IntVect(2), 1, 0);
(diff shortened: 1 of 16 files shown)


mbkuhn:
Other details: I added the unit test (commit 406c00c), but it still passed without the bug fix. However, in debug mode, the test doesn't pass because it can properly detect the out-of-bounds index. Is that good enough? Any suggestions?

With the full PR, the bug is fixed, and it does pass the test in debug.

mbkuhn:
This PR does change a lot of different files, but it's not well suited to being tested with regression tests, because the changes are purely in error measurements going to text or netCDF outputs, not in anything that would change the values of the plt files. However, I think the fact that the regression test run fine and the code compiles fine are good indications that those widespread changes are nothing to worry about. The one bug fix that does change the functionality is addressed with a unit test.

marchdf (approved):

PR #1053, opened 2024-05-10
Remove duplicates from IO

marchdf:
## Summary

If using derived fields it is possible to end up with a bunch of duplicate fields in the output. This attempts to fix that.

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

## Checklist

The existing `abl_sampling` regression test had duplicate fields, now it doesnt. So we can consider that the reg test.

- the unit tests
  - [ ] on GPU 
  - [ ] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU

### DIFF
--- a/amr-wind/ocean_waves/relaxation_zones/waves2amr_ops.H
+++ b/amr-wind/ocean_waves/relaxation_zones/waves2amr_ops.H
@@ -384,8 +384,7 @@ struct InitDataOp<W2AWaves>
 {
     void
     // cppcheck-suppress constParameterReference
-    operator()(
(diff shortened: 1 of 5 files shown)


marchdf:
@mbkuhn can you check that this solves the netcdf problem you were seeing?

mbkuhn (approved):
Can confirm that this eliminated the NetCDF name conflict issue that I had stumbled upon in the abl_sampling reg test. I also looked at a plt Header file and confirmed that there were no duplicates.

moprak-nrel (approved):

PR #1055, opened 2024-05-13
Readd waves2amr

marchdf:
## Summary

I was having all sorts of errors on a machine with an existing clone amr-wind. "fatal: not git repository", "unable to fetch Waves2AMR". This got me through all those errors. I am not sure what's up. Maybe the trailing slash? Another point... we should move this submodule repo out of a private GH account and into something official. We can move it Exawind org if need be. Or if it will only be used by amr-wind, I would prefer not having a submodule and just making it part of the source. 

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes


mbkuhn:
Didn't expect all of these issues! As its own repo, I thought it might be useful for ERF as well, for an evolving lower boundary. I will look into both of your suggestions. By the way, do the changes in this PR fix the errors you were getting?

.gitmodules
```diff
@@ -8,5 +8,7 @@
 	path = submods/AMReX-Hydro
 	url = https://github.com/AMReX-Fluids/AMReX-Hydro.git
 [submodule "submods/Waves2AMR"]
-	path = submods/Waves2AMR
-	url = https://github.com/mbkuhn/Waves2AMR/
\ No newline at end of file
+	path = submods/Wave2AMR
+	url = https://github.com/mbkuhn/Waves2AMR.git
+[submodule "submods/Wave2AMR"]
```
mbkuhn, line 13: I'm seeing a duplication here

mbkuhn (approved):

PR #1069, opened 2024-05-21
Replace mfiter loops with the MF parallel for versions

marchdf:
## Summary

Replace mfiter loops with the MF parallel for versions. These take care of a bunch of defaults for use, creates fused parallelfors for the GPU (nominally results in faster code), and generally simplifies the code.

I am looking for volunteers. I have annotated all the mfiter loops that can be simplified. I probably missed some and some might not be doable on a second glance. I did a couple to show how it is done and there are other examples in the code. 

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

## Checklist

- the unit tests
  - [ ] on GPU 
  - [ ] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU


amr-wind/core/Field.cpp
```diff
@@ -426,21 +421,18 @@ void Field::to_stretched_space() noexcept
...
-                [=] AMREX_GPU_DEVICE(int i, int j, int k, int n) noexcept {
-                    field(i, j, k, n) *= fac(i, j, k, n) / detJ(i, j, k);
-                });
-        }
+
+        const auto& fac = mesh_fac(lev).const_arrays();
+        const auto& detJ = mesh_detJ(lev).const_arrays();
+        const auto& field = operator()(lev).arrays();
+        amrex::ParallelFor(
+            mesh_fac(lev), amrex::IntVect(0), operator()(lev).nComp(),
```
mbkuhn, line 429: is the IntVect(0) to communicate the ghost cells? shouldn't it be nonzero because the previous code has a grown box?

marchdf: Yes and yes it should using field->ngrow. Good catch

marchdf:
We took care of most of these.

PR #1075, opened 2024-05-24
Move channel flow wall model assert on body force to a better place

moprak-nrel:
## Summary

We require that the normal direction is 2 and that the body force in the 2 direction is 0 when using a wall model. However the constructor for the wall model gets initialized even when a wall model for the channel flow even when not using one. This PR moves the assert to the right place.

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [x] on CPU 
- the regression tests
  - [ ] on GPU 
  - [x] on CPU 

## Additional background
This fixes the reg test `channel_mol_mesh_map_z` from failing.

Issue Number:


marchdf (approved):

marchdf:
Thank you!

PR #1078, opened 2024-05-28
Remove ChannelFlow.density and replace with just incflo.density

moprak-nrel:
## Summary

`ChannelFlow` seems to define it's own density, and it is unclear why this needs to be a separate parameter from `incflo.density`. This PR consolidates those.

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Other (please describe):
Consolidating density into one parameter for `ChannelFlow`
## Checklist

The following is included:

- [ ] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [x] on CPU 
- the regression tests
  - [ ] on GPU 
  - [x] on CPU 

## Additional background

Issue Number:


moprak-nrel:
This PR also fixes a potential bug with the wall function, previously this assumed that density is 1.0, and set `nu = mu` instead of `nu = mu/rho`.

PR #1079, opened 2024-05-28
Conform multiphase CFL to expectations

mbkuhn:
## Summary

Multiplies advective CFL by 2 near interface so that different constraints, in different parts of the domain, can apply together.

## Pull request type

- [ ] Bugfix
- [x] Feature
- [x] Code style update (formatting, renaming)
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [x] on CPU 
- the regression tests
  - [ ] on GPU 
  - [x] on CPU 

## Additional background

The volume-of-fluid scheme implemented in AMR-Wind requires a CFL of < 0.5 to ensure conservation of each phase. Therefore, it has been the recommendation to run multiphase simulations at CFL < 0.5. However, for simulations where there are dynamics both near and far from the interface, such as an ABL simulation with waves, this may be unnecessarily restrictive, e.g., when the flow in the air is at a higher CFL compared to the flow near the waves. The changes in this PR multiply the advective CFL calculated near the interface by 2, so it can be compared with 1 instead of with 0.5. This will prevent dt from being overly restricted by the CFL in a multiphase ABL case.

### DIFF
--- a/amr-wind/incflo_compute_dt.cpp
+++ b/amr-wind/incflo_compute_dt.cpp
@@ -112,6 +112,11 @@ void incflo::ComputeDt(bool explicit_diffusion)
                         result = std::abs(v_bx(i, j, k, 0)) * dxinv[0] / fac_x +
                                  std::abs(v_bx(i, j, k, 1)) * dxinv[1] / fac_y +
                                  std::abs(v_bx(i, j, k, 2)) * dxinv[2] / fac_z;
+
+                        // Multiply advective CFL by 2 when near interface
+                        result *= 2.0
+                        // CFL requirement to ensure vof conservation is 0.5;
+                        // this is half the typical concept of a CFL (1.0)
                     }
                     return result;
                 });
--- a/amr-wind/incflo_compute_dt.cpp
+++ b/amr-wind/incflo_compute_dt.cpp
@@ -114,7 +114,7 @@ void incflo::ComputeDt(bool explicit_diffusion)
                                  std::abs(v_bx(i, j, k, 2)) * dxinv[2] / fac_z;
 
                         // Multiply advective CFL by 2 when near interface
-                        result *= 2.0
+                        result *= 2.0;
                         // CFL requirement to ensure vof conservation is 0.5;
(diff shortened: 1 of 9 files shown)


mbkuhn:
The reg tests have been updated to reflect this new usage of CFL for multiphase cases. These changes will only affect the reg test results in multiphase cases with a variable dt. From among those run by ctest, this is only the inertial_drop case, and I have confirmed that this is the only one that changes.

amr-wind/incflo_compute_dt.cpp
```diff
@@ -112,6 +112,11 @@ void incflo::ComputeDt(bool explicit_diffusion)
                         result = std::abs(v_bx(i, j, k, 0)) * dxinv[0] / fac_x +
                                  std::abs(v_bx(i, j, k, 1)) * dxinv[1] / fac_y +
                                  std::abs(v_bx(i, j, k, 2)) * dxinv[2] / fac_z;
+
+                        // Multiply advective CFL by 2 when near interface
+                        result *= 2.0;
```
gdeskos, line 117: Hmm is this 0.5, to do with the mass-momentum consistency criterion?

mbkuhn: pretty much -- the VOF scheme needs CFL < 0.5 to ensure conservation of mass. By multiplying this (the calculated advective CFL) by 2, that criterion becomes CFL < 1.0.

Essentially, with this change, the user can enforce CFL < 0.5 near the interface and enforce CFL < 1 elsewhere by setting the max CFL to 1 in the input file.

gdeskos: okay, make sure you remove any restrictions for CFL=0.5 from other parts of the code.

mbkuhn: That's what I did with the reg tests. This change doesn't apply to the CFL calculation when the velocities are prescribed, so I didn't change the reg tests with prescribed velocities.

gdeskos (approved):

Issue #1076, opened 2024-05-24
MAC velocity not getting filled correctly at the low boundary


mukul1992:
## Bug description
I am using `u_mac` (MAC velocity in x-direction) as an example. The index `i=0` lies at the boundary for u_mac, but does not get filled by `fillpatch_sibling_fields` called from `preadvect` in `icns_advection.H`. 

## Steps to reproduce

1. Compiler used
   - [x] GCC
   - [ ] LLVM
   - [ ] oneapi (Intel)
   - [ ] nvcc (NVIDIA)
   - [ ] rocm (AMD)
   - [ ] with MPI
   - [ ] other:
2. Operating system
   - [ ] Linux
   - [x] OSX
   - [ ] Windows
   - [ ] other (do tell ;)):
3. Hardware:
   - [x] CPU
   - [ ] GPU
4. Machine details: Apple M3 laptop
5. Input file: `rankine.inp`

## AMR-Wind information
```
==============================================================================
                AMR-Wind (https://github.com/exawind/amr-wind)

  AMR-Wind version :: d6d42205
  AMR-Wind Git SHA :: d6d422058c48cf378117e1ab61c3c5d499ac5a2a
  AMReX version    :: 24.05-20-g5d02c6480a0d

  Exec. time       :: Fri May 24 12:09:29 2024
  Build time       :: May 24 2024 09:19:32
  C++ compiler     :: GNU 13.2.0

  MPI              :: ON    (Num. ranks = 1)
  GPU              :: OFF
  OpenMP           :: OFF

  No additional third-party libraries enabled

           This software is released under the BSD 3-clause license.
 See https://github.com/Exawind/amr-wind/blob/development/LICENSE for details.
------------------------------------------------------------------------------

```

## Additional context
Apart from the fact that technically `i=0` lies at the boundary and hence should be filled, the main issue arises for me while implementing the inflow-outflow BC which can have cells with Dirichlet treatment on both the low and high boundaries. The MAC velocity at the high boundary (e.g. `u_mac` at the `xhi` boundary) does get filled in the above function and this creates an unexpected asymmetry. For example, if there are 16 cells in the x-direction, `i=0` and `i=16` would lie on the low and high boundaries respectively. `i=0` does not get filled, but `i=16` does get filled, hence breaking the symmetry. 

Also, I think this issue would only surface for a **time-varying boundary condition**.

[...]

PR #1084, opened 2024-05-30
One way of fixing UDF fillpatches for face quantities (mac velocities)

mbkuhn:
## Summary

Using template arguments to modify the boundary index checks for face quantities. Ended up needing a decent amount of new function / type declarations, though, and I'm not sure how best to use the loop within fillpatch_sibling_fields now that each physbc has a different type.

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [ ] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU 

## Additional background

This may not be the best way to do it... it may make things harder when we try to improve the UDFs. Obviously, this is starting off sloppy and will need a thorough pass through and probably unit tests added as well.

Issue Number: #1076


mbkuhn:
Running the rankine test with this branch give the following output (relevant parts only)
`
// Fill mac velocities using velocity BCs
sibling fields FPSL dir i = 0 (umac)
Rankine.H op: filling 10.73181747 at i=-1
Rankine.H op: filling 10.13957005 at i=41
sibling fields FPSL dir i = 1 (vmac)
Rankine.H op: filling 1.016180833 at i=-1
Rankine.H op: filling 0.5357916479 at i=40
sibling fields FPSL dir i = 2 (wmac)
Rankine.H op: filling 0 at i=-1
Rankine.H op: filling 0 at i=40
`

amr-wind/core/FieldBCOps.H
```diff
@@ -11,6 +11,17 @@
 
 namespace amr_wind {
 
+namespace {
+int CC_idx(int /* idir */)
```
marchdf, line 15: I wonder if for some of these we can use stuff in: https://github.com/Exawind/amr-wind/blob/main/amr-wind/utilities/DirectionSelector.H#L56. Kinda like we do for `IndexSelector`

amr-wind/core/FieldBCOps.H
```diff
@@ -50,6 +64,9 @@ struct FieldBCNoOp
     constexpr explicit FieldBCNoOp(const Field& /*unused*/) {}
 
     FunctorType operator()() const { return *this; }
+    FunctorXType opfx() const { return *this; }
```
marchdf, line 67: Could we template this so we don't have repeats? `template <T> T opf()` ?

amr-wind/core/FieldBCOps.H
```diff
@@ -169,9 +186,10 @@ struct DirichletOp
                 }
 
                 // Check if the point in question is on the boundary
+                int mod_bigEnd = domain_box.bigEnd(idir) + IdxOp(idir);
```
marchdf, line 189: `const` and maybe a more descriptive name? `bigEnd_depending_on_if_you_are_cc_or_fc` ;-) obv not that either.

amr-wind/core/FieldBCOps.H
```diff
@@ -203,11 +221,17 @@ struct DirichletOp
...
-    using FunctorType = DirichletOp<ConstDirichlet, ConstDirichlet>;
+    using FunctorType = DirichletOp<ConstDirichlet, ConstDirichlet, &CC_idx>;
+    using FunctorXType = DirichletOp<ConstDirichlet, ConstDirichlet, &FX_idx>;
+    using FunctorYType = DirichletOp<ConstDirichlet, ConstDirichlet, &FY_idx>;
+    using FunctorZType = DirichletOp<ConstDirichlet, ConstDirichlet, &FZ_idx>;
 
     explicit FieldBCDirichlet(const Field& fld) : m_field(fld) {}
 
     inline FunctorType operator()() const { return FunctorType(m_field); }
+    inline FunctorXType opfx() const { return FunctorXType(m_field); }
```
marchdf, line 232: I gotta think there's a way to template these... `template <typename T> T opf() const {return T(m_field);}`. Can you see if that's possible?

amr-wind/core/FieldBCOps.H
```diff
@@ -217,7 +241,10 @@ struct BCOpCreator
 {
     using InflowOpType = typename InflowOp::DeviceType;
     using WallOpType = typename WallOp::DeviceType;
-    using FunctorType = DirichletOp<InflowOpType, WallOpType>;
+    using FunctorType = DirichletOp<InflowOpType, WallOpType, &CC_idx>;
```
marchdf, line 244: I think we don't need `CC_idx` I think we can leave it empty and use a default.

amr-wind/core/FieldBCOps.H
```diff
@@ -235,6 +262,27 @@ struct BCOpCreator
             m_wall_op.device_instance()};
     }
 
+    inline FunctorXType opfx() const
```
marchdf, line 265: with the templating I think we consolidate all these.

amr-wind/core/FieldFillPatchOps.H
```diff
@@ -270,13 +273,34 @@ public:
     {
 
         if (lev == 0) {
-            amrex::PhysBCFunct<amrex::GpuBndryFuncFab<Functor>> physbc(
-                m_mesh.Geom(lev), bcrec, bc_functor());
+            amrex::PhysBCFunct<amrex::GpuBndryFuncFab<FunctorX>> physbcx(
+                m_mesh.Geom(lev), bcrec, bc_functor_x());
```
marchdf, line 277: What would be neat here is to keep the for loop and instead do `bc_functor_f(i)`. Or maybe it becomes easier with the templating.

marchdf (changes requested):

marchdf:
Nice work! I left a bunch of comments/thoughts/things to try out. LMK if you want help with these.

I think this mostly takes care of the first bit: getting the right indices to be used in the UDF `op`. The comments I had in the original issue were more focused on defining a different `op` for the faces. I am not sure we want to go down that road but it's the "right thing to do" eventually. Though again, getting those precise details right is probably only important for BDS and the mac vel in the ghost cells.

amr-wind/core/FieldBCOps.H
mbkuhn: took some inspiration from there.

amr-wind/core/FieldBCOps.H
mbkuhn: I didn't include FunctorType in the templating but that is definitely possible too.

amr-wind/core/FieldFillPatchOps.H
mbkuhn: I'm having a hard time getting the loop to work. The problem is that the type of physbcx depends on the direction, in addition to the bc_functor type. I have templatized the FunctorF type, but the template argument has to be a const expression, so it cannot be according to the index i. Making bc_functor output different things depending on an input argument is a problem, too, because each output would be a different type. I also can't put the physbcs into a vector because they're different types. Not sure if there is a way forward here.

marchdf:
This is done in #1093

PR #1092, opened 2024-06-04
Another way to fix the UDF MAC fills

marchdf:
## Summary

The dirichlet fill now depends on if we are filling a face-centered or node centered field.

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

## Checklist

- the unit tests
  - [ ] on GPU 
  - [ ] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU 

Issue Number: #1076


marchdf:
Still need to do it for lev != 0

marchdf:
This PR has made a lot of people very angry and been widely regarded as a bad move.

(the real reason for closing: templating wont work because for lev != 0 we need to pass in an `Array` of bc functs and so they need to have the same type. It was fun while it lasted. #1093 should be a better fix)

PR #1093, opened 2024-06-06
Fix ghost cells being used for mac velocity BC functions

marchdf:
## Summary

Fixing the fact that the UDFs are called on cell-centered and face centered velocities and, therefore, the indices of the cells should be set appropriately.

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

## Checklist

Left to do:
- [x] I need to do the same for `lev != 0`

The following is included:

- [ ] new unit-test(s)
- [x] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [x] on CPU 
- the regression tests
  - [ ] on GPU 
  - [x] on CPU 

## Additional background

With these changes, the fillpatch of the mac velocities now touches the following indices depending on the face-centered mac velocity component. For umac, we need to be touching -1 and 41 (with n_cell = 40). For the others it is -1 and 40.

```
  MAC_projection                 9         0.01766980429       1.787159811e-14
this face_idir: 0
Rankine.H op: filling 10.73181747 at i=-1
Rankine.H op: filling 10.13957005 at i=41
this face_idir: 1
Rankine.H op: filling 1.016180833 at i=-1
Rankine.H op: filling 0.5357916479 at i=40
this face_idir: 2
Rankine.H op: filling 0 at i=-1
Rankine.H op: filling 0 at i=40
```

Issue Number: #1076

### DIFF
--- a/amr-wind/core/FieldBCOps.H
+++ b/amr-wind/core/FieldBCOps.H
@@ -49,7 +49,7 @@ struct FieldBCNoOp
     AMREX_GPU_HOST
     constexpr explicit FieldBCNoOp(const Field& /*unused*/) {}
 
-    FunctorType operator()(const int face_dir = -1) const { return *this; }
+    FunctorType operator()(const int /*face_dir*/ = -1) const { return *this; }
 
     AMREX_GPU_DEVICE
     void operator()(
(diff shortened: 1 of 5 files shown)


mbkuhn:
🏆 for this PR name!

marchdf:
If it turns out that stuff in IAMR/PeleLMeX is needed I can add it as a separate PR. Right now it is looking like it is not needed (at least for most cases, maybe needed for variable density?).  Discussion is ongoing but shouldn't hold this PR up. I don't expect any diffs with this PR.

mbkuhn (approved):

PR #1124, opened 2024-07-09
Pass correct half-time values for MAC vel bndry fills

mukul1992:
## Summary

Passes half-time values to the UDFs for filling the MAC velocity boundaries, which is the correct behavior. Before this change, the initial iterations passed zero and the advancing iterations passed the "new time" values instead of the half-time values.

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [ ] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU

### DIFF
--- a/amr-wind/equation_systems/PDE.H
+++ b/amr-wind/equation_systems/PDE.H
@@ -137,7 +137,7 @@ public:
     {
         BL_PROFILE(
             "amr-wind::" + this->identifier() + "::pre_advection_actions");
-        m_adv_op->preadvect(fstate, m_time.deltaT(), m_time.new_time());
+        m_adv_op->preadvect(fstate, m_time.deltaT(), m_time.current_time());
     }
 
     void compute_predictor_rhs(const DiffusionType difftype) override
--- a/amr-wind/equation_systems/icns/icns_advection.H
+++ b/amr-wind/equation_systems/icns/icns_advection.H
@@ -305,7 +305,8 @@ struct AdvectionOp<ICNS, fvm::Godunov>
         if (fvm::Godunov::nghost_state > 0) {
             amrex::Array<Field*, AMREX_SPACEDIM> mac_vel = {
                 AMREX_D_DECL(&u_mac, &v_mac, &w_mac)};
-            dof_field.fillpatch_sibling_fields(time + 0.5*dt, u_mac.num_grow(), mac_vel);
+            dof_field.fillpatch_sibling_fields(
+                time + 0.5*dt, u_mac.num_grow(), mac_vel);
         }
 
         for (int lev = 0; lev < repo.num_active_levels(); ++lev) {
--- a/amr-wind/equation_systems/icns/icns_advection.H
+++ b/amr-wind/equation_systems/icns/icns_advection.H
@@ -305,7 +305,7 @@ struct AdvectionOp<ICNS, fvm::Godunov>
         if (fvm::Godunov::nghost_state > 0) {
             amrex::Array<Field*, AMREX_SPACEDIM> mac_vel = {
                 AMREX_D_DECL(&u_mac, &v_mac, &w_mac)};
-            dof_field.fillpatch_sibling_fields(time, u_mac.num_grow(), mac_vel);
+            dof_field.fillpatch_sibling_fields(time + 0.5*dt, u_mac.num_grow(), mac_vel);
         }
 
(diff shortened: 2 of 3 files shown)


marchdf:
I am getting for following tests failing on CPU:
```
The following tests FAILED:
        108 - abl_bndry_input_native (Failed)
        110 - abl_bndry_input_amr_native (Failed)
        111 - abl_bndry_input_amr_native_mlbc (Failed)
        112 - abl_godunov_forcetimetable (Failed)
        113 - abl_bndry_input (Failed)
        114 - abl_bndry_input_amr (Failed)
        115 - abl_bndry_input_amr_inflow (Failed)
```
They are failing on:
```
terminate called after throwing an instance of 'terminate called after throwing an instance of 'terminate called after throwing an instance of 'std::runtime_error'
  what():  Assertion `std::abs(time - m_in_data.tinterp()) < 1e-12' failed, file "/mnt/vdb/home/mhenryde/exawind/exawind-manager/environments/amr-wind-cpu-test/amr-wind/amr-wind/wind_energy/ABLBoundaryPlane.cpp", line 881
std::runtime_error'
  what():  Assertion `std::abs(time - m_in_data.tinterp()) < 1e-12' failed, file "/mnt/vdb/home/mhenryde/exawind/exawind-manager/environments/amr-wind-cpu-test/amr-wind/amr-wind/wind_energy/ABLBoundaryPlane.cpp", line 881
SIGABRT
SIGABRT
std::runtime_error'
terminate called after throwing an instance of '  what():  Assertion `std::abs(time - m_in_data.tinterp()) < 1e-12' failed, file "/mnt/vdb/home/mhenryde/exawind/exawind-manager/environments/amr-wind-cpu-test/amr-wind/amr-wind/wind_energy/ABLBoundaryPlane.cpp", line 881
SIGABRT
std::runtime_error'
  what():  Assertion `std::abs(time - m_in_data.tinterp()) < 1e-12' failed, file "/mnt/vdb/home/mhenryde/exawind/exawind-manager/environments/amr-wind-cpu-test/amr-wind/amr-wind/wind_energy/ABLBoundaryPlane.cpp", line 881
SIGABRT
See Backtrace.2 file for details
See Backtrace.1 file for details
See Backtrace.3 file for details
See Backtrace.0 file for details
Abort(6) on node 2 (rank 2 in comm 496): application called MPI_Abort(comm=0x84000001, 6) - process 2
Abort(6) on node 1 (rank 1 in comm 496): application called MPI_Abort(comm=0x84000001, 6) - process 1
Abort(6) on node 0 (rank 0 in comm 496): application called MPI_Abort(comm=0x84000002, 6) - process 0
Abort(6) on node 3 (rank 3 in comm 496): application called MPI_Abort(comm=0x84000001, 6) - process 3
<end of output>
Test time =   4.10 sec
```

which makes sense... kinda. Need to think about the right fix...

marchdf:
Basically when we use boundary data to fill velocities from precursor data, we have velocity data interpolated at a given time (not the half time). Right now we just use that for the mac vels. Moving that to a half time would mean reinterpolating that data in time. Which isn't cheap... We could just ignore the half-time right here: https://github.com/Exawind/amr-wind/blob/main/amr-wind/wind_energy/ABLFillInflow.cpp#L91 and keep new time to preserve the existing behavior?

mukul1992:
I added the fix that @marchdf suggested above. It does pass all the tests.

marchdf:
Thanks! The existential question is whether we think this is the "right" fix (just talking about the boundary plane thing):

Pros: 
- No need for reinterpolation (possibly rereading the file, dealing with final times, etc)
- Ghost cell mac vels arent used other than BDS. So this probably doesn't matter other than in that scheme (which no one uses)
- Simple
- Preserves existing behavior

Cons:
- Not strictly correct.
- Probably introduces some error that breaks any order of accuracy claims. Though this kind of claim is broken in general for this kind of case because we are just doing simple interpolation in time anyway, we aren't doing anything for local outflow possibilities, etc. 

Anyone else has thoughts? @mbkuhn, @moprak-nrel ?  Right now I am leaning towards the simple/easy thing.

marchdf (approved):

PR #1125, opened 2024-07-09
Skip boxes that don't have actuators

marchdf:
## Summary

Skip boxes where the actuator won't have an influence by using the actuator's `bound_box`.

For a large case (thanks @hgopalan!) with O(3k) actuator disks: 

Original profile:
```
Time spent in InitData():    5595.328444
Time spent in Evolve():      4523.809873


TinyProfiler total time across processes [min...avg...max]: 1.012e+04 ... 1.012e+04 ... 1.012e+04

------------------------------------------------------------------------------------------------------------------
Name                                                               NCalls  Excl. Min  Excl. Avg  Excl. Max   Max %
------------------------------------------------------------------------------------------------------------------
amr-wind::ActSrcOp<UniformCtDisk>                                  328603      200.2      596.4       4695  46.39%
FillBoundary_finish()                                               49052      23.43       3851       4479  44.27%
amr-wind::actuator::Actuator::post_init_actions                         2       2678       3185       3902  38.56%
amr-wind::RefineCriteriaManager::initialize                             1      635.8        943       1372  13.56%
amr-wind::incflo::ReadCheckpointFile()                                  1    0.02672      345.9       1128  11.15%
FabArray::ParallelCopy_finish()                                      8814      12.83      192.3      708.8   7.00%
DistributionMapping::LeastUsedCPUs()                                   16    0.02703      557.4      643.2   6.36%
MLLinOp::makeSubCommunicator()                                         13     0.1679      84.76        108   1.07%
amr-wind::incflo::ErrorEst()                                            3      86.24      97.05      105.3   1.04%
FPinfo::FPinfo()                                                        9    0.04631      18.73      74.71   0.74%
MLNodeLaplacian::Fsmooth()                                           5160      41.84      45.45      46.89   0.46%
FillBoundary_nowait()                                               49052      16.56      19.55      25.42   0.25%
```

With this PR:
```
Time spent in InitData():    5369.242606
Time spent in Evolve():      165.6200254


TinyProfiler total time across processes [min...avg...max]: 5535 ... 5535 ... 5535

---------------------------------------------------------------------------------------------------------------------
Name                                                                  NCalls  Excl. Min  Excl. Avg  Excl. Max   Max %
---------------------------------------------------------------------------------------------------------------------
amr-wind::actuator::Actuator::pre_init_actions                             1       2664       3034       3979  71.88%
amr-wind::incflo::ReadCheckpointFile()                                     1    0.02621      313.5       1834  33.14%
DistributionMapping::LeastUsedCPUs()                                      16    0.01368       1084       1392  25.14%
amr-wind::RefineCriteriaManager::initialize                                1      663.1      873.9       1378  24.91%
FillBoundary_finish()                                                  52854       16.6      83.13      89.02   1.61%
amr-wind::ActSrcOp<UniformCtDisk>                                      54477     0.2865      2.766      65.97   1.19%
amr-wind::incflo::ErrorEst()                                               3      30.58      36.08       40.2   0.73%
MLNodeLaplacian::Fsmooth()                                              5128      14.54      16.97      17.85   0.32%
```

This is a *27X speedup* of the time evolution step. The source calculation `amr-wind::ActSrcOp<UniformCtDisk>` went from 46% of the runtime to 1% of the runtime, or a *220X speedup*. 

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Other (please describe): speedup code

## Checklist

- the unit tests
  - [ ] on GPU 
  - [x] on CPU 
- the regression tests
  - [ ] on GPU 
  - [x] on CPU

### DIFF
--- a/amr-wind/wind_energy/actuator/ActSrcLineOp.H
+++ b/amr-wind/wind_energy/actuator/ActSrcLineOp.H
@@ -91,6 +91,13 @@ void ActSrcOp<ActTrait, ActSrcLine>::operator()(
     BL_PROFILE("amr-wind::ActSrcOp<" + fname + ">");
 
     const auto& bx = mfi.tilebox();
+
+    const auto bxa = utils::realbox_to_box(actObj.m_data.info().bound_box, geom);
+    const auto& bxi = bx & bxa;
+    if (bxi.isEmpty()) {
+      return;
+    }
+    
     const auto& sarr = m_act_src(lev).array(mfi);
     const auto& problo = geom.ProbLoArray();
     const auto& dx = geom.CellSizeArray();
--- a/amr-wind/wind_energy/actuator/ActSrcLineOp.H
+++ b/amr-wind/wind_energy/actuator/ActSrcLineOp.H
@@ -95,9 +95,9 @@ void ActSrcOp<ActTrait, ActSrcLine>::operator()(
     const auto bxa = utils::realbox_to_box(m_data.info().bound_box, geom);
     const auto& bxi = bx & bxa;
     if (bxi.isEmpty()) {
-      return;
+        return;
     }
-    
+
     const auto& sarr = m_act_src(lev).array(mfi);
     const auto& problo = geom.ProbLoArray();
     const auto& dx = geom.CellSizeArray();
--- a/amr-wind/wind_energy/actuator/ActSrcLineOp.H
+++ b/amr-wind/wind_energy/actuator/ActSrcLineOp.H
@@ -92,7 +92,7 @@ void ActSrcOp<ActTrait, ActSrcLine>::operator()(
 
     const auto& bx = mfi.tilebox();
 
-    const auto bxa = utils::realbox_to_box(actObj.m_data.info().bound_box, geom);
+    const auto bxa = utils::realbox_to_box(m_data.info().bound_box, geom);
     const auto& bxi = bx & bxa;
     if (bxi.isEmpty()) {
       return;
--- a/amr-wind/wind_energy/actuator/Actuator.cpp
+++ b/amr-wind/wind_energy/actuator/Actuator.cpp
@@ -56,28 +56,19 @@ void Actuator::post_init_actions()
     BL_PROFILE("amr-wind::actuator::Actuator::post_init_actions");
 
     amrex::Vector<int> act_proc_count(amrex::ParallelDescriptor::NProcs(), 0);
-    {
-        BL_PROFILE(
-            "amr-wind::actuator::Actuator::post_init_actions::determine_root_"
-            "proc");
-        for (auto& act : m_actuators) {
-            act->determine_root_proc(act_proc_count);
-        }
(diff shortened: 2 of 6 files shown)


marchdf:
On a case with o(1000) turbines I am seeing:
`amr-wind::ActSrcOp<UniformCtDisk>                                       1342     0.6288      1.347      30.77   5.77%`
go down to this:
`amr-wind::ActSrcOp<UniformCtDisk>                                       1342    0.00011    0.02295      1.048   0.20%`
with this PR. 

If this works out on other cases, I will clean up the PR and push the same kind of fix to the other disks/spreading functions.

mbkuhn:
do the changes in this PR also apply to post_regrid_actions and not just post init? I think they do, but I just wanted to check. You may want to add the profiling calls to that section too, while you're at it.

marchdf:
Actually didn't mean to have those profiling calls still in there. Let me remove those. This is just for the src term calculation: skip all the boxes where the actuator won't have an influence on the src term.

mbkuhn (approved):

marchdf:
The questions I have for those more familiar with this code:

- I don't get diffs from the reg tests and some simple 2 turbine cases I ran. But can we be sure that checking the intersection of the `bound_box` with the mfiter box is enough?
- I would like to do the same kind of check for this one: https://github.com/Exawind/amr-wind/blob/main/amr-wind/wind_energy/actuator/turbine/ActSrcDiskOp_Turbine.H#L86. But it looks like it has a tower and hub. Does the `bound_box` include these? I couldn't figure out a sample case that uses this and where the bound box is computed for that. Does anyone know? @psakievich looks like you are on this blame but please redirect me if there is someone else who might know.

mbkuhn:
I know from experience with 2D and moving fixed wing cases that if the expected force distribution exceeds the bounding box, it won't be there -- it won't get calculated on the mesh. So, I would think that the bound_box would have to cover everywhere that force is expected to be calculated.

Looking at where bound_box is calculated for different things:
- for standalone actuator disks (joukowsky, uniform_ct), there appears to be no tower representation in the reg tests, and when the bounding box is calculated in wind_energy/actuator/disk/disk_ops.cpp, only the disk itself is considered.
- for turbines (OpenFAST coupled, both disk and lines), the bounding box includes the tower (and then some) when it is calculated in wind_energy/actuator/turbine/turbine_utils.cpp. In z, the lower bound is the base position (already at the ground) - 1.25 * the radius. Probably should change that to just the base position, but it shouldn't be hurting anything the way it is.

marchdf:
Ok cool, thanks for double checking that. And yes, the bound_box is originally used to check the procs that are influenced by a given actuator. So I would expect nothing to happen if the force extent is larger than the bound box extend. This PR extends that to to bx's used for the source term. So I think we are all good here and I will make sure this also happens for turbines.

psakievich (approved):

rybchuk:
Huge!! Thanks Marc!

PR #1128, opened 2024-07-11
Modify apply_bcs (called for diffusion terms) to use correct time

mbkuhn:
## Summary

apply_bcs(), which is called when calculating explicit diffusion terms or prepping an implicit diffusion solve, previously only passed the new time value as an argument, regardless of the moment within the algorithm it was being called. Now, based on the state passed into apply_bcs, the routine chooses either the new time or the current time. This allows the correct time to be passed through, which is important for time-varying BCs.

To implement this change, the correct state needs to be passed to the diff_term calculation in incflo_advance(). This change prevents the current, hacky way of using the state to choose when to multiply or divide icns terms by zero (impacts explicit diffusion terms and source terms in icns because the momentum equation is being solved, but for the MAC velocity, things need to be phrased in terms of velocity instead). This PR moves the different treatments of density from the source and diffusion term calculations and puts it into incflo_advance().

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [x] on CPU 
- the regression tests
  - [ ] on GPU 
  - [x] on CPU 

## Additional background

Part of the effort to properly implement inflow-outflow BCs.


mbkuhn:
This is a good PR for relying on the reg tests. Here is what I've found.

There are a handful of reg tests that fail, but by a very small margin. These include tests that have nonuniform or non-unity densities. I was able to confirm that these diffs are only caused by the different order/arrangement of floating point multiplication and division. For the viscous term, the numbers are slightly different because it is divided by rho instead of multiplied by 1/rho. For the source term, the difference is because instead of skipping multiplication prior to preadvect, the terms are multiplied by rho and then divided by rho. No meaningful differences there.

There are a few cases with more noticeable diffs. rayleigh_taylor_mol has the largest diffs, but I'm not sure if understanding those should take any of our time. The godunov version falls under the previous category, with very small diffs. ow_linear has larger diffs, but only for pressure, and this didn't show up when I did a visual comparison of the pressure data. I'm not really worried about that one. The last one is rankine, which is actually a time-dependent UDF case, so it should have diffs with this change.

Lastly, the cases with boundary plane inputs all failed due to time correspondence problems. However, I am hoping that once I incorporate the latest merged PR #1124, these will be addressed.

mbkuhn:
I had to add a change to the treatment of the time passed to the boundary plane routine in fillphysbc, which parallels the change made for the other PR. Now the reg tests with boundary planes are running fine. @marchdf, what do you think?

@mukul1992, could you check this with your print statements?

marchdf:
Good work! I think it looks fine. I am going to run the tests on my end as well.

amr-wind/core/field_ops.H
```diff
@@ -67,6 +67,71 @@ add(T1& dst, const T2& src, int srccomp, int dstcomp, int numcomp, int nghost)
...
+ *  \param [in] dstcomp Starting component index of destination field
+ *  \param [in] numcomp Number of components to be updated
+ *  \param [in] nghost Number of ghost cells to be updated
+ *  \param [in] iter_src_comp Flag of whether src components should be iterated
+ */
+template <typename T1, typename T2>
+inline void divide(
+    T1& dst,
+    const T2& src,
+    int srccomp,
```
marchdf, line 87: const all these.

amr-wind/core/field_ops.H
```diff
@@ -67,6 +67,71 @@ add(T1& dst, const T2& src, int srccomp, int dstcomp, int numcomp, int nghost)
...
+    int dstcomp,
+    int numcomp,
+    const amrex::IntVect& nghost,
+    const bool iter_src_comp)
+{
+    const int nlevels = dst.repo().num_active_levels();
+    for (int lev = 0; lev < nlevels; ++lev) {
+        if (iter_src_comp) {
```
marchdf, line 95: I think (?) I would prefer 2 different versions of `divide` instead of a bool that decides what version is being used... I think that might be clearer.

amr-wind/equation_systems/BCOps.H
```diff
@@ -39,7 +39,11 @@ struct BCOp<PDE, std::enable_if_t<std::is_base_of_v<ScalarTransport, PDE>>>
     {
         amrex::IntVect ng_diff(1);
         auto& field = m_fields.field;
-        field.fillphysbc(m_time.new_time(), ng_diff);
+        // Currently, this is only called from old or new states
```
marchdf, line 42: instead of a comment, maybe an abort if it is anything other than old or new? Maybe that's too tricky...

marchdf:
I am getting the following:
```
         25 - abl_stable (Failed)                 -> machine diffs
         31 - abl_multiphase_laminar (Failed).    -> machine diffs
         49 - rayleigh_taylor_godunov (Failed).   -> machine diffs
         50 - rayleigh_taylor_mol (Failed)        -> LARGE diffs
         55 - dam_break_godunov (Failed)          -> machine diffs
         68 - rain_drop (Failed)                  -> machine diffs
         69 - inertial_drop (Failed)              -> machine diffs
         70 - ow_linear (Failed)                  -> machine diffs EXCEPT p  9.81           0.0009978830054
         71 - ow_stokes (Failed)                  -> machine diffs
         79 - hbl_godunov (Failed)                -> machine diffs
         83 - rankine (Failed)                    -> LARGEish diffs: 1e-6, 1e-7
         84 - rankine-sym (Failed)                -> LARGEish diffs: 1e-6, 1e-7
         92 - abl_meso_input_dpa (Failed)         -> machine diffs
         93 - abl_meso_input_ipa (Failed)         -> machine diffs
         94 - abl_kosovic_neutral (Failed)        -> machine diffs
        107 - ow_w2a (Failed)                     -> machine diffs
```

I think that matches what you found @mbkuhn and I agree with your conclusions.

mukul1992:
> @mukul1992, could you check this with your print statements?
@mbkuhn 
I checked out your branch and ran the freestream_godunov test with godunov-weno_z after inserting print statements in the LinearProfile UDF.

All boundary fills except the MAC velocity fills happen at time zero.

For the time-advance, the physbc fills happen with old time at some places and new time at others, and the MAC velocity fills still happen at half-times as expected. The time values and where they are happening make sense to me as per my understanding of the scheme. I'll have to insert more elaborate print statements if we want to track down where specific fills are being called from, but it is probably not needed? I'm attaching the log.
[log.txt](https://github.com/user-attachments/files/16181721/log.txt)

amr-wind/core/field_ops.H
mbkuhn: instead of a bool, I went with specifying the number of components for both dst and src

amr-wind/core/field_ops.H
```diff
@@ -67,6 +67,79 @@ add(T1& dst, const T2& src, int srccomp, int dstcomp, int numcomp, int nghost)
...
+            for (int n = 0; n < ncomp_dst; ++n) {
+                amrex::MultiFab::Divide(
+                    dst(lev), src(lev), srccomp, dstcomp + n, ncomp_src,
+                    nghost);
+            }
+        } else {
+            amrex::Abort(
+                "field_ops::divide : number of divisor components (ncomp_src) "
+                "must either be 1 or match number of dividend components "
+                "(ncomp_dst).\n");
```
marchdf, line 109: I don't think you want the `.\n`. Abort adds punctuation for us

amr-wind/equation_systems/BCOps.H
```diff
@@ -39,7 +39,16 @@ struct BCOp<PDE, std::enable_if_t<std::is_base_of_v<ScalarTransport, PDE>>>
...
         auto& field = m_fields.field;
-        field.fillphysbc(m_time.new_time(), ng_diff);
+        const auto bc_time = (rho_state == FieldState::Old)
+                                 ? m_time.current_time()
+                                 : m_time.new_time();
+        if ((rho_state != FieldState::Old && rho_state != FieldState::New)) {
+            amrex::Abort(
+                "BCOps.H apply_bcs(): a state other than New or Old was used. "
+                "The method of setting bc_time must be evaluated before using "
+                "a different state for this routine.\n");
```
marchdf, line 49: same comment here

amr-wind/equation_systems/vof/vof_bcop.H
```diff
@@ -31,7 +31,16 @@ struct BCOp<VOF>
...
         auto& field = m_fields.field;
-        field.fillphysbc(m_time.new_time(), ng_diff);
+        const auto bc_time = (rho_state == FieldState::Old)
+                                 ? m_time.current_time()
+                                 : m_time.new_time();
+        if ((rho_state != FieldState::Old && rho_state != FieldState::New)) {
+            amrex::Abort(
+                "vof_bcop.H apply_bcs(): a state other than New or Old was "
+                "used. The method of setting bc_time must be evaluated before "
+                "using a different state for this routine.\n");
```
marchdf, line 41: and here

marchdf (changes requested):

amr-wind/core/field_ops.H
mbkuhn:
```suggestion
                "(ncomp_dst).");
```

amr-wind/equation_systems/BCOps.H
mbkuhn:
```suggestion
                "a different state for this routine.");
```

amr-wind/equation_systems/vof/vof_bcop.H
mbkuhn:
```suggestion
                "using a different state for this routine.");
```

marchdf (approved):

PR #1129, opened 2024-07-11
set_inflow routine for MAC velocity (enforce Dirichlet BC prior to MAC projection)

mbkuhn:
## Summary

For better boundary conditions, there is a need to enforce/populate the MAC velocity at valid boundary cells. Implementing a set_inflow routine, similar to what's used for the nodal projection, is a method to do that.

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [x] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [x] on CPU 
- the regression tests
  - [ ] on GPU 
  - [x] on CPU 

## Additional background

Issue Number:


amr-wind/equation_systems/icns/icns_advection.cpp
```diff
@@ -109,6 +109,24 @@ void MacProjOp::init_projector(const amrex::Real beta) noexcept
...
 
+void MacProjOp::set_inflow_velocity(amrex::Real time)
+{
+    auto& velocity = m_repo.get_field("velocity");
+    auto& u_mac = m_repo.get_field("u_mac");
+    auto& v_mac = m_repo.get_field("v_mac");
+    auto& w_mac = m_repo.get_field("w_mac");
+
+    for (int lev = 0; lev < m_repo.num_active_levels(); ++lev) {
+        amrex::Array<amrex::MultiFab*, ICNS::ndim> mac_vec;
```
marchdf, line 120: Condense all this:
`amrex::Array<amrex::MultiFab*, AMREX_SPACEDIM> mac_vec = {AMREX_D_DECL(&u_mac(lev), &w_mac(lev), &w_mac(lev))};`

amr-wind/core/Field.H
```diff
@@ -352,6 +352,12 @@ public:
         amrex::MultiFab& mfab,
         const amrex::IntVect& nghost) noexcept;
 
+    void set_inflow_sibling_fields(
+        int lev,
```
marchdf, line 356: const most of these arguments.

amr-wind/core/Field.H
```diff
@@ -379,6 +385,15 @@ public:
         set_inflow(lev, time, mfab, amrex::IntVect(nghost));
     }
 
+    inline void set_inflow_sibling_fields(
+        int lev,
```
marchdf, line 389: same here re const

amr-wind/core/FieldFillPatchOps.H
```diff
@@ -397,6 +403,62 @@ public:
         }
     }
 
+    void set_inflow_sibling_fields(
+        int lev,
+        amrex::Real time,
+        amrex::Array<amrex::MultiFab*, AMREX_SPACEDIM> mfabs,
+        const amrex::IntVect& nghost) override
+    {
+        const int ng = nghost[0];
```
marchdf, line 412: if we are passing an intvect, it makes more sense to move this `ng` and `domain` into the oit loop and get the nghost[dir] there.

amr-wind/equation_systems/icns/icns_advection.cpp
```diff
@@ -117,10 +117,8 @@ void MacProjOp::set_inflow_velocity(amrex::Real time)
     auto& w_mac = m_repo.get_field("w_mac");
 
     for (int lev = 0; lev < m_repo.num_active_levels(); ++lev) {
-        amrex::Array<amrex::MultiFab*, ICNS::ndim> mac_vec;
-        mac_vec[0] = &u_mac(lev);
-        mac_vec[1] = &v_mac(lev);
-        mac_vec[2] = &w_mac(lev);
+        amrex::Array<amrex::MultiFab*, AMREX_SPACEDIM> mac_vec = {
+            AMREX_D_DECL(&u_mac(lev), &w_mac(lev), &w_mac(lev))};
```
marchdf, line 121: well check that out. clang-tidy complains of an unused variable and there is indeed a mistake here. You have `w_mac` instead of `v_mac`.

marchdf (changes requested):

amr-wind/equation_systems/icns/icns_advection.cpp
mbkuhn: nice! not sure how that got there, I thought I copy-pasted

mbkuhn:
Results from the reg tests: 
- rankine and rankine-sym have moderate diffs (10% for velocity_z, .1% for other variables). Diffs are expected for this case; this is the actual goal of the feature. Would like @mukul1992 to confirm manually that the right values are going into this case at the right time.
- abl_godunov_mpl and abl_godunov_mpl_amr have significant diffs, around 10-100%. This is likely related to point 2. above because the nodal projection set_inflow has a special treatment for mpl cases, which I haven't added to this mac projection set_inflow.
- abl_bndry_input* cases and abl_godunov_forcetimetable all have significant diffs, and these are all cases that use boundary planes as input. This also is related to point 2. above. There may still be diffs after I add the special treatment for these cases, but I think they should be smaller than they are currently.

mukul1992:
>     * rankine and rankine-sym have moderate diffs (10% for velocity_z, .1% for other variables). Diffs are expected for this case; this is the actual goal of the feature. Would like @mukul1992 to confirm manually that the right values are going into this case at the right time.

Hi @mbkuhn with this code, I do see my print statements in the LinearProfile UDF being triggered for the MAC velocity valid cells at the half-time before the projection.

mbkuhn:
> with this code, I do see my print statements in the LinearProfile UDF being triggered for the MAC velocity valid cells at the half-time before the projection.

Awesome, thanks for checking that.

mbkuhn:
Update:
I found that the existing MPL code has faulty box logic (just using enclosedCells() does not work for getting the right face indices on the hi side), and that part of the code seems like it should be better integrated with the UDF capability. Because that looks like a refactoring task, I decided that it's beyond the scope of this PR and Mukul's needs.
Second, there are ways that the boundary plane implementation can be improved, but for the sake of this feature, I think we can continue to employ it as is. As a result, there are no diffs for the MPL or boundary plane input cases now.
Lastly, I am updating the MPL reg tests because the deltaT (delta_t) argument changed from a recent merged PR.

@marchdf would you be willing to double check the reg test results? As it stands, I think this is ready to go. We'll categorize the MPL and boundary plane stuff as future work / future study.

marchdf:
I am seeing this fail:
```
The following tests FAILED:
          7 - abl_godunov_mpl (Failed)
         85 - rankine (Failed)
         86 - rankine-sym (Failed)
```

I think that's expected?

amr-wind/core/FieldFillPatchOps.H
```diff
@@ -85,6 +85,11 @@ public:
         amrex::MultiFab& mfab,
         const amrex::IntVect& nghost,
         const FieldState fstate = FieldState::New) = 0;
+
+    virtual void set_inflow_sibling_fields(
+        int lev,
```
marchdf, line 90: missing const on these?

amr-wind/core/FieldFillPatchOps.H
```diff
@@ -397,6 +402,57 @@ public:
...
+        const amrex::Real time,
+        const amrex::Array<amrex::MultiFab*, AMREX_SPACEDIM> mfabs) override
+    {
+        const auto& bctype = m_field.bc_type();
+        const auto& geom = m_mesh.Geom(lev);
+        const auto& gdata = geom.data();
+        const auto& bcfunc = bc_functor();
+
+        for (amrex::OrientationIter oit; oit != nullptr; ++oit) {
+            auto ori = oit();
```
marchdf, line 416: const

amr-wind/core/FieldFillPatchOps.H
```diff
@@ -397,6 +402,57 @@ public:
...
+                if (idir != fdir) {
+                    continue;
+                }
+                const auto& dbx = ori.isLow() ? amrex::bdryLo(domain_box, idir)
+                                              : amrex::bdryHi(domain_box, idir);
+
+#ifdef AMREX_USE_OMP
+#pragma omp parallel if (amrex::Gpu::notInLaunchRegion())
+#endif
+                auto& mfab = *mfabs[fdir];
```
marchdf, line 435: const

marchdf (commented):
Just some minor things.

amr-wind/core/FieldFillPatchOps.H
mbkuhn: can't do this one, mfab gets modified in the set_inflow call

marchdf (approved):

mbkuhn:
> The following tests FAILED:
>           7 - abl_godunov_mpl (Failed)
>           8 - abl_godunov_mpl_amr (Failed)
>          85 - rankine (Failed)
>          86 - rankine-sym (Failed)
> ```

Concluded that the MPL diffs are only due to the correction of the deltaT (delta_t) argument in the reg test input, not due to the executable itself.

PR #1134, opened 2024-07-12
fixes bug in WENO calculations at the xhi domain boundary

### LINKED ISSUE #1126: Potential unidentified bug causes an asymmetry in the solution field
## Bug description
While working with the new in-out BC #1066 and the associated test that has two z-layers of the flow going in opposite directions, I noticed an unexpected asymmetry developing in the second time step. To isolate the issue, I reproduced it with the `mass_inflow` BC. I also tried to create a clean test case which does not rely on the new "two-layer" UDF but wasn't immediately able to reproduce the issue. 

## Steps to reproduce

1. Clone and checkout the [erf-model:inoutbc branch](https://github.com/erf-model/amr-wind/tree/inoutbc). I understand that this is not ideal. I am willing to work with the developers on creating a test case which does not rely on unmerged developments.
2. Compile `amr_wind`, go to the `bctest` directory, and run the `input_inflow` case.

An asymmetry develops in the x- and z-velocity fields in time step 2. Although it doesn't show up in the min/max values printed here, the first place I see it is in the u_mac field before the projection.
Screenshot 2024-07-09 at 4 57 24 PM

## Expected behavior
Since the imposed conditions are symmetric in x and z-directions, I did not expect an asymmetry. As this is happening with the `mass_inflow` BCs too, it may indicate some issue in the MAC-velocity calculations that is not related to the new in-out BC.

mukul1992:
## Summary

Fixes #1126 where unexpected asymmetries were observed.

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [ ] on CPU 
- the regression tests
  - [ ] on GPU 
  - [x] on CPU 

## Additional information

All regression tests are passing, not including the two that require netcdf. One unit test fails but it seems like a machine precision difference if I'm reading the logs correctly. 

Perhaps this bug was not noticed earlier because it may be uncommon to have `ext_dir` BC type at the xhi boundary?

### DIFF
--- a/amr-wind/convection/incflo_godunov_weno.H
+++ b/amr-wind/convection/incflo_godunov_weno.H
@@ -94,10 +94,10 @@ AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE void Godunov_weno_xbc(
 
         } else if (i == domhi - 1) {
 
-            sm = -0.2 * s(domhi + 1, j, k, n) + 0.75 * s(domhi, j, k, n) +
+            sp = -0.2 * s(domhi + 1, j, k, n) + 0.75 * s(domhi, j, k, n) +
                  0.5 * s(domhi - 1, j, k, n) - 0.05 * s(domhi - 2, j, k, n);
 
-            sp = sedge1;
+            sm = sedge1;
         }
     }
 }


marchdf:
Nice work!!! I am running the tests locally.

marchdf:
I am only seeing `dam_break_godunov` fail:

```
 level = 0
 density                            9.102257081e-08           9.102257081e-11
 vof                                9.111367216e-11           9.111367216e-11
 velocityx                          1.858080606e-09           8.790979919e-09
 velocityz                          4.965767873e-09             3.0163951e-08
 gpx                                2.377245956e-05           1.205223434e-09
 gpz                                1.293450211e-06           1.369910817e-10
```

It is a weno case with slip_walls in x (and z) so I would expect it to diff with these changes.

marchdf (approved):

PR #1158, opened 2024-07-26
Fix regression testing documentation

marchdf:
## Summary

This PR improves the documentation of regression testing, particularly comparisons with gold files (and generation of those). For the kind reviewer, it would be helpful to step through it as well to make sure I didn't forget anything.

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

## Additional background

Thanks @asalmgren for bringing this to my attention

### DIFF
--- a/docs/sphinx/developer/regression_testing.rst
+++ b/docs/sphinx/developer/regression_testing.rst
@@ -152,6 +152,17 @@ The gold files directory is organized by ``${OS}/${COMPILER}/${COMPILER_VERSION}
    # Create the reference version of Golds (following the directory convention used above)
    cp -R golds/tmp/ golds/current/
 
+   # Because of test dependencies, this needs to be done twice
+   ctest -j 8
+   cp -R golds/tmp/ golds/current/
+
+.. tip::
+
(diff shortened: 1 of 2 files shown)


mbkuhn:
Not sure what's going on with the failing Docs CI.

In my experience, when I build with spack, the AMR_WIND_TEST_WITH_FCOMPARE variable is on by default. When running the tests for golds, this is almost ok, but the problem is that none of the pass-dependent cases, i.e., the boundary plane input cases, will run, because all of the boundary plane output cases have failed. So then no golds are created for those cases. It might be best to comment on this variable some because I find it pretty important in my experience. 

Also, you might want to add a sentence more of detail explaining that when the tests complain that the plt file is missing, it means that the test simulation completed successfully but no comparison was made.

marchdf:
Figured out why the doc tests failed. 

Added a line re "explaining that when the tests complain that the plt file is missing"

Added something to the effect that you have to run it twice because of test dependencies.

I tried to stay away from any discussion/usage of spack since I think that would be confusing. That would require a whole different section I think.

mbkuhn (approved):
cheers to a documentation update!

PR #1165, opened 2024-07-26
Make inputs for paper cases publicly available

mbkuhn:
## Summary

To provide the inputs for each case included in the paper (submitted, not yet published)

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Other (please describe): very specific example cases

## Additional background

Intended to only be input files and READMEs. Bash scripts are probably fine too if they avoid replicating input files for cases with multiple runs. 

Documentation should also point to the folder in the repo.


mbkuhn:
This goal will be fulfilled with a separate repository, exawind-benchmarks

PR #1168, opened 2024-07-29
Switch some tests to wenoz, add warning for ppm_nolim

marchdf:
## Summary

Switch tests for ppm_nolim to use wenoz. Warn users about using ppm_nolim and recommend using wenoz.

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Other (please describe): test changes and added warning

## Checklist

The following is included:

- [ ] new unit-test(s)
- [x] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [x] on CPU 
- the regression tests
  - [ ] on GPU 
  - [x] on CPU 

## Additional background

The following tests are going to fail because we changed the scheme:
```
The following tests FAILED:
         25 - abl_stable (Failed)
         26 - abl_unstable (Failed)
         27 - abl_unstable_constant_wall_model (Failed)
         28 - abl_unstable_local_wall_model (Failed)
         29 - abl_unstable_schumann_wall_model (Failed)
         40 - uniform_ct_disk (Failed)
         41 - uniform_ct_disk_gaussian (Failed)
         42 - joukowsky_disk (Failed)
         88 - cylinder_refinement (Failed)
```


mbkuhn (changes requested):
The syntax for weno-z is "weno_z". These tests are defaulting to ppm. Which also begs the questions: should the default be weno-z? should the parser also accept "wenoz"?

While we're at it (aligning reg test arguments with standard practice), the diffusion_type should be changed from 2 to 1 in these tests, or that line can just be removed. diffusion_type = 2 is the most stable but the least accurate in time from among the 3 options.

marchdf:
whoops good catch. I am fine with `weno_z` and would prefer just one way of doing it instead of accepting multiple inputs.

Good point on the other but that will be a different PR (coming shortly).

mbkuhn (approved):

PR #1184, opened 2024-08-05
Anelastic projection: multiply/divide all components of velocity

mbkuhn:
## Summary

Found this while working on something else: the anelastic projection is supposed to operate on rho*u, requiring the velocity field to be multiplied by density a priori and divided by density a posteriori. But it looks like only one component of velocity is getting multiplied.

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [x] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU


marchdf:
Good catch!!

PR #1187, opened 2024-08-06
Changing Freesurface post-processing routine to a Sampler

mbkuhn:
## Summary

FreeSurface is designed to find the location of the air-water interface in simulations with vof. This PR converts that capability to a Sampler type, enabling sampling of fields at the interface and native output with AMReX particles.

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [x] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [x] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU


mbkuhn:
I plan to change the names of the files from FreeSurface to FreeSurfaceSampler, but that would make it much harder to review. So I'm saving that for last.

mbkuhn:
Also, I have python scripts to read in the data for every sampler type (ascii, native, and netcdf). I think these should be added to tools, and they'd be set up to read the output from the reg test that I added this sampler too. Thoughts? 

Figuring out the layout of any of these types took me some time, and having an example should be helpful.

mbkuhn:
Screenshot 2024-08-06 at 11 39 27 AM
Screenshot 2024-08-06 at 11 22 02 AM

some evidence it works

marchdf:
Examples are useful! Maybe better in the documentation? I worry about adding python scripts that could go out of date because we can't test them automatically. We can chat more about that.

mbkuhn:
> Examples are useful! Maybe better in the documentation? I worry about adding python scripts that could go out of date because we can't test them automatically. We can chat more about that.

That's a good idea. I learned that jupyter notebooks can be included in sphinx documentation, so I might try that.

mbkuhn:
I set up the docs so that they actually run the jupyter notebooks and generate the plots, and the notebooks are displayed in the docs for easy copy-paste.

mbkuhn:
Here's how the docs look with these changes.

[docs_with_notebooks.tar.gz](https://github.com/user-attachments/files/16590047/docs_with_notebooks.tar.gz)

mbkuhn:
@jrood-nrel do you think you could take a look at what I'm doing with the docs, namely, changing the conf.py and docs.yml files?

@moprak-nrel do you have time to look over the stuff in the source code?

mbkuhn:
> `conf.py` looks fine. I think I would object to committing several binary files here though.

Thanks for the feedback. I'll talk to Marc about what's the best way to incorporate this stuff.

marchdf (approved):

tools/sampling_dam_break_godunov_ascii.py
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/env python3
+
+import numpy as np
+import matplotlib.pyplot as plt
+
+plt.rc('text',usetex=True)
```
marchdf, line 6: in general I like `usetex=True` but this adds the additional requirement that tex be available on the system used. I could go either way on this.

mbkuhn: this is leftover from other scripts where it was necessary. I think I can remove it.

PR #1206, opened 2024-08-19
Include postprocessing scripts in regression testing

mbkuhn:
## Summary

Enables the designation of postprocessing regression tests; the idea is to ensure that the scripts successfully plot the data given the current format of the data that the code generates. They're not intended to check whether or not the values have changed.

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [x] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [ ] on CPU 
- the regression tests
  - [ ] on GPU 
  - [x] on CPU 

## Additional background

Need to figure out how to properly set up the python modules instead of how I have it now.

This test focuses on the three FreeSurfaceSampler scripts and employs the amrex particle script as well. Provides a template for how other scripts (abl stats, actuator) can be included.

Should be merged after #1187.


marchdf:
Can I review this after 1187? Easier to see what the diff will be.

test/CMakeLists.txt
```diff
@@ -78,10 +83,38 @@ macro(setup_test)
     endif()
 endmacro(setup_test)
 
+macro(setup_pp_test)
```
jrood-nrel, line 86: Rather than duplicate this macro, maybe use `if(AMR_WIND_TEST_WITH_PYTHON)` where necessary in the original macro?

test/CMakeLists.txt
```diff
@@ -147,6 +180,25 @@ function(add_test_u TEST_NAME)
...
 endfunction(add_test_u)
 
+# Add test that also checks post-processing scripts: sampling
+function(add_test_pps TEST_NAME FORMAT NSTEPS)
+    setup_pp_test()
+    set(ADDL_RUNTIME_OPTIONS "sampling.output_format=${FORMAT} time.plot_interval=-1 time.checkpoint_interval=-1 time.max_step=${NSTEPS}")
+    # need to somehow only install the requirements once, prior to testing, and then reference their location in testing
+    add_test("${TEST_NAME}_sampling_${FORMAT}" sh -c "${MPI_COMMANDS} ${CMAKE_BINARY_DIR}/${amr_wind_exe_name} ${MPIEXEC_POSTFLAGS} \
+            \ ${CURRENT_TEST_BINARY_DIR}/${TEST_NAME}.inp ${RUNTIME_OPTIONS} ${ADDL_RUNTIME_OPTIONS} > ${TEST_NAME}_sampling_${FORMAT}.log && \
+            \ ${Python_EXECUTABLE} -m venv .venv && source .venv/bin/activate && .venv/bin/pip3 install ${PYTHON_REQUIREMENTS} && \
```
jrood-nrel, line 190: We shouldn't be trying to fulfill dependencies in CMake.

CMakeLists.txt
```diff
@@ -22,6 +22,7 @@ option(AMR_WIND_ENABLE_TESTS "Enable testing suite" OFF)
 option(AMR_WIND_TEST_WITH_FCOMPARE "Check test plots against gold files" OFF)
 option(AMR_WIND_SAVE_GOLDS "Provide a directory in which to save golds during testing" OFF)
 option(AMR_WIND_ENABLE_FPE_TRAP_FOR_TESTS "Enable FPE trapping in tests" ON)
+option(AMR_WIND_TEST_WITH_PYTHON "Enable testing relying on python, like postprocessing" ON)
```
jrood-nrel, line 25: Default for this should be `OFF`.

mbkuhn: thanks, that was an accident

mbkuhn, line 25:
```suggestion
option(AMR_WIND_TEST_WITH_PYTHON "Enable testing relying on python, like postprocessing" OFF)
```

test/CMakeLists.txt
```diff
@@ -4,6 +4,11 @@
 
 configure_file(${CMAKE_CURRENT_SOURCE_DIR}/CTestCustom.cmake ${CMAKE_BINARY_DIR}/CTestCustom.cmake)
 
+# Temporary - needs to be replaced with something better
+if (AMR_WIND_TEST_WITH_PYTHON)
+  set(PYTHON_REQUIREMENTS "numpy netCDF4 matplotlib pathlib pandas")
```
jrood-nrel, line 9: This should be removed. CMake should only be checking for dependencies and not fulfilling them. Also CMake can find Python and numpy I think https://cmake.org/cmake/help/latest/module/FindPython.html but not much beyond that. Maybe there are good ways for CMake to check for Python packages, like `pip freeze` or something.

mbkuhn:
> Can I review this after 1187? Easier to see what the diff will be.

it's a draft, no rush

test/CMakeLists.txt
mbkuhn: realized I didn't need the modified version after all

test/CMakeLists.txt
mbkuhn: good call

test/CMakeLists.txt
mbkuhn: got inspiration from here: https://stackoverflow.com/questions/62553007/cmake-check-python3-and-python-packages-presence

decided that warnings are sufficient; wouldn't want to fail to compile because a post-processing script test won't pass.

mbkuhn:
Little explainer:
* testing python scripts requires python + specific modules; set that up (with warnings about modules) in CMake files
* the reg test first generates the data, which requires specifying the output format and the number of steps (because the typical 10 steps would be too short). This simulation doesn't take too long to do 150 steps.
* after finishing the required number of steps, the corresponding script from tools/  is run, with the expectation that a corresponding .pdf file will be created.
* I have them set up to run in the same directory as the base name, but they output specialized .log files and do not output checkpoint or plot files. That could be changed, but I felt that it'd be unnecessary to make new test_files directories and inputs for them.

CMakeLists.txt
```diff
@@ -192,6 +193,21 @@ if(AMR_WIND_ENABLE_HELICS)
   target_compile_definitions(${amr_wind_lib_name} PRIVATE AMR_WIND_USE_HELICS)
 endif()
 
+if(AMR_WIND_TEST_WITH_PYTHON)
+  set(CMAKE_PREFIX_PATH ${PYTHON_DIR} ${CMAKE_PREFIX_PATH})
+  find_package(Python3 REQUIRED COMPONENTS Interpreter)
+  # Developer note: modify this list if modifying scripts used in python-based regression tests
+  set(PYTHON_REQUIREMENTS numpy netCDF4 matplotlib pathlib pandas)
+  foreach(pymod IN LISTS PYTHON_REQUIREMENTS)
```
marchdf, line 201: could we `venv` this stuff? I would not want to mess with the user's python modules. To be honest, we should "package" the python scripts up properly and then maybe we can use that?

mbkuhn: This is meant to check for the required modules, it doesn't do any installation or setup.

marchdf: ah sorry I missed the 'show"

marchdf:
Can you add the new cmake option to the documentation. Somewhere around: https://exawind.github.io/amr-wind/user/build.html#cmake-configuration-reference

CMakeLists.txt
```diff
@@ -192,6 +193,21 @@ if(AMR_WIND_ENABLE_HELICS)
   target_compile_definitions(${amr_wind_lib_name} PRIVATE AMR_WIND_USE_HELICS)
 endif()
 
+if(AMR_WIND_TEST_WITH_PYTHON)
+  set(CMAKE_PREFIX_PATH ${PYTHON_DIR} ${CMAKE_PREFIX_PATH})
+  find_package(Python3 REQUIRED COMPONENTS Interpreter)
+  # Developer note: modify this list if modifying scripts used in python-based regression tests
+  set(PYTHON_REQUIREMENTS numpy netCDF4 matplotlib pathlib pandas)
```
marchdf, line 200: pathlib should be removed since it is core python. Otherwise you get:
`WARNING: Package(s) not found: pathlib` 

```suggestion
  set(PYTHON_REQUIREMENTS numpy netCDF4 matplotlib pandas)
```

mbkuhn: does that mean I don't need to have import Path from pathlib in the scripts? I think it's in the amrex particle script

docs/sphinx/user/build.rst
```diff
@@ -126,6 +126,11 @@ Other AMR-Wind specific options
 
    Directory for saving gold files when using :program:`fcompare`. Default: ""
 
+.. cmakeval:: AMR_WIND_TEST_WITH_PYTHON
+
+   Enable performing tests that rely on python, primarily consists of checks 
+   for post-processing scripts in tools directory. Default: OFF
+
```
mbkuhn, line 133: I said "primarily" here because I'm assuming this option is also needed for the verification test (not currently running) that has a python script. This PR doesn't hook up this option to the verification test, but that might need to be done in the future

CMakeLists.txt
marchdf: no you still need to import it, it's just always there as part of python, it's not an external package

CMakeLists.txt
```diff
@@ -192,6 +193,21 @@ if(AMR_WIND_ENABLE_HELICS)
...
+  foreach(pymod IN LISTS PYTHON_REQUIREMENTS)
+    execute_process(COMMAND pip3 show ${pymod}
```
marchdf, line 202: instead of using pip here, could we capture the exit code of running `python -c "import ${pymod}"`  Then we are sure it is using the right interpreter.

marchdf (approved):

CMakeLists.txt
mbkuhn: I like that; good to know for the future. I was running into mismatching pips and pythons on my mac.

PR #1207, opened 2024-08-19
Reduce memory usage of native sampling

marchdf:
## Summary

Reduce memory usage of native sampling by not allocating/using `m_output_buf` and `m_sample_buf`, which are only used for the netcdf pathway. I've included a script to diff particle files. We can hook it up when #1206 is merged. 

## Pull request type

Please check the type of change introduced:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe):

## Checklist

The following is included:

- [ ] new unit-test(s)
- [x] new regression test(s)
- [ ] documentation for new capability

This PR was tested by running:

- the unit tests
  - [ ] on GPU 
  - [ ] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU

Issue #1186, opened 2024-08-06
Out-of-memory crash with large sampling


rthedin:
[static_box.txt](https://github.com/user-attachments/files/16618593/static_box.txt)
AMR-Wind crashes on Kestrel CPU nodes with out-of-memory error when large sampling is requested. The error is below:
```
 slurmstepd: error: Detected 1 oom_kill event in StepId=4893832.0. Some of the step tasks have been OOM Killed.
 srun: error: x1008c5s2b0n0: task 108: Out Of Memory
```

Here is the sampling portion that is creating the out-of-memory error:
```
incflo.post_processing                =  box_lr 

box_lr.output_format    = netcdf
box_lr.output_frequency = 4
box_lr.fields           = velocity
box_lr.labels           = Low1 

box_lr.Low1.type         = PlaneSampler
box_lr.Low1.num_points   = 1451 929
box_lr.Low1.origin       = -985.0000 -10045.0000 5.0000
box_lr.Low1.axis1        = 29000.0000 0.0 0.0
box_lr.Low1.axis2        = 0.0 18560.0000 0.0
box_lr.Low1.normal       = 0.0 0.0 1.0
box_lr.Low1.offsets      = 0.0 20.0 40.0 60.0 80.0 100.0 120.0 140.0 160.0 180.0 200.0 220.0 240.0 260.0 280.0 300.0 320.0 340.0 360.0 380.0 400.0 420.0 440.0 460.0 480.0 500.0 520.0 540.0 560.0 580.0
```

I understand I'm asking for ~43M points, which will come with performance slowdowns. I can deal with slowdown, but I find it weird it is crashing. The memory footprint of u, v, and w of all these values should be about 1GB.

Things I tried (none worked):
- AMR-Wind compiled with GNU
- AMR-Wind compiled with Intel (oneapi classic)
- AMR-Wind at different commit points (_not_ the very latest, though-- note my `offset` keyword hasn't changed yet.)
- `native` instead of `netcdf`
- Split the large sampling into smaller ones (keeping the total sampling the same)
- Runs with different number of nodes (all the way to 50)

Some observations:
- If I comment out about half of the `offsets` and request >=4 nodes, it works. 
  - If I request very few nodes (2 or 3), it crashes in the first time step (around the temperature_solve), before any actual sampling is about to take place.
  - If I request more nodes, (>=4), the main time loop starts, and it crashes on the very first time step where such sampling is happening (in the example above, the 4th). 
[...]

PR #1209, opened 2024-08-21
Reduce memory usage of the samplers again

marchdf:
## Summary

Another round of memory optimization for the samplers. It is only necessary to reallocate the container if the positions of the sample points are updated. So this changes it such that if the particles have changed positions, then the container is updated. Otherwise, it skips it. See below for more information.

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [ ] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU 

## Additional background

This is using the same case I described here: https://github.com/Exawind/amr-wind/issues/1186#issuecomment-2294018942, using 30 planes.

### Main branch (after #1207, my previous optimization)
memory_30plane_v1

```
Time spent in InitData():    17.13284979
Time spent in Evolve():      17.09655093
TinyProfiler total time across processes [min...avg...max]: 34.09 ... 34.23 ... 34.25
``` 

### This PR
memory_30plane_v2

```
Time spent in InitData():    10.87862356
Time spent in Evolve():      4.578913001
TinyProfiler total time across processes [min...avg...max]: 15.3 ... 15.44 ... 15.46
```

### Conclusion

- This PR removes (over 2 time steps), 3 of the 4 huge memory spikes on rank 0.
- Instead of 40GB of memory spike, it is now only a single 10GB memory spike, or a 4x improvement
- We also got a speedup: 4.25X per time step (2X over the total run time, 1.7 for init)

### Next steps:

See if I can remove the memory spike all together by allocating the particles in parallel instead of rank 0 + redistribute. Or maybe I go back to looking at the netcdf path.

Following up on issue #1186 and PR #1207

### DIFF
--- a/amr-wind/utilities/sampling/Sampling.cpp
+++ b/amr-wind/utilities/sampling/Sampling.cpp
@@ -164,10 +164,14 @@ void Sampling::update_sampling_locations()
 {
     BL_PROFILE("amr-wind::Sampling::update_sampling_locations");
 
-    if (std::any_of(m_samplers.begin(), m_samplers.end(), [](const auto& obj) {
-            return obj->update_sampling_locations();
-        })) {
-        update_container();
+    amrex::Vector<bool> updated_position(false);
+    for (const auto& obj : m_samplers) {
+      const bool updated_pos = obj->update_sampling_locations();
+      updated_position.push_back(updated_pos);
+    }
+
+    if (std::any_of(updated_position.begin(), updated_position.end(), [](const auto& v) { return v; })) {
+      update_container();
     }
 }
 
--- a/amr-wind/utilities/sampling/Sampling.cpp
+++ b/amr-wind/utilities/sampling/Sampling.cpp
@@ -164,7 +164,7 @@ void Sampling::update_sampling_locations()
 {
     BL_PROFILE("amr-wind::Sampling::update_sampling_locations");
 
-    amrex::Vector<bool> updated_position(false);
+    amrex::Vector<bool> updated_position;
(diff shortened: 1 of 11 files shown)


amr-wind/utilities/sampling/Sampling.cpp
```diff
@@ -136,11 +136,11 @@ void Sampling::update_sampling_locations()
 {
     BL_PROFILE("amr-wind::Sampling::update_sampling_locations");
 
-    for (const auto& obj : m_samplers) {
-        obj->update_sampling_locations();
+    if (std::any_of(m_samplers.begin(), m_samplers.end(), [](const auto& obj) {
+            return obj->update_sampling_locations();
+        })) {
+        update_container();
```
mbkuhn, line 142: I could be wrong, but it seems like update_container is overkill for when the positions of the samplers change. Some samplers change the number of points, and update_containers may be required for those, but it seems off that the particle container would need to be set up from scratch every time the positions change. but maybe that distinction could be put in another PR (e.g., flag that differentiates between samplers that are static, samplers that do not change number of points but change positions, and samplers that change both).

marchdf: You are absolutely right. Trying to keep these PRs small. But that is another area of improvement.

mbkuhn (approved):
love the confidence in your commit message

Issue #1058, opened 2024-05-15
Add time-vary & spatially varying body force for MMC simulations


lawrenceccheung:
## Is this feature request related to a problem? Please describe.

For MMC coupled wind farm simulations, we need to add a body force which varies in both z and in time.  The body forces are derived from the `ABLMesoscaleForcing` and are horizontally homogeneous.  This feature was alluded to in #1022 and #1025, and would generalize the capabilities currently available.

## Describe the preferred solution

One possibility is to have the body forces be read from the abl stats netcdf file directly, which would avoid any reformatting of data (the abl stats file contains `abl_meso_forcing_mom_x` and `abl_meso_forcing_mom_y` which is what we want).

The same body forcing term would be necessary for the temperature equation here: https://github.com/Exawind/amr-wind/blob/main/amr-wind/equation_systems/temperature/source_terms/BodyForce.cpp.

## Additional context

This capability is/will be necessary for FLOWMAS, AWAKEN, and other such projects.

ewquon:
Hi all, is anyone actively looking to implement this? Didn't realize this capability wasn't in the code yet. I also need to be able to apply momentum/temperature source terms that are time and height varying. Thanks

marchdf:
I don't know that anyone is actively implementing this feature. Do you want to take a stab? Happy to discuss but I think you implemented the first part of this so you probably know more than I do ;)

udhaya-chandiran:
I am participating in the AWAKEN Benchmark simulation campaign (IEA Wind Task 57). AMR-Wind will best suit my needs if it simulates a diurnal cycle. I can see from release notes [1.4.0](https://github.com/Exawind/amr-wind/releases/tag/v1.4.0) and [2.1.0](https://github.com/Exawind/amr-wind/releases/tag/v2.1.0) that time-varying ABLForcing is implemented and updated. However, I would like to know if time-varying surface temperature or surface heat flux can be prescribed as a boundary condition. If not, could you let me know when I can expect this capability?

moprak-nrel:
I'm curious if the stuff in `ABLMesoForcingMom` can be used for this? It seems to have the forcing as a function of both `z` and `t`.

I don't have the context for this code path, but it seems to be adding a body force term based on the velocity plane averages and the mesoscale inputs from a netcdf file in this function:
[...]

PR #1133, opened 2024-07-12
Non Uniform Roughness, Temperature and Heat Flux at Surface 

hgopalan:
## Summary

This draft PR proposes a framework to include non-uniform roughness length, temperature and heat flux at the lower surface for handling spatially varying roughness length in  large wind farms, spatial-temporal varying one-way roughness-based wave coupling, spatial-temporal varying temperature and sensible heat flux. 

MOST is replaced with an explicit stratification based parametrization. Currently, the method does not work work terrains and will be updated in the future. 

The data is read in an ascii format to avoid the netCDF dependency. Will create a separate preprocessing tool to convert netCDF data into the ascii format. 

ToDo: 

1. Add wave-roughness information 
2. Add variable heat flux 

Caution: It is assumed that the time-period of input dataset is greater than CFD run-time. This avoids the need for special one-sided interpolation at the end of the dataset. 

Please check the type of change introduced:

- [ ] Bugfix
- [ x ] Feature
- [ ] Code style update (formatting, renaming)
- [ x ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe):

## Checklist

The following is included:

- [ ] new unit-test(s)
- [ x ] new regression test(s)
- [ ] documentation for new capability

This PR was tested by running:

- the unit tests
  - [ ] on GPU 
  - [ ] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ x ] on CPU 

## Additional background

Issue Number:

PR #1219, opened 2024-08-28
Fixes to MMC tendency forcing

moprak-nrel:
## Summary

This PR tracks fixes to MMC tendency forcing. The idea behind tendency forcing is to apply a time/height varying body force based on an input file, which can be used either with outputs from `ABLMesoForcingMom` or just a general list of time/z varying body forces in `u, v, theta` and a time varying surface-flux. 

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [x] on CPU 
- the regression tests
  - [ ] on GPU 
  - [x] on CPU 

## Additional background

Issue Number: #1058


moprak-nrel:
Note that this is still in draft stage. I'm still in the process of tying some loose ends with the temperature flux, and need to run more tests to ensure this works with refinement zones. Docs and regtests are incoming too.

hgopalan:
Thank you for taking up this task. How are we going to plan the work flow for the MMC coupling since there are several ways of doing it: 

1. Drive using WRF data boundary conditions - We will need a temperature perturbations in upstream to generate flow structures. We can use the method integrated into ERF. 
2. Drive using surface conditions from WRF - We would need heterogeneous roughness lengths, temperature/heat flux for this. 
3.  Drive using WRF data as forcing 
4. Drive using met-mast profile as forcing. 

Do you want to keep the different driving mechanisms separate or have a single unified framework for all of them?

moprak-nrel:
We might want to move this discussion to a relevant issue, but some quick responses to your questions below:

> Thank you for taking up this task. How are we going to plan the work flow for the MMC coupling since there are several ways of doing it:
> 
>     1. Drive using WRF data boundary conditions - We will need a temperature perturbations in upstream to generate flow structures. We can use the method integrated into ERF.

I'm not entirely sure what you mean here, are you talking about an inflow boundary condition with temperature perturbations? Or is this on the wall?

>     2. Drive using surface conditions from WRF - We would need heterogeneous roughness lengths, temperature/heat flux for this.
I thought #1133 was for addressing some of these?

>     3. Drive using WRF data as forcing
Not sure what this means either, are we talking about time varying mean profiles? 3D flow-fields? boundary conditions? Some combination of all of those?

>     4. Drive using met-mast profile as forcing.
Is this a 1-D profile of velocities/temperatures? The current `ABLMesoMomForcing` should work right?

 
> Do you want to keep the different driving mechanisms separate or have a single unified framework for all of them?

It's hard to say without understanding all the driving mechanisms.

lawrenceccheung:
Just to provide some context here, this PR is meant to help address https://github.com/Exawind/amr-wind/issues/1058.  This method is meant to provide the "internal forcing" MMC coupling approach, where the horizontally averaged profiles are driven to match given 1D profiles in z and in time.  In particular, this should help provide internal body forces for the wind farm runs, when the precursor runs are complete and `ABLMesoMomForcing` and `ABLMesoTempForcing` are turned off (@ewquon please correct anything I'm saying here).

The source of the forcing data in this case would be from the precursor run, and the source of the profile data for the precursor run could be from either WRF data, lidar profilers, or met-mast data, depending on what's available.  This method _won't_ provide any help for the "boundary forcing" approach, where we'd take inflow data from WRF/ERF simulations and drive AMR-Wind with that.  After we get this settled, we might have a meeting with @ewquon, @mchurchf, and any other interested parties to see where to take things next.

Lawrence

amr-wind/equation_systems/icns/source_terms/ABLMesoForcingMom.cpp
```diff
@@ -53,29 +57,21 @@ void ABLMesoForcingMom::mean_velocity_init(const ABLMesoscaleInput& ncfile)
...
-    m_vAvg_vals.resize(vavg.ncell_line());
-
-    m_meso_avg_error.resize(vavg.ncell_line());
-
-    m_error_meso_avg_U.resize(vavg.ncell_line());
-    m_error_meso_avg_V.resize(vavg.ncell_line());
-
-    m_err_U.resize(vavg.ncell_line());
-    m_err_V.resize(vavg.ncell_line());
+    m_velAvg_ht.resize(m_nht);
```
marchdf, line 70: silly comment but I will be taking care of this at some point. Can you make this a snake case variable name?

marchdf (approved):

marchdf:
Can you add some documentation about this? Or maybe it's already there? And also a reg test (unless that is also covered)?

moprak-nrel:
Thanks for reviewing, doc changes and reg tests are incoming shortly, wanted to get started on the code review process.

moprak-nrel:
@ewquon Would like to get your eyes on this PR if you get a chance too. For some reason I couldn't add you as a reviewer.

moprak-nrel:
Documenting of some of my tests:

[Here](https://github.com/user-attachments/files/16949748/norm_diffs.pdf) are normed differences of statistics of velocity/theta, the three lines are:

1. `PR`: This PR vs the main branch using MMC forcing
2. `tendency`: Tendency forcing vs MMC forcing
3. `tendency interpolated`: Tendency forcing vs MMC forcing, but the input file is at a coarser time/space resolution

As can be seen, 1 and 2 match exactly, and there is only a small interpolation error in 3.

moprak-nrel:
Documentation has no changes, this is already present [here](https://exawind.github.io/amr-wind/user/inputs_ABL_meso_forcing.html).

PR #1226, opened 2024-09-02
Fixed point iterations within a physical time step for the advection term

itopcuoglu:


amr-wind/equation_systems/icns/icns_advection.H
```diff
@@ -190,11 +190,14 @@ struct AdvectionOp<ICNS, fvm::Godunov>
                 m_allow_inflow_on_outflow = true;
             }
 
+            // if state is NPH, then n and n+1 are known, and only
+            // spatial extrapolation is performed
+            amrex::Real dt_arg = (fstate == FieldState::NPH) ? 0.0 : dt;
```
marchdf, line 195: let's give this a better name than `dt_arg`. `dt_for_hydro` ? `dt_local` ? `dt_adjusted`?

amr-wind/equation_systems/AdvOp_Godunov.H
```diff
@@ -168,6 +168,10 @@ struct AdvectionOp<
                         (godunov_scheme == godunov::scheme::WENO_JS)) {
                         m_allow_inflow_on_outflow = true;
                     }
+                    // if state is NPH, then n and n+1 are known, and only
+                    // spatial extrapolation is performed
+                    amrex::Real dt_arg = (fstate == FieldState::NPH) ? 0.0 : dt;
```
marchdf, line 173: same comment below about naming

amr-wind/incflo.H
```diff
@@ -172,6 +174,9 @@ private:
     // Prescribe advection velocity
     bool m_prescribe_vel = false;
 
+    // Advection iterations every timestep
```
marchdf, line 177: maybe something with non_linear_iterations in the name? I am a bit wary of our naming for these things because we often mean different things when we say "nonlinear". I am not sure what the right name is but we need to be clear and make sure we stick with it. I have no problem with long, explicit names: e.g. `m_non_linear_advection_iterations`.  Let's add some documentation too.

amr-wind/incflo.cpp
```diff
@@ -372,6 +378,10 @@ void incflo::init_physics_and_pde()
         if (activate_overset) {
             m_sim.activate_overset();
         }
+
+        // Get number of advection iterations
+        pp.query("advection_iterations", m_adv_iters);
```
marchdf, line 383: and when we update the naming, make sure the user facing input makes sense within that context as well.

amr-wind/incflo.cpp
```diff
@@ -372,6 +378,10 @@ void incflo::init_physics_and_pde()
...
+        m_adv_iters = std::max(1, m_adv_iters);
```
marchdf, line 384: do we want to silently do this? Or abort on an invalid value? Or warn that we are doing this?

amr-wind/incflo_advance.cpp
```diff
@@ -399,6 +410,288 @@ void incflo::ApplyPredictor(bool incremental_projection)
     }
 }
 
+// Predictor step for subsequent non-linear iterations within each timestep
+void incflo::ApplyPredictorNonLinear(bool incremental_projection)
+{
+    BL_PROFILE("amr-wind::incflo::ApplyPredictorNonLinear");
```
marchdf, line 416: I've mentioned this before but noting it here so I don't forget (it's not you, it's me). I don't like all the duplicate code this triggers with the other ApplyPredictor. It would be good to merge somehow.

amr-wind/utilities/console_io.cpp
```diff
@@ -214,4 +215,75 @@ void print_tpls(std::ostream& out)
...
+            level_mask.setVal(1);
+        }
+
+        const auto& vnew = velocity_new(lev);
+        // const auto& velstar = vel_np1_old(lev);
+        // auto& veldiff = vel_diff(lev);
+        const auto& velstar = vel_star(lev);
+        auto& veldiff = vel_diff(lev);
+
+        for (amrex::MFIter mfi(velocity_new(lev)); mfi.isValid(); ++mfi) {
```
marchdf, line 250: make this a fused mfiter

amr-wind/utilities/console_io.cpp
```diff
@@ -214,4 +215,75 @@ void print_tpls(std::ostream& out)
...
+            const auto bx = mfi.tilebox();
+            const auto& velstar_arr = velstar.const_array(mfi);
+            const auto& veldiff_arr = veldiff.array(mfi);
+            const auto& velnew_arr = vnew.const_array(mfi);
+            const auto& levelmask_arr = level_mask.const_array(mfi);
+
+            amrex::ParallelFor(
```
marchdf, line 257: make a parallelfor on AMREX_SPACEDIM

amr-wind/utilities/console_io.cpp
```diff
@@ -214,4 +215,75 @@ void print_tpls(std::ostream& out)
...
+                        (velnew_arr(i, j, k, 1) - velstar_arr(i, j, k, 1)) *
+                        levelmask_arr(i, j, k);
+                    veldiff_arr(i, j, k, 2) =
+                        (velnew_arr(i, j, k, 2) - velstar_arr(i, j, k, 2)) *
+                        levelmask_arr(i, j, k);
+                });
+        }
+    }
+
+    amrex::Real rms_ucell = 0.0;
```
marchdf, line 272: use amrex::Array of size spacedim, stuff the norms into there with a component loop.

amr-wind/utilities/console_io.cpp
```diff
@@ -214,4 +215,75 @@ void print_tpls(std::ostream& out)
...
+    amrex::Real rms_vcell = 0.0;
+    amrex::Real rms_wcell = 0.0;
+
+    for (int lev = 0; lev < nlevels; ++lev) {
+        rms_ucell += vel_diff(lev).norm2(0);
+        rms_vcell += vel_diff(lev).norm2(1);
+        rms_wcell += vel_diff(lev).norm2(2);
+    }
+    amrex::Print() << "Non-linear residual for u velocity " << rms_ucell
```
marchdf, line 281: for loop these, or print as a single line.

marchdf: Also make sure the answer is the same when using MPI and multiple procs

amr-wind/utilities/console_io.cpp
```diff
@@ -214,4 +215,75 @@ void print_tpls(std::ostream& out)
     }
 }
 
+void print_nonlinear_residual(
```
marchdf, line 218: I am not sure this print function belongs here. It feels different than the others. Not sure where to put it yet. I wonder if we can also make it more generic by templating to accept fields. The fields should also have the repo/mesh info so we don't need to pass in sim I think. Maybe this is more of a field utils kinda thing.

marchdf (changes requested):

amr-wind/equation_systems/icns/icns_advection.H
itopcuoglu: Maybe dt_extrap as in dt for extrapolation?

amr-wind/incflo.H
itopcuoglu: We could go for something like `m_outer_loop_iterations`. I am going to modify the loops to make this compatible with exawind-driver, and I think that is going to make it clear that these are outer loop iterations

amr-wind/equation_systems/icns/icns_advection.H
itopcuoglu: Changed this into dt_extrap, as in the dt that is used for extrapolation.

amr-wind/equation_systems/AdvOp_Godunov.H
itopcuoglu: Changed this into dt_extrap as well.

amr-wind/incflo.cpp
itopcuoglu: Got rid of this.

amr-wind/utilities/console_io.cpp
itopcuoglu: This has been fixed.

amr-wind/utilities/console_io.cpp
itopcuoglu: This has been fixed.

amr-wind/utilities/console_io.cpp
itopcuoglu: I changed this into printing a single line.

The non-linear solver residuals are the same down to roughly the 5th decimal between 4, 6 and 8 ranks.

amr-wind/utilities/console_io.cpp
marchdf: fused mfiter operate on multifabs. no need for fields.

itopcuoglu: I converted this into a fused MFIter.

818Nawaf (approved):

818Nawaf (approved):

mbkuhn:
I don't have anything to add to Marc's remaining comments. This looks really good, I'm impressed with how simply and directly you have managed to set this up.

amr-wind/equation_systems/AdvOp_Godunov.H
```diff
@@ -168,6 +168,10 @@ struct AdvectionOp<
                         (godunov_scheme == godunov::scheme::WENO_JS)) {
                         m_allow_inflow_on_outflow = true;
                     }
+                    // if state is NPH, then n and n+1 are known, and only
+                    // spatial extrapolation is performed
+                    amrex::Real dt_extrap = (fstate == FieldState::NPH) ? 0.0 : dt;
```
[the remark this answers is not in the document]
itopcuoglu: Addressed this in a separate commit.

amr-wind/incflo_advance.cpp
```diff
@@ -336,7 +351,11 @@ void incflo::ApplyPredictor(bool incremental_projection)
     }
 
     // With scalars computed, compute advection of momentum
-    icns().compute_advection_term(amr_wind::FieldState::Old);
+    if (inonlin == 0) {
```
marchdf, line 354: can we do: `const auto fstate = (inonlin==0)? amr_wind::FieldState::Old: amr_wind::FieldState::NPH;` and then just `icns().compute_advection_term(fstate);`

amr-wind/incflo_advance.cpp
```diff
@@ -47,15 +47,17 @@ void incflo::pre_advance_stage2()
  *
  * \callgraph
  */
-void incflo::advance()
+void incflo::advance(int inonlin)
 {
     BL_PROFILE("amr-wind::incflo::advance");
+    if (inonlin == 0) m_sim.pde_manager().advance_states();
```
marchdf, line 53: I don't understand why the CI isn't complaining about this.

marchdf: looks like we just don't enforce it. Weird.

itopcuoglu: I'll manually add it for now.

amr-wind/incflo_advance.cpp
itopcuoglu: I changed this as such.

amr-wind/equation_systems/icns/icns_advection.H
```diff
@@ -190,11 +190,14 @@ struct AdvectionOp<ICNS, fvm::Godunov>
                 m_allow_inflow_on_outflow = true;
             }
 
+            // if state is NPH, then n and n+1 are known, and only
+            // spatial extrapolation is performed
+            amrex::Real dt_extrap = (fstate == FieldState::NPH) ? 0.0 : dt;
```
marchdf, line 195:
```suggestion
            const amrex::Real dt_extrap = (fstate == FieldState::NPH) ? 0.0 : dt;
```

amr-wind/incflo_advance.cpp
```diff
@@ -47,15 +47,20 @@ void incflo::pre_advance_stage2()
  *
  * \callgraph
  */
-void incflo::advance()
+void incflo::advance(int inonlin)
```
marchdf, line 50:
```suggestion
void incflo::advance(const int inonlin)
```

amr-wind/incflo.H
```diff
@@ -96,8 +96,9 @@ public:
     bool regrid_and_update();
     void pre_advance_stage1();
     void pre_advance_stage2();
-    void do_advance();
-    void advance();
+    void do_advance(int inonlin);
```
marchdf, line 99:
```suggestion
    void do_advance(const int inonlin);
```

amr-wind/incflo.H
```diff
@@ -133,7 +134,7 @@ public:
     void ComputeDt(bool explicit_diffusion);
     void ComputePrescribeDt();
 
-    void ApplyPredictor(bool incremental_projection = false);
+    void ApplyPredictor(bool incremental_projection = false, int inonlin = 0);
```
marchdf, line 137:
```suggestion
    void ApplyPredictor(bool incremental_projection = false, const int inonlin = 0);
```

amr-wind/incflo_advance.cpp
```diff
@@ -47,15 +47,20 @@ void incflo::pre_advance_stage2()
...
 {
     BL_PROFILE("amr-wind::incflo::advance");
+    if (inonlin == 0) {
+        m_sim.pde_manager().advance_states();
+    };
```
marchdf, line 55:
```suggestion
    }
```

amr-wind/incflo_advance.cpp
```diff
@@ -165,10 +170,9 @@ void incflo::advance()
  *  <li> \ref incflo::ApplyProjection "Apply projection"
  *  </ol>
  */
-void incflo::ApplyPredictor(bool incremental_projection)
+void incflo::ApplyPredictor(bool incremental_projection, int inonlin)
```
marchdf, line 173:
```suggestion
void incflo::ApplyPredictor(bool incremental_projection, const int inonlin)
```

amr-wind/incflo_advance.cpp
```diff
@@ -268,7 +282,11 @@ void incflo::ApplyPredictor(bool incremental_projection)
     }
 
     // Extrapolate and apply MAC projection for advection velocities
-    icns().pre_advection_actions(amr_wind::FieldState::Old);
+    if (inonlin == 0) {
```
marchdf, line 285: Store the field state with a ternary expression then call pre_advection_actions with that variable.

amr-wind/utilities/console_io.H
```diff
@@ -20,6 +21,9 @@ void print_mlmg_info(const std::string& solve_name, const amrex::MLMG& mlmg);
 
 void print_tpls(std::ostream& /*out*/);
 
+void print_nonlinear_residual(
+    CFDSim& sim, ScratchField& vel_diff, ScratchField& vel_star);
```
marchdf, line 25:
```suggestion
    const CFDSim& sim, const ScratchField& vel_diff, const ScratchField& vel_star);
```

amr-wind/utilities/console_io.cpp
```diff
@@ -214,4 +215,61 @@ void print_tpls(std::ostream& out)
...
+                        (velnew_arr[nbx](i, j, k, n) -
+                         velstar_arr[nbx](i, j, k, n)) *
+                        levelmask_arr[nbx](i, j, k);
+                }
+            });
+    }
+
+    amrex::Array<amrex::Real, AMREX_SPACEDIM> rms_vel;
+
+    for (int idim = 0; idim < AMREX_SPACEDIM; ++idim) rms_vel[idim] = 0.;
```
marchdf, line 263: remove this line. When declaring just fill it with zeros there. `rms_vel = {0.0};`

marchdf (changes requested):
Let some suggestions you should be able to just hit the commit button on. There's just a couple little pieces left to address. Good work!!

amr-wind/utilities/console_io.cpp
itopcuoglu: Fixed this.

amr-wind/incflo_advance.cpp
```diff
@@ -336,11 +354,18 @@ void incflo::ApplyPredictor(bool incremental_projection)
     }
 
     // With scalars computed, compute advection of momentum
-    icns().compute_advection_term(amr_wind::FieldState::Old);
+    // if (inonlin == 0) {
```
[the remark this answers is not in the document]
itopcuoglu: The comments have been removed.

amr-wind/incflo_advance.cpp
itopcuoglu: I removed the if statements and added a ternary expression the same way it was done for `icns()compute_advection_term()`.

amr-wind/utilities/console_io.cpp
```diff
@@ -214,4 +215,60 @@ void print_tpls(std::ostream& out)
...
+
+    for (int lev = 0; lev < nlevels; ++lev) {
+        for (int idim = 0; idim < AMREX_SPACEDIM; ++idim) {
+            rms_vel[idim] += vel_diff(lev).norm2(idim);
+        }
+    }
+
+    amrex::Print() << "Norm of change over advection iterations in u  "
+                   << rms_vel[0] << " v " << rms_vel[1] << " w " << rms_vel[2]
+                   << std::endl;
```
marchdf, line 271:
```suggestion
    amrex::Print() << "Norm of change over advection iterations in u:  "
                   << rms_vel[0] << ", v: " << rms_vel[1] << ", w: " << rms_vel[2]
                   << std::endl;
```

amr-wind/incflo.H
```diff
@@ -96,8 +96,9 @@ public:
     bool regrid_and_update();
     void pre_advance_stage1();
     void pre_advance_stage2();
-    void do_advance();
-    void advance();
+    void do_advance(const int inonlin);
+    void post_overset_work();
```
mbkuhn, line 100: is this leftover from previous development work? this appears to be the only instance of `post_overset_work`

mbkuhn:
We were going to rename it from "advection iterations", right? Because it is redoing the whole time step, not just the advection. Not sure if you guys decided whether nonlinear iterations or outer loop iterations was a better name.

amr-wind/incflo.H
itopcuoglu: Yes, the rebase must have left it in there. I will correct it. Good catch.

itopcuoglu: Fixed this.

itopcuoglu:
> We were going to rename it from "advection iterations", right? Because it is redoing the whole time step, not just the advection. Not sure if you guys decided whether nonlinear iterations or outer loop iterations was a better name.

I did not want to conflate the non-linear iterations done by exawind-driver for the hybrid solver with the AMR-Wind only iterations that can be used when AMR-Wind is run stand-alone. So I chose to to leave the `m_adv_iters` naming when the number of iterations are parsed through AMR-Wind, and I called it non-linear iterations when the number of iterations are parsed through exawind-driver. Especially since we aren't using the `NPH` state in the diffusion step yet. 

I am definitely open to suggestions on the naming convention.

amr-wind/incflo_advance.cpp
```diff
@@ -397,8 +414,25 @@ void incflo::ApplyPredictor(bool incremental_projection)
     if (m_verbose > 2) {
         PrintMaxVelLocations("after nodal projection");
     }
+    // ScratchField to store the old np1
+    if (inonlin > 0) {
```
mbkuhn, line 418: I think it might be good to add a verbosity condition here:
```suggestion
    if (inonlin > 0 && m_verbose > 0) {
```

amr-wind/incflo.cpp
```diff
@@ -309,15 +310,17 @@ void incflo::Evolve()
...
     if (m_sim.has_overset()) {
         m_ovst_ops.pre_advance_work();
     }
-    if (m_prescribe_vel) {
+    if (m_prescribe_vel && inonlin == 0) {
         prescribe_advance();
     } else {
-        advance();
+        amrex::Print() << "Iteration " << inonlin
+                       << " in advection iteration loop" << std::endl;
```
mbkuhn, line 322: does this get printed only in the amr-wind standalone case, and not when running exawind-driver? Just wanted to check, related to your comment on distinguishing the two.

mbkuhn:
There is at least some precedent to the fixed point name: https://www.sciencedirect.com/science/article/pii/S0021999114006317

But I'm also open to other suggestions.

itopcuoglu:
Maybe we could change the iteration loop index `inonlin` in the function `Evolve()` to `iadv` to present that we are doing iterations over advection. And then the `inonlin`argument in the `do_advance()` function prototype could be changed into `iter_fixed_pt` to generalize its definition, and to keep it fairly consistent with calls from both exawind-driver and from AMR-Wind. And exawind-driver could keep calling `do_advance()` with the argument `inonlin` since we have a non-linear iteration loop that is already defined in exawind-driver. That way the `do_advance()` function itself it defined with "fixed point iterations", but we are also allowing the context of the call to make itself obvious to other developers.

marchdf:
Fine with all that, except I advocate for full names rather than abbreviations. `ifixedpt` -> `fixed_point_iteration`

amr-wind/incflo.H
```diff
@@ -172,6 +174,9 @@ private:
     // Prescribe advection velocity
     bool m_prescribe_vel = false;
 
+    // Fixed point iterations every timestep
+    int m_fixed_pt_iters{1};
```
marchdf, line 178: `m_fixed_pt_iters` -> `m_fixed_point_iterations`

amr-wind/incflo.H
```diff
@@ -96,8 +96,8 @@ public:
     bool regrid_and_update();
     void pre_advance_stage1();
     void pre_advance_stage2();
-    void do_advance();
-    void advance();
+    void do_advance(const int ifixed_point_iteration);
```
marchdf, line 99: I think we can just do `fixed_point_iteration` (remove the `i`, it's not really a convention in amr-wind)

amr-wind/incflo.cpp
```diff
@@ -372,6 +377,12 @@ void incflo::init_physics_and_pde()
         if (activate_overset) {
             m_sim.activate_overset();
         }
+
+        // Get number of advection iterations
```
marchdf, line 381: remove comment

amr-wind/incflo_advance.cpp
```diff
@@ -47,15 +47,20 @@ void incflo::pre_advance_stage2()
...
 
-    m_sim.pde_manager().advance_states();
-
-    ApplyPredictor();
+    ApplyPredictor(false, ifixed_point_iteration);
 
     if (!m_use_godunov) {
+        AMREX_ALWAYS_ASSERT_WITH_MESSAGE(
+            ifixed_point_iteration == 0,
+            "Advection iterations are not supported for MOL");
```
marchdf, line 62: The error message needs to be adjusted now that we call this fixed point iterations

amr-wind/incflo.H
```diff
@@ -133,7 +133,9 @@ public:
     void ComputeDt(bool explicit_diffusion);
     void ComputePrescribeDt();
 
-    void ApplyPredictor(bool incremental_projection = false);
+    void ApplyPredictor(
+        bool incremental_projection = false,
```
marchdf, line 137: while you are at it, `const` the first arg as well.

amr-wind/incflo.H
```diff
@@ -96,8 +96,8 @@ public:
     bool regrid_and_update();
     void pre_advance_stage1();
     void pre_advance_stage2();
-    void do_advance();
-    void advance();
+    void do_advance(const int ifixed_point_iteration);
+    void advance(int inonlin);
```
marchdf, line 100: rename `inonlin`

marchdf:
All tests pass! (well except the ones irrelevant to this PR for which I created issues).

marchdf (approved):

PR #1243, opened 2024-09-16
Fix interface_band

mbkuhn:
## Summary

Looks like the interface_band was turning on in all liquid. That's a problem. Unit test didn't cover that, now it does.

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [x] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [x] on CPU 
- the regression tests
  - [ ] on GPU 
  - [x] on CPU 

## Additional background

- correct check against tiny
- allow tiny to be an input argument
- update unit test to cover previous failure

I'll run the reg tests, it will probably affect many of the multiphase tests.

### DIFF
--- a/amr-wind/equation_systems/vof/volume_fractions.H
+++ b/amr-wind/equation_systems/vof/volume_fractions.H
@@ -493,11 +493,10 @@ AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE bool interface_band(
     const int j,
     const int k,
     amrex::Array4<amrex::Real const> const& volfrac,
-    const int n_band = 1,
-    const amrex::Real tiny = 1e-12) noexcept
+    const int n_band = 1) noexcept
 {
     // n_band must be <= number of vof ghost cells (3)
-
+    const amrex::Real tiny = 1e-12;
     // Check if near interface
     amrex::Real VOF_max = 0.0;
     amrex::Real VOF_min = 1.0;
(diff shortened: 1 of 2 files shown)


mbkuhn:
This PR causes diffs in

 - abl_multiphase_laminar
 - dam_break_godunov
 - vortex_patch_scalar_vel
 - rain_drop
 - inertial_drop
 - ow_linear
 - ow_stokes
 - ow_w2a

These are expected, and they are minor. The changes in this PR should make everything more accurate.
I will run a handful of these longer to ensure stability.

amr-wind/equation_systems/vof/volume_fractions.H
```diff
@@ -493,10 +493,11 @@ AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE bool interface_band(
     const int j,
     const int k,
     amrex::Array4<amrex::Real const> const& volfrac,
-    const int n_band = 1) noexcept
+    const int n_band = 1,
+    const amrex::Real tiny = 1e-12) noexcept
 {
     // n_band must be <= number of vof ghost cells (3)
```
marchdf, line 499: Just some small questions: Do you want to make this check explicit in the places where you call this function? Also, I don't see you using `tiny` as an argument so do you want to consider leaving it as an variable inside this function as it was?

mbkuhn: good questions. I think it would be a lot of scattered checks if I did make it explicit, and I don't think I have enough information inside the function to do the check. I don't think it's a big enough concern to add checks.

I found this bug while I was developing something else, which was using tiny as an argument. I'll remove that from this PR and add it if I use it in another PR.

marchdf (approved):

mbkuhn:
I ran the inertial_drop and vortex_patch_scalar_vel cases for much longer and looked at the results. I also ran a stokes waves case for a while. Looks good enough to merge, no apparent issues. Generally higher velocities, but not by much, which is expected.

PR #1244, opened 2024-09-16
Correct parsing logic for ABLMeanBoussinesq

### LINKED ISSUE #1237: Contradictory input file behavior possible for ABLMeanBoussinesq
## Bug description
In ABLMeanBoussinesq.cpp, the parser checks for the existence of "read_temperature_profile" and never uses the value of it. This means that the argument "read_temperature_profile" could be set to false but have the effect of being set to true.

https://github.com/Exawind/amr-wind/blob/main/amr-wind/equation_systems/icns/source_terms/ABLMeanBoussinesq.cpp#L39

Either the parser should read the value of "read_temperature_profile" or that argument should be deprecated and the parser should just look for the existence of "tprofile_filename".

This bug was found by looking at the code, not during a run.

mbkuhn:
## Summary

Previously, the read_temperature_profile input could be specified to "false" but would still be used. This is replaced with the following logic:

If read_temperature_profile is not in the input file, but tprofile_name is, then the temperature profile will be used. Alternatively, the temperature profile will not be used unless read_temperature_profile is present and true.

Any oppositions to this new approach?

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [ ] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU 

## Additional background

Need to run reg tests with ABLMeanBoussinesq

Issue Number: closes #1237

### DIFF
--- a/amr-wind/equation_systems/icns/source_terms/ABLMeanBoussinesq.cpp
+++ b/amr-wind/equation_systems/icns/source_terms/ABLMeanBoussinesq.cpp
@@ -36,7 +36,12 @@ ABLMeanBoussinesq::ABLMeanBoussinesq(const CFDSim& sim) : m_mesh(sim.mesh())
     amrex::ParmParse pp_incflo("incflo");
     pp_incflo.queryarr("gravity", m_gravity);
 
-    if (pp_boussinesq_buoyancy.contains("read_temperature_profile")) {
+    bool read_temp_prof = false;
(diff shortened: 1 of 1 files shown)


marchdf:
Tests pass.

marchdf (approved):

PR #1254, opened 2024-09-19
Change to multiphase hybrid coupling - both projections

mbkuhn:
## Summary

Previously, spurious currents in multiphase hybrid cases were reduced by not coupling the nodal projection directly to the incoming pressure values. We have found issues with that, and the proposed change is coupling single-phase regions directly and loosely coupling around the interface.

We also made the same change for the MAC projection coupling, which means this PR enables a single approach that is compatible for both single-phase and multiphase overset needs.

## Pull request type

- [x] Bugfix
- [x] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [x] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [x] on CPU 
- the regression tests
  - [ ] on GPU 
  - [x] on CPU 

## Additional background

Changes to regression tests of exawind-driver will go along with this, but they run fine with a few tweaks.


amr-wind/equation_systems/vof/volume_fractions.H
```diff
@@ -493,10 +493,11 @@ AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE bool interface_band(
     const int j,
     const int k,
     amrex::Array4<amrex::Real const> const& volfrac,
-    const int n_band = 1) noexcept
+    const int n_band = 1,
+    const amrex::Real tiny = 1e-12) noexcept
```
marchdf, line 497: default to constants:TIGHT_TOL?

marchdf (approved):

PR #1264, opened 2024-09-27
Improvements to wave generation - ocean waves

mbkuhn:
## Summary

I found that some cases are not stable, with problems arising from the wave generation region. This PR modifies the forcing in 2 ways to make it more reasonable and more robust.

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [x] on CPU 
- the regression tests
  - [ ] on GPU 
  - [x] on CPU 

## Additional background

Creates diffs for ow_* regression tests, which are expected

### DIFF
--- a/amr-wind/ocean_waves/relaxation_zones/RelaxationZones.H
+++ b/amr-wind/ocean_waves/relaxation_zones/RelaxationZones.H
@@ -34,7 +34,7 @@ struct RelaxZonesBaseData
     bool has_beach{true};
     bool has_outprofile{false};
 
-    amrex::Real ramp_period{2.0};
+    amrex::Real ramp_period;
 
     bool regrid_occurred{false};
 };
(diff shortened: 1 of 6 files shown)


mbkuhn:
This PR also has some refactoring (introducing loops over the three components of velocity) and a bug fix (making the ramp function work as intended).

rybchuk:
Looking forward to this! Thanks for putting this together.

mbkuhn:
@rybchuk have you had stability issues within the forcing regions?

rybchuk:
I've had simulations where the CFL kept blowing up after a while. I tried to get those simulations to stabilize, but never was able to. My workaround was dialing back the wave amplitude from 6.4 m (as well as 3.2 m) to 1.6 m. I figured that my code was crashing because of wave physics (are they allowed to get that big?) instead of something tied to numerics

rybchuk:
I couldn't tell you if these simulations were crashing because of behavior in the wave input region, wave dissipation region, or elsewhere though

mbkuhn:
Gotcha. It's very possible it's related to this.

mbkuhn:
@marchdf Aside from the small tweaks and refactoring of things, this PR has two modifications to how the wave forcing is applied:
1. When liquid is added to a cell, assume it has the target velocity, not the current velocity of the cell
2. Avoid forcing the velocity to a target velocity when the target velocity is invalid (because the current cell is not within the target wave).

Also, I've done several things to check this manually.

marchdf (approved):

marchdf:
Sounds good! I can't run any tests right now with the downtime. We can merge and deal with the consequences or wait it out. Up to you.

PR #1274, opened 2024-10-07
Set up fill_boundaries as a step prior to ApplyPredictor

mbkuhn:
## Summary

In order to accommodate the boundary plane implementation and correctly fill physical boundary cells at the right times, this PR sets up a framework to call "fill_boundaries" prior to the ApplyPredictor step, which enables avoiding calling fills at older times (n, n+1/2) once the time step has begun. With all of this new code, this PR is set up to avoid diffs, ensuring the framework does not do something unintentionally. Future PR will introduce diffs by filling BCs at the desired times, instead of conforming to the times the code currently uses.

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [x] on CPU 
- the regression tests
  - [ ] on GPU 
  - [x] on CPU 

## Additional background

The "no diffs" component of development intended by #1210.


amr-wind/wind_energy/ABLFillInflow.cpp
```diff
@@ -63,7 +66,8 @@ void ABLFillInflow::fillpatch_sibling_fields(
     const amrex::Vector<amrex::BCRec>& bcrec,
     const FieldState fstate)
 {
-    bool plane_data_unchanged = m_bndry_plane.tinterp() - time > 1e-12;
+    // Avoid trying to read after planes have already been populated
+    bool plane_data_unchanged = m_bndry_plane.is_data_newer_than(time);
```
marchdf, line 70:
```suggestion
    const bool plane_data_unchanged = m_bndry_plane.is_data_newer_than(time);
```

marchdf:
I am seeing this:
```
The following tests FAILED:
          7 - abl_godunov_mpl (Failed)
         87 - rankine (Failed)
         88 - rankine-sym (Failed)
         92 - freestream_godunov_inout (Failed)
        123 - abl_bndry_input_amr_native_xhi (Failed)
        124 - abl_bndry_input_amr_native_mlbc (Failed)
        128 - abl_bndry_input_amr_inflow (Failed)
```
I think we might expect the MPL ones? What about the others?

marchdf:
We are at:
```
The following tests FAILED:
          7 - abl_godunov_mpl (Failed)
```

which is good news. Let's do this.

marchdf (approved):

PR #1278, opened 2024-10-09
Remove unnecessary warning for ABL inflow planes.

lawrenceccheung:
## Summary

Currently when boundary plane faces are used on the `xhi` or `yhi` faces, amr-wind outputs the following warning:
```
We typically don't inflow boundary planes on the high side.!
```
However, depending on the wind direction, boundary planes can be on the `xhi` or `yhi` faces, and AMR-Wind should be capable of handling them.

## Pull request type

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [X] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU 

## Additional background

Issue Number:

### DIFF
--- a/amr-wind/wind_energy/ABLBoundaryPlane.cpp
+++ b/amr-wind/wind_energy/ABLBoundaryPlane.cpp
@@ -902,11 +902,6 @@ void ABLBoundaryPlane::populate_data(
             amrex::Abort("No inflow data at this level.");
         }
 
-        if (ori.isHigh()) {
-            amrex::Warning(
-                "We typically don't inflow boundary planes on the high side.");
-        }
-
         const size_t nc = mfab.nComp();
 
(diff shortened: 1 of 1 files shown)


marchdf:
Thanks for looking at this! I agree that it _should_ be able to handle it and that it _should_ work. At the time, I wasn't able to confirm it. Were you able to confirm that it _can_ actually do it on the high end? That would be fantastic.

lawrenceccheung:
Yes, I tested this on a case with a combination of ylo+xhi boundary inflow surfaces, and it seemed to work well.  Note: this case was also only using a level 0 grid.

marchdf (approved):

PR #1283, opened 2024-10-10
Ocean Waves: give added liquid target velocity in numerical beach

mbkuhn:
## Summary

Same idea as #1264, but for the numerical beach instead of the wave generation zone

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [ ] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU 

## Additional background

Likely no diffs on ow_* regression tests because they don't run long enough to hit numerical beach. I'll add a test to change that.


mbkuhn:
there are very small diffs with ow_linear and ow_stokes, as expected. new regression test is set up to better test numerical beach functionality by initializing with a full wave field, which is present in the outlet region from the very beginning.

mpolimeno (approved):

marchdf (approved):

PR #1286, opened 2024-10-14
Add dry run capability

mbkuhn:
## Summary

Had a request for a dry run capability so the setup could be viewed without running a simulation. This is a sort-of automated way of doing the least amount of work while still populating arrays and outputting a plotfile. I also made sure that a plotfile will still be written when doing a dry run for a restart.

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [ ] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU 

## Additional background

Before merging, I'll need to add this to the documentation. I could incorporate it into it's own reg test, but I'm not sure it's worth the hassle.

### DIFF
--- a/amr-wind/setup/init.cpp
+++ b/amr-wind/setup/init.cpp
@@ -29,7 +29,7 @@ void incflo::CheckAndSetUpDryRun()
     // Zero time steps, write plotfile and not checkpoint
     {
         ParmParse pp("time");
-        pp.add("max_step",(int)0);
+        pp.add("max_step", (int)0);
         pp.add("plot_interval", (int)1);
         pp.add("checkpoint_inteval", (int)-1);
     }
--- a/amr-wind/setup/init.cpp
+++ b/amr-wind/setup/init.cpp
@@ -10,6 +10,38 @@
 
 using namespace amrex;
 
+void incflo::CheckAndSetUpDryRun()
+{
+    // Check if dry run is requested; exit if not
+    {
+        ParmParse pp("incflo");
+        pp.query("dry_run", m_dry_run);
+        if (!m_dry_run) {
+            return;
+        }
+    }
+    // Disable additional computations associated with initialization
+    {
+        ParmParse pp("incflo");
+        pp.add("initial_iterations", (int)0);
+        pp.add("do_initial_proj", (bool)false);
+    }
+    // Zero time steps, write plotfile and not checkpoint
+    {
+        ParmParse pp("time");
+        pp.add("max_step",(int)0);
+        pp.add("plot_interval", (int)1);
+        pp.add("checkpoint_inteval", (int)-1);
+    }
+    // Give prefix to plotfile
+    {
+        ParmParse pp("io");
+        std::string current_plt{"plt"};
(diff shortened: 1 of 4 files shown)


mbkuhn:
It would be good to chat with @hgopalan, @rthedin, and @MaxPal to see if this would be useful for their workflows.

hgopalan:
I think if it is for the complex terrain support, we need to read the TerrainDrag module which initializes the terrain_blank field and provides information on the terrain mapping before running simulations. I register the velocity field to create amrex::MFIter mfi(velocity); mfi.isValid(); ++mfi since I did not know how to do it otherwise with all new variables. If there is a way to avoid calling the velocity field, I can update the TerrainDrag and we would need only to register the new variables and plot them.

amr-wind/setup/init.cpp
```diff
@@ -10,6 +10,38 @@
 
 using namespace amrex;
 
+void incflo::CheckAndSetUpDryRun()
```
marchdf, line 13: hmmm I am not sure the order of operations in the parmparse. Could you check with, e.g. having 
```
incflo.initial_iterations = 3
incflo.do_initial_proj =1
incflo.dry_run =1
```
in the input file? Do those values get reset by this function?

marchdf:
I don't think we need a regtest for this. But documentation would be good ;)

mbkuhn:
> I think if it is for the complex terrain support, we need to read the TerrainDrag module which initializes the terrain_blank field and provides information on the terrain mapping before running simulations. I register the velocity field to create amrex::MFIter mfi(velocity); mfi.isValid(); ++mfi since I did not know how to do it otherwise with all new variables. If there is a way to avoid calling the velocity field, I can update the TerrainDrag and we would need only to register the new variables and plot them.

Thanks for sharing. For the sake of the terrain, there probably could be a standalone tool devised to address that need, which would be nimbler in terms of memory needs. This approach is more general and should help others who would appreciate the capability with different problems. One nice thing with this approach is it uses the actual code path of AMR-Wind, so it should run into the same problems that a real run would (e.g., if something is not set up properly in the input).

amr-wind/setup/init.cpp
mbkuhn: I checked by doing dry runs with abl_godunov. Those values get reset with this function. Regardless of the arguments in the input file or the command line, the initial iterations get set to 0 and the initial projection gets skipped. That's the intention. (If someone wanted those to happen, it's still possible to do the same thing, just not using the dry run option)

mbkuhn: I also made sure that it works with abl_bndry_input_native; plotfile written successfully.

marchdf: awesome. Thanks!

marchdf (approved):

moprak-nrel (approved):

PR #1287, opened 2024-10-14
Make clipping true default for pressure_outflow and deprecate pressure_inflow

mbkuhn:
## Summary

Make default behavior of pressure_outflow to disallow inflow; users can turn on inflow at a pressure_outflow if desired. Removing the pressure_inflow option because 1) it is not used and 2) it is not set up to be stable.

## Pull request type

- [x] Bugfix
- [x] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [ ] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU 

## Additional background

We would like to be using the pressure_outflow condition in some wave simulations, but unless there is a strong outward flow, this is unstable without clipping. We need more flexibility either way, and it makes sense to improve the defaults now.


mbkuhn:
pressure_inflow capability still remains by using pressure_outlfow and turning on inflow at outflow.

hgopalan:
Was there any specific case in which there was a reverse flow at the outlet? You can add a sponge layer if you want a stronger outflow boundary condition.

mbkuhn:
> Was there any specific case in which there was a reverse flow at the outlet? You can add a sponge layer if you want a stronger outflow boundary condition.

The cases need an intermittent outlet that isn't following a prescribed solution. I think the clipping will do the trick.

marchdf (approved):

mbkuhn:
this should create small diffs for pressure outflow cases that use weno_z. that is ok

PR #1298, opened 2024-10-16
Correctly populate mac vel ghost cells at high boundaries in auxiliary fills

### LINKED ISSUE #1296: Usage of enclosedCells() is not valid for hi boundaries
## Bug description
A few places in the code (affiliated with filling boundary cells), the "enclosedCells" call is used to convert face-centered boxes to cell-centered boxes. However, this is not a legitimate way to do it because it is only valid for lo boundaries. The orientation (hi/lo) needs to be considered to be valid for both.

https://github.com/Exawind/amr-wind/blob/main/amr-wind/wind_energy/ABLBoundaryPlane.cpp#L915
https://github.com/Exawind/amr-wind/blob/main/amr-wind/wind_energy/ABLModulatedPowerLaw.cpp#L178
this one doesn't matter because we don't have any temperature fields located at faces, but it would be good to correct it as well.
https://github.com/Exawind/amr-wind/blob/main/amr-wind/wind_energy/ABLModulatedPowerLaw.cpp#L255

## Steps to reproduce
I added a print statement to output the i index when the populate_data call is executed for filling umac.
when running the "abl_bndry_input_amr_native_xhi" reg test, the print statement shows that the index 48 is being written to, even though this is a valid face in the domain, not a ghost face, as the routine intends.

1. Compiler used
   - [ ] GCC
   - [x] LLVM
   - [ ] oneapi (Intel)
   - [ ] nvcc (NVIDIA)
   - [ ] rocm (AMD)
   - [ ] with MPI
   - [ ] other:
2. Operating system
   - [ ] Linux
   - [x] OSX
   - [ ] Windows
   - [ ] other (do tell ;)):
3. Hardware:
   - [x] CPU
   - [ ] GPU

## Additional context
This is likely why there are diffs for this reg test for the PR #1274

mbkuhn:
## Summary

MPL and boundary plane fills did not get the correct indices to fill when operating on the high side. This will fix that. Should only introduce diffs to the _xhi reg test.

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [x] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [x] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU 

## Additional background

This doesn't make everything perfect when using cell-centered plane data to fill ghost cells at the face. It does not address how there is an offset in position (e.g., filling umac on a y-boundary using cell-centered u velocity).

Issue Number: closes #1296


marchdf:
Only 1 failed test:
```
123 - abl_bndry_input_amr_native_xhi (Failed)
```
I think this one is expected.

marchdf (approved):
CodeQL is not happy but I think you can easily fix that.

unit_tests/boundary_conditions/test_auxiliary_fill.cpp
```diff
@@ -64,16 +64,23 @@ amrex::Real get_field_err(
...
                         // Checking u
                         if (check_all_ghost) {
                             // Ghost cells in boundaries, but not corners
-                            if (((i == -1 || i == 9) && (j >= 0 && j <= 7) &&
-                                 (k >= 0 && k <= 7)) ||
-                                ((j == -1 || j == 8) && (i >= 0 && i <= 7) &&
-                                 (k >= 0 && k <= 7)) ||
-                                ((k == -1 || k == 8) && (j >= 0 && j <= 7) &&
-                                 (i >= 0 && i <= 7))) {
+                            bool in_xf_bdy = (i == -1 || i == 9);
```
marchdf, line 80: const these ones?

PR #1302, opened 2024-10-21
Perform prepare_boundaries at correct times

mbkuhn:
## Summary

Follow on to #1274 and fulfilling intention of #1210. Two changes:

1. fillphysbc calls at n+1/2 time to put time-correct values into nph variable states, which are already prepared to be used in advection. Without these calls (as it is after the no-diffs PR #1274), the nph states are just the same as the old (n) states.
2. convert the usage of m_intended_interp_time to actually calculate the interpolation time. This is to interpolate the plane data at n+1/2 prior to the timestep and then at n+1 at the beginning of the timestep.

## Pull request type

Please check the type of change introduced:

- [x] Bugfix
- [x] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe):

## Checklist

The following is included:

- [ ] new unit-test(s)
- [ ] new regression test(s)
- [ ] documentation for new capability

This PR was tested by running:

- the unit tests
  - [ ] on GPU 
  - [ ] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU 

## Additional background

Expected diffs are in the time-dependent BC cases (freestream_inout, rankine*) and the abl boundary input cases (abl_bndry_input*, abl_godunov_forcetimetable).

PR #1303, opened 2024-10-21
in apply_bcs, only perform fillphysbc for new time

mbkuhn:
## Summary

Avoid filling physbc at old times, they should already be filled. Necessary for #1302.

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [ ] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU 

## Additional background

Should introduce no diffs, but may introduce diffs for abl_godunov_wenoz_fixedpt. However, this part of the algorithm needs to be adjusted for fixed point iterations, which I've talked to Ilker about.


mbkuhn:
The reason why this needs to be adjusted is the compute_diff_term always operates on the new state, even when the values are old. This isn't necessary; we should be able to operate on the old state, and that would fix things.

mbkuhn:
I'm thinking I'll add that tweak to this PR

marchdf:
I am seeing these fail:
```
The following tests FAILED:
          7 - abl_godunov_mpl (Failed)
         60 - abl_godunov_wenoz_fixedpt (Failed)
        121 - abl_bndry_input_native (Failed)
        123 - abl_bndry_input_amr_native (Failed)
        124 - abl_bndry_input_amr_native_xhi (Failed)
        125 - abl_bndry_input_amr_native_mlbc (Failed)
        126 - abl_godunov_forcetimetable (Failed)
        127 - abl_bndry_input (Failed)
        128 - abl_bndry_input_amr (Failed)
        129 - abl_bndry_input_amr_inflow (Failed)
        130 - nrel_terrain (Failed)
```
The fixedpt one is just missing golds. Not sure about the others.

mbkuhn:
I think these all make sense, actually. The boundary planes are all currently hard-coded to use new_time when there is a fillphysbc. Therefore, when apply_bcs takes place at the old time, it puts new data in to the boundaries of the new state, which hasn't been updated yet, so it is just old data with new boundaries. This PR takes that away.

Ok, well, one thing that pokes a hole in this is that many of these cases have incflo.diffusion = 2, which means it's pure implicit. Not sure what compute_diff_term is doing in that case, gonna have to investigate further.

mbkuhn:
Ok, I stand by my original theory. Even though many of these use implicit diffusion, the mac projection still uses the explicit diffusion calculation. The diffs only show up where the boundaries are coded to update (even when they shouldn't, at old times).

mbkuhn:
Supporting evidence:
if I change the state to "New" instead of "Old" in this line, then the diffs go away. The changes only prep diffusion terms for the MAC extrapolation step.
https://github.com/Exawind/amr-wind/blob/main/amr-wind/incflo_advance.cpp#L250

PR #1321, opened 2024-10-31
Integrate inflow-outflow BC into ABL

mukul1992:
## Summary

These changes allow the use of the `mass_inflow_outflow` boundary condition with the ABL physics simulations that involve reading boundary planes. It also adds a new test to test this capability and updates the older inflow-outflow test to involve flow in both x- and y-directions. Finally, updating AMReX-Hydro solves a [bug](https://github.com/AMReX-Fluids/AMReX-Hydro/pull/153) in the routines that enforce solvability by scaling the outflow to match the inflow.

The new capability was verified by a qualitative comparison of the new `abl_bndry_input_native_inout` test with the older `abl_bndry_input_native` test which uses the simple `mass_inflow` and `pressure_outflow` BCs.

## Pull request type

- [x] Bugfix
- [x] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [x] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [x] on CPU (Apple M3, macOS, gcc-13)
- the regression tests
  - [ ] on GPU 
  - [x] on CPU (Apple M3, macOS, gcc-13)


amr-wind/boundary_conditions/MassInflowOutflowBC.cpp
```diff
@@ -72,4 +72,4 @@ void MassInflowOutflowBC::operator()(
     }
 }
 
-} // namespace amr_wind
\ No newline at end of file
+} // namespace amr_wind
```
marchdf, line 75: spurious commit I think

marchdf (commented):
Going to run on the test machine. Otherwise looks good!

amr-wind/boundary_conditions/MassInflowOutflowBC.cpp
mukul1992: From what I understand, it is good practice to have a newline at the end of file. So I turned on automatically adding a newline in VScode and kept this change. Let me know if that is not correct and I can undo it.

marchdf:
The only diff I see on CPU is `freestream_godunov_inout`. Running the GPU tests now.

marchdf:
Closes #1319

mukul1992:
> The only diff I see on CPU is `freestream_godunov_inout`. Running the GPU tests now.

Yes, those are expected as I changed the test to involve flow along both x and y instead of just in x.

marchdf (approved):

marchdf:
Everything looked fine on GPU, not sure why docker is failing. I think it is irrelevant to this PR.

PR #1323, opened 2024-11-01
Use linear interpolation in DragForcing

marchdf:
## Summary

This is much more efficient than the current implementation since it does a bisection search. Also instead of a nearest neighbor it uses a linear interpolation. One can recover the previous implementation by just using the new `nearest_search`. 

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [x] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [x] on CPU 
- the regression tests
  - [ ] on GPU 
  - [x] on CPU

### DIFF
--- a/amr-wind/equation_systems/icns/source_terms/DragForcing.cpp
+++ b/amr-wind/equation_systems/icns/source_terms/DragForcing.cpp
@@ -4,6 +4,7 @@
 #include "AMReX_Gpu.H"
 #include "AMReX_Random.H"
 #include "amr-wind/wind_energy/ABL.H"
+#include "amr-wind/utilities/linear_interpolation.H"
 
 namespace amr_wind::pde::icns {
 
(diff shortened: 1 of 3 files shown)


hgopalan:
Here are some timing tests for a case with ~ 13 million cells on 208 cores 

Older nearest search: Time spent in Evolve():      2914.012737
New linear interpolation: Time spent in Evolve():      2579.18432

hgopalan (approved):

marchdf:
@hgopalan Thanks for doing that comparison. I thought of a better/less complicated way. Would you mind checking the timing with the latest?

moprak-nrel (approved):

moprak-nrel (approved):

marchdf:
1.25X speedup. I will take it ;)

PR #1329, opened 2024-11-05
bdy plane interp time: match expressions to pass assertion

mbkuhn:
## Summary

Due to recent changes to the boundary plane times being used, this assertion can sometimes get false negatives. This is because the calculation of n+1/2 time is done two different ways, leading to floating point differences. This PR changes the expressions that calculate n+1/2 time to be the same (current_time + 0.5*dt).

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [ ] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU 

## Additional background

Before the overhaul of the boundary planes, which introduced getting planes at n+1/2 and at n+1, it made sense for this assertion to be very tight, because the expectation was that the variables would be identical. new_time is saved as m_tinterp when the data is interpolated, then when new_time is passed in later, it should be no different. However, in the new implementation, the n+1/2 time is calculated initially using new_time + 0.5 * (current_time - new_time), which gets stored as m_tinterp, and then later that gets compared with current_time + 0.5 * dt. Apparently, these calculations are sufficiently different to occasionally trip the assertion (this is what @hgopalan has seen on a few different runs). By changing the expressions to be the same, the assertion shouldn't get tripped.

### DIFF
--- a/amr-wind/wind_energy/ABLBoundaryPlane.cpp
+++ b/amr-wind/wind_energy/ABLBoundaryPlane.cpp
@@ -893,7 +893,7 @@ void ABLBoundaryPlane::populate_data(
 
     AMREX_ALWAYS_ASSERT(
         ((m_in_data.tn() <= time) || (time <= m_in_data.tnp1())));
-    AMREX_ALWAYS_ASSERT(std::abs(time - m_in_data.tinterp()) < 1e-12);
+    AMREX_ALWAYS_ASSERT(std::abs(time - m_in_data.tinterp()) < 1e-6);
 
     for (amrex::OrientationIter oit; oit != nullptr; ++oit) {
         auto ori = oit();
--- a/amr-wind/wind_energy/ABLBoundaryPlane.cpp
+++ b/amr-wind/wind_energy/ABLBoundaryPlane.cpp
@@ -770,8 +770,8 @@ void ABLBoundaryPlane::read_file(const bool nph_target_time)
 
     // populate planes and interpolate
     const amrex::Real time =
-        m_time.new_time() + (nph_target_time ? 0.5 : 0.0) *
-                                (m_time.current_time() - m_time.new_time());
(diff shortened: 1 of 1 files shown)


marchdf:
```
The following tests FAILED:
         75 - ow_linear_init_waves (Failed)
        122 - abl_bndry_input_native (Failed)
        123 - abl_bndry_input_native_inout (Failed)
        125 - abl_bndry_input_amr_native (Failed)
        126 - abl_bndry_input_amr_native_xhi (Failed)
        127 - abl_bndry_input_amr_native_mlbc (Failed)
        128 - abl_godunov_forcetimetable (Failed)
        129 - abl_bndry_input (Failed)
        130 - abl_bndry_input_amr (Failed)
        131 - abl_bndry_input_amr_inflow (Failed)
```

But I think they are small diffs. e.g `abl_bndry_input_native_inout` 
```
 level = 0
 density                                          0                         0
 gpx                                3.078266808e-15           1.692866604e-13
 gpy                                1.202163369e-15           5.537836584e-14
 gpz                                 1.13776176e-15           1.261724108e-13
 mu_turb                            2.164934898e-15           5.454117097e-15
 p                                   6.87505608e-14            5.72463436e-14
 temperature                        1.136868377e-13           3.920235784e-16
 temperature_mueff                  6.439293543e-15           5.406895103e-15
 velocityx                          3.552713679e-15            3.23575245e-16
 velocityy                          6.106226635e-16           6.164154056e-16
 velocityz                           5.26488575e-16           1.348645256e-14
 velocity_mueff                     2.164934898e-15           5.453979694e-15
```
Looking through the others.

marchdf:
Ok yeah, all tiny diffs.

marchdf (approved):

Issue #1343, opened 2024-11-11
Plane Sampling complains of out of bounds plane


marchdf:
## Bug description
This:
```
geometry.prob_lo        =   0.       0.     0.  # Lo corner coordinates
geometry.prob_hi        =   1000.  1000.  1000.  # Hi corner coordinatesplane_sampling.plane2.type          = PlaneSampler
plane_sampling.plane2.axis1         = 1000.0 0.0 0.0
plane_sampling.plane2.axis2         = 0.0 1000.0 0.0
plane_sampling.plane2.origin        = 0.0 0.0 500.0
plane_sampling.plane2.num_points    = 10 10
```

will lead to this:
```
libc++abi: terminating due to uncaught exception of type std::runtime_error: PlaneSampler: Point out of domain. Redefine your planes so they are completely inside the domain.
```

Which I knew about when I coded up the check but I thought it would be less annoying than it turned out to be. The problem I faced is that correcting user input of planes that go out of bounds is very non-trivial (because of the offsets and offset vectors, planes could go anywhere and I couldn't figure out how to correct this). My thinking now is the provide a "base" correction that would partially tweak the user inputs so that things work but won't be expected to fix it all.

PR #1344, opened 2024-11-11
Fix plane bounds checks

marchdf:
## Summary

If the user wants to specify planes that reach the domain bounds, adjust the `origin`, `axis1` and `axis2` so that the points are within the domain. This isn't a completely fullproof adjustment of the user values but I hope it catches most of them.

The check is based on a small fraction of the cell size so it won't really change the QOIs.

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [x] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [x] on CPU 
- the regression tests
  - [ ] on GPU 
  - [x] on CPU 

## Additional background

Closes: #1343

### DIFF
--- a/amr-wind/utilities/sampling/PlaneSampler.cpp
+++ b/amr-wind/utilities/sampling/PlaneSampler.cpp
@@ -58,6 +58,40 @@ void PlaneSampler::check_bounds()
     const auto* prob_lo = m_sim.mesh().Geom(lev).ProbLo();
     const auto* prob_hi = m_sim.mesh().Geom(lev).ProbHi();
 
+    // Do the checks based on a small fraction of the minimum dx
+    const auto& fine_geom = m_sim.mesh().Geom(m_sim.mesh().finestLevel());
+    amrex::Real min_dx = fine_geom.CellSize(0);
+    for (int d = 1; d < AMREX_SPACEDIM; ++d) {
+        min_dx = std::min(fine_geom.CellSize(d), min_dx);
+    }
+    const auto tol = std::max(1e-10 * min_dx, bounds_tol);
+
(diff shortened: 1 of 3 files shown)


amr-wind/utilities/sampling/PlaneSampler.cpp
```diff
@@ -58,6 +58,40 @@ void PlaneSampler::check_bounds()
...
+    amrex::Real min_dx = fine_geom.CellSize(0);
+    for (int d = 1; d < AMREX_SPACEDIM; ++d) {
+        min_dx = std::min(fine_geom.CellSize(d), min_dx);
+    }
+    const auto tol = std::max(1e-10 * min_dx, bounds_tol);
+
+    // First fix the origin so that it is within bounds, if it is close enough
+    for (int d = 0; d < AMREX_SPACEDIM; ++d) {
+        if (amrex::Math::abs(m_origin[d] - prob_lo[d]) < tol) {
+            m_origin[d] = prob_lo[d] + tol;
```
moprak-nrel, line 72: Curious if we want `(1+tol) * prob_lo` and `(1-tol) * prob_hi` instead of `+-tol` directly. 

(Given the tests, I'm guessing not, but if we get to `numeric_limits::epsilon` instead of `1e-10*dx`, the `tol` wouldn't add anything right?)

moprak-nrel: Oh nvm, just saw what the actual `if` condition is checking.

moprak-nrel (approved):

PR #1346, opened 2024-11-14
Update checkpoint refinement to have a max level

marchdf:
## Summary

This lets the user do `amr.max_refined_level=1`:
```
Input grid summary:
  Level 0   216 grids  110592 cells  100 % of domain
            smallest grid: 8 x 8 x 8  biggest grid: 8 x 8 x 8
  Level 1   27 grids  13824 cells  1.5625 % of domain
            smallest grid: 8 x 8 x 8  biggest grid: 8 x 8 x 8

Uniformly refining mesh
Refinement time elapsed: 1.657051958
Refined grid summary:
  Level 0   216 grids  110592 cells  100 % of domain
            smallest grid: 8 x 8 x 8  biggest grid: 8 x 8 x 8
  Level 1   1728 grids  884736 cells  100 % of domain
            smallest grid: 8 x 8 x 8  biggest grid: 8 x 8 x 8
  Level 2   512 grids  262144 cells  3.703703704 % of domain
            smallest grid: 8 x 8 x 8  biggest grid: 8 x 8 x 8

Writing refined levels: 1 - 1
```

So that only (in this example) level 1 is written out. By default it would write level 1 and 2.

### DIFF
--- a/tools/utilities/refine-chkpt/RefineCheckpt.cpp
+++ b/tools/utilities/refine-chkpt/RefineCheckpt.cpp
@@ -66,9 +66,14 @@ void RefineCheckpt::run_utility()
     } else {
         refine_chkpt_file();
         const int start_level = 1;
+        int end_level = finestLevel();
+        {
+            amrex::ParmParse pp("amr");
+            pp.query("max_refined_level", end_level);
+        }
         amrex::Print() << "Writing refined levels: " << start_level << " - "
-                       << finestLevel() << std::endl;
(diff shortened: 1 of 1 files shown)


moprak-nrel:
Worth adding this to the input file documentation?

moprak-nrel (approved):

marchdf:
If only this tool was documented ;) I think we should add this to our todo list. These are auxiliary tools that we should document someday.

PR #1347, opened 2024-11-15
Allow string periodic for periodic BCs

marchdf:
## Summary

Allow string periodic for periodic BCs.

Basically it's weird to get an error message that says:
```
std::runtime_error: Setting periodic BC on a periodic face xlo is not allowed
```

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [ ] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU

### DIFF
--- a/amr-wind/boundary_conditions/BCInterface.cpp
+++ b/amr-wind/boundary_conditions/BCInterface.cpp
@@ -47,7 +47,8 @@ void BCIface::read_bctype()
 
         // Protect against copy/paste errors where user intended to add a BC but
         // forgot to turn off periodic in that direction, or vice versa.
-        if (geom.isPeriodic(ori.coordDir()) && bcstr != "null") {
+        if (geom.isPeriodic(ori.coordDir()) &&
+            ((bcstr != "null") && (bcstr != "periodic"))) {
             amrex::Abort(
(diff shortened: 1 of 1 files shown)


mbkuhn:
so, the geom.is_periodic still has to be correct, but this now allows the user to put a periodic argument for that boundary face instead of having to remove the boundary face argument. What happens if a user puts periodic for the face but has geom.is_periodic = 0 for that direction?

marchdf:
> the geom.is_periodic still has to be correct, but this now allows the user to put a periodic argument for that boundary face instead of having to remove the boundary face argument.

yes or (currently) it has to be `xlo.type = null`, which is weird. 

> What happens if a user puts periodic for the face but has geom.is_periodic = 0 for that direction?

If I do `xlo.type=periodic geometry.is_periodic=0   1   0` then it aborts with:
```
Invalid BC specification for non-periodic boundary = periodic
```

mbkuhn (approved):
works for me!

PR #1349, opened 2024-11-15
Enable int field refinement

marchdf:
## Summary

Enable int field refinement for the user.  For example, tagging on terrain blanking:

Screenshot 2024-11-18 at 10 47 49 AM

@mbkuhn This should also work with the multiphase stuff your were talking about last week. 

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [x] new unit-test(s)
- [x] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [ ] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU


marchdf:
Wait for #1353 to merge

marchdf:
No documentation needed because our current documentation is misleading in that in does not distinguish int vs normal fields for refinement ;)

mbkuhn:
love it, well done. do you know if it would be easy to add refinement based on derived fields? I don't need it, but I could see that being useful, and I know we added both int and derived fields to the sampling recently. If it's not straightforward, though, I'd wait until someone requests it.

mbkuhn (approved):

marchdf:
It might be useful but it's a bit annoying. Derived fields are `ScratchFields`. So now you need to parse that, populate the scratch field and then you can refine it. And they are a different type than `Field` or `IntField` so you have to be even more clever with what the "FieldRefinement" has in its member vars.  

I don't like what I had to do with the `m_field` and `m_int_field` but we don't have a base class for fields that would help with templating (which is what I initially tried). It gets messy quick. Mostly because at heart, `MultiFAB` and `iMultiFAB` are completely separate classes so that propagates down the line.

PR #1353, opened 2024-11-18
Ensure consistent ghost cells for terrain

marchdf:
## Summary

Ensure consistent ghost cells for terrain. Make sure that the interior ghost cells of the int fields are properly filled.

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [x] on CPU 
- the regression tests
  - [ ] on GPU 
  - [x] on CPU


marchdf:
One thing to note... for int fields we don't do actual domain BCs so I am only filling the interior BCs for the int fields. One issue with this is that the refiners and samplers, if they happen at a domain BC, will see a default value. The user can be careful about that and fill the domain ghost cells appropriately. Or I can code in the infracstructure for something like an hoextrap for int fields. Thoughts?

hgopalan:
hoextrap should be fine since at the domain boundaries the terrain gets smoothed out close to zero.

marchdf:
This is an example of what I don't want to have happen. Basically it tagged everything in the boundary because the boundary values are "wrong".

Screenshot 2024-11-18 at 10 25 40 AM

marchdf:
Ok I found a better way to do this without fillpatch/fillboundary. Basically I am having the parallelfor loops fill the ghost values properly with what they should be initialized too. It looks like a lot of diffs but I basically just moved code from post_init to initialize_fields. And then used fused mfiter loops that loop on ghost cells as well.

marchdf:
Now my refinement looks like this. Which is what I wanted.

hgopalan (approved):

PR #1355, opened 2024-11-18
Update terrain blanking

marchdf:
## Summary

Update terrain blanking so that the ghost cells are not tagged as blanked. This ensures only the actual part of the blanking gets tagged for refinement and not the whole bottom domain.

Tagging now looks like:
Screenshot 2024-11-18 at 4 06 30 PM

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [x] on CPU 
- the regression tests
  - [ ] on GPU 
  - [x] on CPU

### DIFF
--- a/amr-wind/physics/TerrainDrag.cpp
+++ b/amr-wind/physics/TerrainDrag.cpp
@@ -110,7 +110,7 @@ void TerrainDrag::initialize_fields(int level, const amrex::Geometry& geom)
                 xterrain_ptr, xterrain_ptr + xterrain_size, yterrain_ptr,
                 yterrain_ptr + yterrain_size, zterrain_ptr, x, y);
             levelBlanking[nbx](i, j, k, 0) =
-                static_cast<int>((z <= terrainHt) && (z > 0));
+                static_cast<int>((z <= terrainHt) && (z > prob_lo[2]));
             levelheight[nbx](i, j, k, 0) =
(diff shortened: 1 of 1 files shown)


amr-wind/physics/TerrainDrag.cpp
```diff
@@ -109,7 +109,8 @@ void TerrainDrag::initialize_fields(int level, const amrex::Geometry& geom)
             const amrex::Real terrainHt = interp::bilinear(
                 xterrain_ptr, xterrain_ptr + xterrain_size, yterrain_ptr,
                 yterrain_ptr + yterrain_size, zterrain_ptr, x, y);
-            levelBlanking[nbx](i, j, k, 0) = static_cast<int>(z <= terrainHt);
+            levelBlanking[nbx](i, j, k, 0) =
+                static_cast<int>((z <= terrainHt) && (z > 0));
```
moprak-nrel, line 113: Could this check be `z > prob_lo[2]` instead? Wondering if this is assuming `z = 0` is the boundary.

marchdf: oh good point! yes

moprak-nrel (approved):

PR #1356, opened 2024-11-19
Enable viz of native boundary files

marchdf:
## Summary

This PR add "Header" files for the native boundary plane files. We couldn't visualize these before, now we can. For example, this is viewing the xlo boundary plane and the mesh in the background is the full domain from the plt file:

Screenshot 2024-11-19 at 10 41 47 AM

The one thing is that Visit and Paraview expect plain "Header" file names so to viz different planes you need to simlink the header file you want to viz to "Header".

This also reduces the amount written to disk. For higher levels we wrote the entire BC to disk but now I use a min box to only write out the level data that is populated.

Still a couple things I want to check and also add documentation. 

This is a stepping stone to other things but is useful in and of itself.

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [x] on CPU 
- the regression tests
  - [ ] on GPU 
  - [x] on CPU


marchdf:
Do we know if people tend to run ABL precursors with multiple levels at the BCs? My idea of writing out just the data we need at the BC and then reading that is breaking backwards compatibility. But this would only affect users that have multiple levels touching the boundary and using the native format.

mchurchf:
Marc,

Yes, people increasingly run precursors with multiple levels of refinement that completely span the horizontal extent of the domain, so the boundary has multiple refinement levels on it.

Matt

marchdf:
Thanks, Matt. I am debating what to do here. I like that this is only outputting the right data and not extra data. But I don't like that this might mean some boundary files will be obsolete. grrr. Maybe I just give up the idea of not doing the right thing?

hgopalan:
This can be useful when you have inflow and outflow in the same boundary plane.

marchdf:
Great! I figured out a way to not break backwards compatibility so that won't be a concern.

marchdf:
So things got a little bit crazy in trying to make this work. Basically if it's using the new capability, it uses the header files to read in the boundary geom/domain. If not, it goes back to what was before. 

I am pretty sure I didn't break backwards compatibility. But I would really appreciate help from @hgopalan and/or @rybchuk. Could you run a restart using this branch using a precursor run you already have (ie not using the stuff in this PR)? Basically I want to be sure I didn't break your current files. Thanks in advance!!

rybchuk:
> So things got a little bit crazy in trying to make this work. Basically if it's using the new capability, it uses the header files to read in the boundary geom/domain. If not, it goes back to what was before.
> 
> I am pretty sure I didn't break backwards compatibility. But I would really appreciate help from @hgopalan and/or @rybchuk. Could you run a restart using this branch using a precursor run you already have (ie not using the stuff in this PR)? Basically I want to be sure I didn't break your current files. So don't rerun the precursor! Just the restart. Thanks in advance!!

Looks good to me! I grabbed the code from commit `6241ead` and ran 10 timesteps of an inflow-outflow simulation using BC data that was generated with an older commit.

hgopalan:
Same. Worked fine.

marchdf:
Thanks much! I am assuming "worked fine and results looked reasonable". 

To be clear, this viz capability will only be available for now with new files that are generated past this commit (I might write a script to generate the headers post-facto). All the regtests pass so hopefully we are good to go. Messing with the boundary planes makes me nervous because they are so integral to so many workflows. Let me know ASAP if things aren't working anymore and apologies in advance ;)

marchdf:
@moprak-nrel and @mbkuhn, would you mind checking this? I think I did want I intended to do with this one.

PR #1358, opened 2024-11-19
skip populate_data if requested time is after interpolate time

mbkuhn:
## Summary

Time to debate if this is a good idea... 

This is to allow for that fillpatch(new_time) that gets called by the overset solver prior to the timestep starting. The boundary information does not matter at this point in the algorithm; it'll get filled later when necessary. The downside is that the assert is much less strict, i.e., it can't tell between a mistake and this intended exception.

Any alternative would seem to rely on the code knowing whether the timestep (advance) has begun or not, e.g., if advance_states has been called. But that seems pretty complicated to set up.

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [ ] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU 

## Additional background

Issue Number:


amr-wind/CFDSim.H
```diff
@@ -117,6 +117,13 @@ public:
 
     bool has_mesh_mapping() const { return m_mesh_mapping; }
 
+    void during_overset_advance(const bool flag)
+    {
+        m_during_overset_advance = flag;
+    }
+
+    bool during_overset_advance() const { return m_during_overset_advance; }
```
marchdf, line 125: can you have them have different names? maybe this one should be `is_overset_advance` ? And the other should be `set_overset_advance`? Or maybe just plain `overset_advance`

marchdf:
I like where this is going. I just had a minor thought on clarity of naming functions.

amr-wind/CFDSim.H
mbkuhn: great idea

mbkuhn:
The way I set it up should still be correct with multiple iterations in AMR-Wind, too

amr-wind/CFDSim.H
mbkuhn: I think I need them a little more verbose, though

marchdf: yeah good point.

mbkuhn:
it would be good to accompany this with an exawind-driver reg test

marchdf:
Good work! `100% tests passed, 0 tests failed out of 136`

marchdf (approved):

PR #1359, opened 2024-11-20
Add abort when using a GPU backend with OpenMP

jrood-nrel:


marchdf (approved):
Excellent idea. We have never tested this combination and I wouldn't trust it. We would need extensive tests to make sure that the results are valid before we allowed this combination.

PR #1360, opened 2024-11-20
Post regrid actions for physics on restart with regrid

marchdf:
## Summary

When restarting a simulation with more levels than the checkpoint, regrid is called. This is typically not a problem, except if the fillpatch from coarse is not good enough and the physics class does some non-trivial post-regrid action. This was the case for the terrain fields. I kept it simple by only doing the physics post-regrid to avoid any other unintended consequence.

I accidentally broke the terrain fields on a restart with more levels in #1353 when I moved their initialization from post_init to initialize_fields. This was the right move but didn't account for restarting with more levels. Additionally, having the reinit in the post-regrid means that this will now work with dynamic mesh refinement. 

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [x] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [x] on CPU 
- the regression tests
  - [ ] on GPU 
  - [x] on CPU


hgopalan:
Thank you for the quick fix.

hgopalan:
Works perfectly and no issues with convergence for RANS. 
AMR_RANS

marchdf:
Thanks for checking, @hgopalan !

hgopalan (approved):

amr-wind/physics/TerrainDrag.H
```diff
@@ -28,14 +28,12 @@ public:
 
     void post_init_actions() override {}
 
-    void post_regrid_actions() override {}
+    void post_regrid_actions() override;
```
moprak-nrel, line 31: Any reason why this is changed back to `;` vs `{}`

marchdf: because I actually define this function now with something non-trivial in the cpp file.

moprak-nrel (approved):

PR #1361, opened 2024-11-20
Bounded terrain refinement

hgopalan:
The field refinement with terrain_blank is unbounded and provides refinements in regions which are not of interest. This modification limits the bounds and levels of field refinement based on the user-specified  input file. 

- [ ] Bugfix
- [X ] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [X ] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ X] on CPU 

To do: 

1. Create a regression test 
2. Create a unit test 
BoxRefinement

### DIFF
--- a/amr-wind/utilities/tagging/BoundFieldRefinement.H
+++ b/amr-wind/utilities/tagging/BoundFieldRefinement.H
@@ -1,0 +1,182 @@
+#ifndef BOUNDFIELDREFINEMENT_H
+#define BOUNDFIELDREFINEMENT_H
+
+#include "amr-wind/utilities/tagging/RefinementCriteria.H"
+
+namespace amr_wind {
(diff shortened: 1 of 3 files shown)


marchdf:
Thanks for putting this together! I think there is a much easier way. It would require a small change of the existing FieldRefinement.

Basically, you add a parmparse param:
```
tagging.b1.box_lo = -0.1  -0.52  -0.85
tagging.b1.box_hi =  3.1 -0.45    0.85
```
Then in the checks you just do a `.contains` on the point and that box. 

Does that make sense? I can throw that together if you would like.

PR #1367, opened 2024-11-22
Loosen tolerances of boundary plane asserts

mbkuhn:
## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [ ] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU 

## Additional background

Issue Number:

### DIFF
--- a/amr-wind/wind_energy/ABLBoundaryPlane.cpp
+++ b/amr-wind/wind_energy/ABLBoundaryPlane.cpp
@@ -892,8 +892,8 @@ void ABLBoundaryPlane::populate_data(
     }
 
     AMREX_ALWAYS_ASSERT(
-        ((m_in_data.tn() <= time) || (time <= m_in_data.tnp1())));
-    AMREX_ALWAYS_ASSERT(std::abs(time - m_in_data.tinterp()) < 1e-12);
+        ((m_in_data.tn() <= time - 1e-8) || (time <= m_in_data.tnp1() - 1e-8)));
+    AMREX_ALWAYS_ASSERT(std::abs(time - m_in_data.tinterp()) < 1e-8);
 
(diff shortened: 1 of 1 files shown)


mbkuhn:
@hgopalan could you try this branch on one of your failing simulations?

mbkuhn:
@marchdf wondering if this is needed for the hybrid solver as well

hgopalan:
I can update my branch with this and keep running and let you know if it happens in the future. My failures are 1 in 10 or 1 in 20 so weirdly not reproducible.

mbkuhn:
> I can update my branch with this and keep running and let you know if it happens in the future. My failures are 1 in 10 or 1 in 20 so weirdly not reproducible.

Sounds good. Hopefully you will see them go down with this.

mbkuhn:
@hgopalan have you seen any differences as a result of using this branch?

hgopalan:
Last few weeks no crashes so far with the latest version.

mbkuhn:
Regis had a consistently failing case that this PR fixed. The main problem was on the final timestep, where the closest_index would find the last time in the series and hit the ASSERT there. This is because the actual simulation time (that the code is trying to interpolate to) was 115, but the last time in the file was recorded as 114.99999999999915. Adding a tolerance to closest_index allows this ASSERT to be avoided while still accomplishing things accurately.

The other tolerances I've introduced are not all directly solving known issues, but I think they are all legitimate ways to handle floating point comparisons and may help with Harish's sporadic crashes.

marchdf (approved):

PR #1369, opened 2024-11-22
Terrain blanking informed by ocean waves

mbkuhn:
## Summary

Leveraging the combination of the IB forcing method for terrain with the wave profiles used in a numerical wave tank setting.

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [x] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [ ] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU


mbkuhn:
Linear Waves as Terrain

Evidence that all of the right pieces are showing up. The blanked cells in blue, the drag cells in red, the ABL flow above the waves, and the wave velocity profile as the target velocity within the blanked cells (arrows).

mbkuhn:
W2A waves as terrain

Working for the waves2amr pathway, i.e. using wave modes from HOS-Ocean. Getting the right velocity inside the waves requires the bug fix #1372 (already included).

mbkuhn:
Will be meeting with Harish to hash out the question marks on the surface + turbulence modeling sides.

mbkuhn:
After discussing with Harish, we concluded that this PR is mostly there. The main thing that remains is adapting the wall-model forcing term (applied to cells flagged as "drag" cells). We may consult Manuel Ayala's work for this.

Other possible concerns are the magnitude of the Cd term (applied to cells flagged as "blanked" cells) - too low may mean that the velocity within the waves doesn't change fast enough, and too high may mean stability issues - and overlap of waves and sponge forcing near outlets - may need to smooth waves at inflow or outflow. But these concerns have to do with robustness rather than accuracy.

Other design choices: for now, we are assuming the ocean surface has a uniform temperature in a single simulation. If the simulations get bigger, we may need to change that. We are also assuming a uniform roughness over the waves (z0). This parameter represents the underresolved waves, not the resolved ones that are modeled as terrain.

mbkuhn:
this PR is focused on representing waves as moving terrain. but other things conveniently included:
* update_volume_fractions separate step which corrects an oversight in the ocean waves boundary approach
* removing prefix from several "m_" variables where it didn't apply
* converting "newtime" to "time" in waves2amr, this is more accurate of a term.

mbkuhn:
Copied Manuel's work to set up a form drag term for when waves are used as terrain. Modified the existing viscous drag approach to consider the velocity relative to the "wall" velocity, which in this case is the wave velocity. Because the drag terms are needed when adjacent to the interface, the wave velocity in the cell below the drag cell is used. This could be modified - cells that have some liquid in them but aren't mostly liquid should still have a valid wave velocity...

hgopalan (approved):

marchdf (approved):

PR #1388, opened 2024-12-05
ALM replace abort with if condition check

### LINKED ISSUE #1385: Assertion `info.root_proc > -1' failed
## Bug description
```
  what():  Assertion `info.root_proc > -1' failed, file "/autofs/nccs-svm1_home1/marchdf/exawind/exawind-manager/environments/amr-wind-of/amr-wind/amr-wind/wind_energy/actuator/turbine/fast/turbine_fast_ops.H", line 186
```

This is due to something I did in #1360: specifically, calling `post_regrid_actions`  on L81 of `incflo.cpp`.  `Actuator::post_regrid_actions()` is doing something different than the others and doesn't have the information it would like...

marchdf:
## Summary

In [turbine_fast_ops](https://github.com/Exawind/amr-wind/blob/main/amr-wind/wind_energy/actuator/turbine/fast/turbine_fast_ops.H#L186), we `AMREX_ALWAYS_ASSERT(info.root_proc > -1);`. We do not do that in https://github.com/Exawind/amr-wind/blob/main/amr-wind/wind_energy/actuator/actuator_opsI.H#L18. In that scenario, we have an `if` condition check and then the fix. I have no idea why one has an abort and the other a simple if check. The git blames do not over any information for that difference. 

This PR makes the `turbine_fast_ops.H` operate the same as the others. Before this is merged, I would like help doing due diligence that this doesn't break something even more.

@lawrenceccheung, @ndevelder , @gyalla can I ask you to run this branch on an ALM case (like one of the calibration tests) and tell me if it looks ok? 

## Pull request type

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [ ] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU 

Closes #1385

### DIFF
--- a/amr-wind/wind_energy/actuator/turbine/fast/turbine_fast_ops.H
+++ b/amr-wind/wind_energy/actuator/turbine/fast/turbine_fast_ops.H
@@ -183,11 +183,12 @@ determine_influenced_procs<TurbineFast>(typename TurbineFast::DataType& data)
     info.procs =
         utils::determine_influenced_procs(data.sim().mesh(), info.bound_box);
 
-    AMREX_ALWAYS_ASSERT(info.root_proc > -1);
-    // During regrid, the influenced processes might have changed and might
-    // no longer include the root proc. We insert it back to ensure that it
-    // is always present on the list.
-    info.procs.insert(info.root_proc);
+    if (info.root_proc > -1) {
+        // During regrid, the influenced processes might have changed and might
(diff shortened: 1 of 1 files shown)


marchdf:
This didn't work.

PR #1389, opened 2024-12-05
Add  ALM reg tests

marchdf:
## Summary

Add an ALM openfast reg tests. It's time. As much as we've been resisting this... It's critical enough and I recently broke something (#1385) that would have been caught if we had a reg test.

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [x] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [ ] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU


marchdf:
So far I have this.  Red = T0, first run, green = T1, first run, blue = T0, restart run, orange = T1, restart run. They all do the same thing, which is expected. And blue and orange start later (since they are a restart)

[turbines.pdf](https://github.com/user-attachments/files/18042561/turbines.pdf)

marchdf:
And with this restart case I can confirm the breakage I saw in #1385 and that breakage is not in 3.2.0.

test/CMakeLists.txt
```diff
@@ -323,6 +323,93 @@ if(AMR_WIND_ENABLE_W2A)
...
+      WORKING_DIRECTORY
+      "${ACT_UNIFORM_ALM_TEST_BINARY_DIR}"
+      COMMAND_ERROR_IS_FATAL ANY)
+    execute_process(
+      COMMAND bash -c "sed -i -e 's/1   WakeMod/0   WakeMod/g' ${RTEST_ACT_UNIFORM_ALM_TEST_BINARY_DIR}/${5MW_LAND_DIR}/${AERO_FILE}"
+      WORKING_DIRECTORY
+      "${ACT_UNIFORM_ALM_TEST_BINARY_DIR}"
+      COMMAND_ERROR_IS_FATAL ANY)
+    execute_process(
+      COMMAND bash -c "sed -i -e 's/True          YawDOF/False         YawDOF/g' ${RTEST_ACT_UNIFORM_ALM_TEST_BINARY_DIR}/${5MW_LAND_DIR}/${ELASTO_FILE}"
```
jrood-nrel, line 402: Unfortunately sed isn't portable.

marchdf: yeah... got another idea?

amr-wind/wind_energy/actuator/Actuator.cpp
```diff
@@ -63,6 +64,7 @@ void Actuator::pre_init_actions()
         pp.query("type", type);
         pp1.query("type", type);
         AMREX_ALWAYS_ASSERT(!type.empty());
+        cnt_turbfastline += (type == "TurbineFastLine") ? 1 : 0;
```
lawrenceccheung, line 67: I think it's the sum total of all OpenFAST (both `TurbFastLine` and `TurbFastDisk`) instances which must be equal to or less than the number of ranks.

lawrenceccheung (changes requested):
Thanks @marchdf, this is great and something that has been sorely needed.  

I'm adding in a note that the input files were generated with the notebook https://gist.github.com/lawrenceccheung/a775e8eb5162e906da2a75ce67169139, in case we want to generate other reg test type problems in the future.  

Also, we should adopt a similar process of downloading the NREL5MW files from `OpenFAST/r-test` for the [benchmark problems](https://github.com/Exawind/exawind-benchmarks/tree/main/exawind/NREL_5MW_Turbine).  Or least, noting which commit they were downloaded from and documenting which changes to the openfast model were necessary for the benchmark problems to run.  I know it's for a separate problem, but we should be handling them consistently (Tagging @ndevelder on this).

Lawrence

amr-wind/wind_energy/actuator/Actuator.cpp
marchdf: good point. Fixed.

test/CMakeLists.txt
marchdf: I actually thought of a better way. We carry a patch and `git apply` it. Added benefit of this is that if the patch doesn't apply due to file changes in openfast, we know about it immediately.

jrood-nrel: That's a great idea.

jrood-nrel (approved):

lawrenceccheung (approved):

PR #1399, opened 2024-12-10
Convert AMREX_GPU_HOST_DEVICE to AMREX_GPU_DEVICE

jrood-nrel:


marchdf:
Good to be more deliberate about this. You won't be able to do some of these because we call on host and on device...

PR #1409, opened 2024-12-13
Add plot file and boundary file Python tools

marchdf:
## Summary

This provides a couple new Python tools to manipulate plot files and boundary planes:

- `amrex_plotfile.py` is analogous to `amrex_particle.py` in that it allows the user to read an AMReX plot file
- `example_plotfile_io.py` shows how one can use that capability to read a plot file and the write it back out. Running this code will generate new plot files that are identical (modulo parallelism and the number of files) in an `fcompare` sense.
- `refine_native_boundary_plane.py` is a tool to refine boundary planes. It also shows, vial `plt.mfs[ilev].to_xp()` how to grab a numpy view of the data in a plot file. Yay. And then it uses scipy interp to perform the interpolation.

Caveats:
- The dependency is on `pyamrex`.
- I am not allowing MPI/OMP for now.
- Really shouldn't be used for post-processing, `yt` is a much better package for that. But I know this comment won't stop you ;) 

These tools give users the ability to modify/edit/define whatever boundary planes they would like using a Python interface. 

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [ ] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU 

## Additional background

Issue Number:


marchdf:
Here's a single level example:

## Original boundary plane
Screenshot 2024-12-16 at 12 10 51 PM

## Interpolated boundary plane
Screenshot 2024-12-16 at 12 11 07 PM

marchdf:
And this is multilevel:
Screenshot 2024-12-16 at 2 28 26 PM

mbkuhn (approved):
nothing to see here ;)

moprak-nrel (approved):

marchdf:
I added one more tool: `generate_native_boundary_plane_header.py`. This generates a header file for boundary planes that were written with amr-wind v3.2.0 and prior.

This was an old boundary plane file that I had lying around and that I can now visualize after adding the header files:
Screenshot 2024-12-17 at 11 55 15 AM

PR #1423, opened 2024-12-23
Fix some codeql warnings

marchdf:
## Summary

Fix some codeql warnings.

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [ ] on CPU 
- the regression tests
  - [ ] on GPU 
  - [x] on CPU 

## Additional background

Issue Number:

### DIFF
--- a/amr-wind/equation_systems/icns/icns_advection.cpp
+++ b/amr-wind/equation_systems/icns/icns_advection.cpp
@@ -26,15 +26,11 @@ amrex::Array<amrex::LinOpBCType, AMREX_SPACEDIM> get_projection_bc(
             r[dir] = amrex::LinOpBCType::Periodic;
         } else {
             auto bc = bctype[amrex::Orientation(dir, side)];
-            switch (bc) {
-            case BC::pressure_outflow: {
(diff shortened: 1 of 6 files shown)


marchdf:
Ok the last three (`Local variable address stored in non-local memory`) are hard. I moved everything to shared_ptr and it all built fine. Except I am getting `libc++abi: terminating due to uncaught exception of type std::__1::bad_weak_ptr: bad_weak_ptr`. Which I thought was because I was registering inside the constructor but I moved it to some other place and I still get it. Maybe it's a problem for another day. I think the issue is the following: I am using shared_from_this() on an object which has no shared_ptr pointing to it.

Which means I would need to `make_shared` on this `MomentumSource`. Which is a bummer because the `create` method of `Factory` creates (sensibly) a `unique_ptr`. 

Keeping the stash of this for a rainy day.

jrood-nrel (approved):

PR #1415, opened 2024-12-19
basic implementation of current for linear waves, numerical beach

mbkuhn:
## Summary

Adding the capability to have current (with waves)

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe):

## Checklist

The following is included:

- [ ] new unit-test(s)
- [x] new regression test(s)
- [ ] documentation for new capability

This PR was tested by running:

- the unit tests
  - [ ] on GPU 
  - [ ] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU 

## Additional background

Issue Number:

PR #1424, opened 2024-12-23
Fill inflow routines for ocean waves

mbkuhn:
## Summary

The boundary conditions for the ocean wave forcing framework have never been thorough - this is a step in the right direction. Modeled after the boundary plane and MPL fill routines, this uses the target fields of the wave forcing to fill in the boundary cells for vof, density, and velocity fields. For velocity, this new fill applies the wave velocity values within the wave, only overwriting the liquid velocities and allowing other BCs to apply to the gas phase. For now, I'm not attempting to make it compatible with the boundary plane inflow; we should just rely on the boundary planes when ocean waves are present.

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [x] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [x] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU 

## Additional background

There have been weird pressure signatures at the inflow boundaries with waves for a while now, and this significantly improves that issue. The BC problems became really apparent when I tried to add current (see #1415), where strong negative velocities were induced in the gas phase, and this also addresses that problem. Right now, the implementation doesn't fully get the temporal discretization right, but I may add to that to improve it.


mbkuhn:
Will need to look at the reg tests and confirm it affects the expected ones and not others.

mbkuhn:
@marchdf could you check this for diffs?

marchdf:
I am seeing:
```
The following tests FAILED:
         31 - abl_multiphase_laminar (Failed)
         74 - ow_linear (Failed)
         75 - ow_linear_init_waves (Failed)
         76 - ow_stokes (Failed)
        124 - ow_w2a (Failed)
        125 - abl_multiphase_w2a (Failed)
        135 - abl_multiphase_laminar_input (Not Run)
```

Is that expected? Some of these have tiny diffs, others large (e.g. `ow_linear_init_waves`).

mbkuhn:
Awesome, thanks for running those! I will look into more detail for them. The ow_linear* and ow_stokes cases should have noticeable diffs. For these cases, the fill routine is on and should be fixing something that was missing before. I will check these manually to make sure it looks good. The abl_multiphase* cases and ow_w2a should have no diffs because they're periodic, but if they have tiny diffs that is encouraging. I have to dig a little to understand exactly where the tiny diffs arise.

Does this track with the diffs you have?

marchdf:
I think so? 
`ow_w2a` -> tiny diffs
`ow_linear_init_waves` -> large diffs
`abl_multiphase_w2a` -> tiny
`ow_linear` -> large
`abl_multiphase_laminar` -> tiny
`ow_stokes` -> large

mbkuhn:
alright, this is ready to go. diffing cases look good to me. I found one more thing that needed to be added, which fixes how the pressure looks when there are multiple mesh levels at the boundary.

mbkuhn:
After these changes, it will make sense to deprecate the "wave_generation" bc and just use mass_inflow in its place. I think the idea behind having wave_generation as a separate bc is now properly fulfilled in this dedicated fill. At this point, the only difference between mass_inflow and wave_generation is the more stringent input line requirements for mass_inflow.

marchdf (approved):

PR #1425, opened 2024-12-23
Vof fillpatch fixes

### LINKED ISSUE #1342: Offshore boundary plane failing time tolerance
## Bug description
Likely related to [this PR](https://github.com/Exawind/amr-wind/pull/1329), and I am using the most recent commit on `main`, `740679c8`. I'm running an offshore ABL simulation with linear waves. I have achieved a spun-up state at iteration `80000`, and now I am trying to run an inflow-outflow simulation. I believe that I have successfully written out boundary planes using the following:
```
ABL.bndry_file                           = bndry_file.native
ABL.bndry_io_mode                        = 0
ABL.bndry_planes                         = ylo xlo
ABL.bndry_output_start_time              = 1.0
ABL.bndry_var_names                      = velocity_mueff velocity density p velocity_src_term gp vof_mueff levelset interface_curvature vof vof_src_term temperature_mueff temperature temperature_src_term ow_levelset ow_vof ow_velocity
```
(I know that I'm probably saving out too many boundary variables, but I saw the same error with `= velocity temperature vof`)

I then kick off the inflow-outflow simulation using the following variables:
```
ABL.bndry_file                           = ../write_lev0/bndry_file.native
ABL.bndry_io_mode                        = 1
ABL.bndry_planes                         = ylo xlo
ABL.bndry_output_start_time              = 1.0
ABL.bndry_var_names                      = velocity_mueff velocity density p velocity_src_term gp vof_mueff levelset interface_curvature vof vof_src_term temperature_mueff temperature temperature_src_term ow_levelset ow_vof ow_velocity # velocity temperature vof
```

The simulation errors out during the first timestep.

## Steps to reproduce
I have shared all the files needed to run the bc-write simulation and bc-read simulation in `/scratch/orybchuk/share/debug_inflow_outflow` on Kestrel. I've uploaded many of the files (aside from the large ones) here too.

[debug_inflow_outflow.zip](https://github.com/user-attachments/files/17706790/debug_inflow_outflow.zip)

1. Compiler used
   - Oneapi
2. Operating system
   - Linux
3. Hardware:
   - CPU
4. Machine details ():
Kestrel
```
Currently Loaded Modules:
  1) intel/2023.2.0	  7) cray-libsci/22.10.1.2
  2) craype/2.7.30        8) PrgEnv-intel/8.5.0
  3) cray-dsmml/0.2.2     9) cray-python/3.11.5
  4) libfabric/1.15.2.0  10) binutils/2.41
  5) craype-network-ofi  11) intel-oneapi-compilers/2023.2.0
  6) cray-mpich/8.1.28
```
5. Input file attachments: See the `.zip` file.
6. Error (paste or attach):
```
Step: 80011 dt: 0.05 Time: 6542.942863 to 6542.992863
CFL: 0.402303 (conv: 0.402303 diff: 0 src: 0 )

Godunov:
  System                     Iters      Initial residual        Final residual
...
  what():  Assertion `std::abs(time - m_in_data.tinterp()) < 1e-12' failed, file "/kfs2/projects/ai4wind/orybchuk/exawind-manager/environments/ai4wind-nov24/amr-wind/amr-wind/wind_energy/ABLBoundaryPlane.cpp", line 896
```
7. If this is a segfault, a stack trace from a debug build (paste or attach):
N/A

## Expected behavior
The bc-read simulation should run past the first timestamp without crashing.

## AMR-Wind information

```
                AMR-Wind (https://github.com/exawind/amr-wind)

  AMR-Wind version :: v3.2.0-8-g740679c8-DIRTY
  AMR-Wind Git SHA :: 740679c87925db41521f477e8724daa0c69cc670-DIRTY
  AMReX version    :: 24.09-45-g6d9c25b989f1

  Exec. time	   :: Mon Nov 11 10:48:20 2024
  Build time	   :: Nov 11 2024 10:17:52
  C++ compiler     :: IntelLLVM 2023.2.0

  MPI              :: ON    (Num. ranks = 384)
  GPU              :: OFF
  OpenMP           :: OFF

  Enabled third-party libraries:
    NetCDF    4.9.2

 See https://github.com/Exawind/amr-wind/blob/development/LICENSE for details.
```

## Additional context
I tried initializing the bc-read simulation from `chk80000` and `chk80010`, and it fails for both checkpoints.

For the record, the timing of `chk80010` looks like `Step: 80011 dt: 0.05 Time: 6542.942863 to 6542.992863`, and the first few timestamps of the BC are:
```
80000 6542.4428627258721
80001 6542.5721400248704
80002 6542.6563112443018
```
so the BC definitely has data before the IC. The precursor simulation was run with an adaptive timestamp, but the bc-write and bc-read simulations are currently being run with a small fixed timestamp. I was also hitting the same error with a version of AMR-Wind that I compiled from `main` in October.

mbkuhn:
## Summary

Some fillpatches were lazy within the vof equation system. Those prevented the boundary planes from working properly. This fixes that.

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [x] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [x] on CPU 
- the regression tests
  - [ ] on GPU 
  - [x] on CPU 

## Additional background

2 parts to these changes (still need to do the second part)

1. providing new_time to the fillpatch calls within vof advection, to avoid failing asserts when reading boundary planes
2. getting time and space right for the boundary values of vof because it has a different advection discretization, not expecting values at the boundary itself nor at n+1/2 in time.

Issue Number: closes #1342


marchdf (approved):

mbkuhn:
New regression test proves that this closes the issue. Need to check the time accuracy next.

mbkuhn:
Will need to check 2-phase regression tests because of the change to set_density_via_vof. Did check abl_multiphase_laminar locally and there were no diffs from that change.

mbkuhn:
On time accuracy, the vof advection discretization operates from n to n+1, meaning that the boundary values need to be at n, as opposed to at n+1/2 like the other variables in the code. This is handled appropriately by the copy that takes place at the beginning of the vof advection step, where the n values of vof are copied to the working state of the vof variable (which happens to be n+1/2). So, even though the vof boundaries would be filled in pre_advance by n+1/2 state, these are overwritten, which is good.

However, the direction-split nature of the advection makes time consistency really complicated. This means that vof is first advected in x, then in y, then in z. And each step has a fillpatch after it. Also, the order (x, y, z) changes every timestep. The changes in the boundary vof values from n to n+1 would need to shift in the same directions as the advection to get it absolutely perfect, but updating to n+1 after the first step is probably the second-best possible way of doing it. This is the approach in this PR, so that is also good.

The final problem is that the vof advection is truly cell-centered, even at boundaries. So this means that the boundary data shouldn't be interpolated to the face; it should just be used as the cell value. Figuring out how to do this right will require digging into the code a bit to see if a modification to the boundary plane interpolation would be feasible, and also to confirm the nature of vof advection at the boundaries. Let's push that to another PR.

mbkuhn:
multiphase reg tests passed locally, except for a few that had diffs at machine precision.

PR #1426, opened 2024-12-27
Meso-Micro Coupling with no precursor for Terrain 

hgopalan:
The current workflow in AMR - Wind requires a precursor simulation to generate boundary plane for running complex terrain simulations. However, this is not a requirement. This PR proposes an alternative path for running simulations over complex terrain. 

1. Based on met-mast measurement, a 1-D vertical wind profile is created along with the required geostrophic wind. 
2. The simulation is initialized with the 1-D profile. 
3.  All the horizontal boundary conditions are replaced with pressure-outflow boundary condition with a sponge layer, forcing the 1-D profile. 
4.  A forcing zone is added on the upper portion of the domain to drive the wind speed and temperature to be consistent with the 1-D vertical profile. 

Future Work:

1. Adding the nudging terms within domain for met-mast or Lidar data 
2.  Replacing pressure-outflow with WRF-mesoscale boundary plane data and removing the flat terrain restrictions at the boundary. 

- [ ] Bugfix
- [ X] Feature
- [ ] Build related changes

SunZia Wind Farm LES - 150 kms 

SunZia_chopped

Flatirons - RANS - 150 kms 

Flaitrons_chopped


lawrenceccheung:
Hi @hgopalan,

Thanks for documenting the process.  One question I have about the MMC coupling with terrain is how we account for the fact that the turbulence will take time to spin-up, and how a time-varying MMC profile can be accounted for.  One way to handle it is to include a large, flat inflow region where the turbulence can develop.  Depending on the ABL stability state, it might require a shorter/longer region for the turbulence profile to adjust.  

Once the 1D MMC profile is time-varying, though, that introduces a time delay between when the velocity/temperature changes and when that turbulence enters the domain.  @mchurchf and @ewquon might have some ideas on what to do too.

Lawrence

hgopalan:
> Hi @hgopalan,
> 
> Thanks for documenting the process. One question I have about the MMC coupling with terrain is how we account for the fact that the turbulence will take time to spin-up, and how a time-varying MMC profile can be accounted for. One way to handle it is to include a large, flat inflow region where the turbulence can develop. Depending on the ABL stability state, it might require a shorter/longer region for the turbulence profile to adjust.
> 
> Once the 1D MMC profile is time-varying, though, that introduces a time delay between when the velocity/temperature changes and when that turbulence enters the domain. @mchurchf and @ewquon might have some ideas on what to do too.
> 
> Lawrence

The purpose of this workflow  is mainly related to providing a simple method for AEP calculation for the RANS model similar to the method used in commercial RANS tools. The usability of the method and adoption for LES needs some investigation. If a terrain is complex, it does not need a lot of time to spin-up as the heterogeneity generate the flow structures. The terrain is smoothed in the horizontal boundaries and when they encounter a change in elevation  it generated the structures for the cases I tested with mountains. 

There is no restriction on specifying a time-varying 1-D profile. We just need to move the profile reader to be time-dependent. Will require some code changes but is doable. 

There is some difference with the method proposed here. Our aim is to match the met-mast data while providing a reasonable boundary condition which will not cause divergence of the pressure poisson solver. We are not worried about flow spin-up and other issues and expect that the presence of complex terrain will generate the structure. 

We may have to tweak the workflow in future if there are discrepancies with LES. However, this method should be acceptable for RANS.

ewquon:
> Hi @hgopalan,
> 
> Thanks for documenting the process. One question I have about the MMC coupling with terrain is how we account for the fact that the turbulence will take time to spin-up, and how a time-varying MMC profile can be accounted for. One way to handle it is to include a large, flat inflow region where the turbulence can develop. Depending on the ABL stability state, it might require a shorter/longer region for the turbulence profile to adjust.
> 
> Once the 1D MMC profile is time-varying, though, that introduces a time delay between when the velocity/temperature changes and when that turbulence enters the domain. @mchurchf and @ewquon might have some ideas on what to do too.
> 
> Lawrence

What we could do for terrain is instead of a planar average, use a temporal average at a specified location (e.g., at the met mast). There is a question of whether you make the forcing constant with altitude or constant with height AGL; you can also blend from the latter to the former.

hgopalan:
LES structures are not generated when the terrain is smooth. Added thermal perturbations to generate the flow structures. Sample result are shown for a flat surface without precursor and LES (dx=dy=60m). This is still preliminary and need further analysis on the strength of the structure on numerical stability for complex terrains. 

Horizontal Plane 

Horizontal

Vertical Plane 

Vertical

hgopalan:
Added regression tests with and without perturbation forcing

docs/sphinx/user/features.rst
```diff
@@ -116,6 +116,8 @@ Flow physics
 
    * Arbitrarily spatially and time varying boundary conditions using Python tools [:ref:`inp <inputs_native_boundary_plane>`]
 
+   * Sponge layer driven ABL simulations
```
marchdf, line 119: I am a bit confused. Why is this in the boundary condition section? This feels like just another element in the bullet `* Atmospheric boundary layer (ABL):`

amr-wind/equation_systems/temperature/source_terms/TemperatureFreeAtmosphereForcing.H
```diff
@@ -0,0 +1,54 @@
...
+    amrex::Vector<amrex::Real> m_theta_heights;
+    amrex::Vector<amrex::Real> m_theta_values;
+    amrex::Gpu::DeviceVector<amrex::Real> m_theta_heights_d;
+    amrex::Gpu::DeviceVector<amrex::Real> m_theta_values_d;
+    amrex::Real m_meso_start{600};
+    amrex::Real m_meso_timescale{30};
+    bool m_horizontal_sponge{false};
+    amrex::Real m_sponge_strength{1.0};
+    amrex::Real m_sponge_density{1.0};
+    amrex::Real m_sponge_distance_west{-1000};
```
marchdf, line 42: are these just mean to be big numbers? If so, use the ones in `constants`

amr-wind/equation_systems/temperature/source_terms/PertForcing.H
```diff
@@ -0,0 +1,43 @@
...
+    const CFDSim& m_sim;
+    const SimTime& m_time;
+    const amrex::AmrCore& m_mesh;
+    amrex::Real m_xstart;
+    amrex::Real m_xend;
+    amrex::Real m_ystart;
+    amrex::Real m_yend;
+    amrex::Real m_zstart;
+    amrex::Real m_zend;
+    int m_time_index{100};
```
marchdf, line 37: is 100 a good default value? I would think it should be either a big value or a small value?

marchdf (commented):
Thanks for adding this in! I made a first pass. LMK when I can do a second one.

docs/sphinx/user/features.rst
hgopalan: This is a different way of specifying the boundary conditions through a fringe or sponge layer to force the simulations.

amr-wind/equation_systems/temperature/source_terms/TemperatureFreeAtmosphereForcing.H
hgopalan: The distances are measured in meters within the domain. We use between 1-3 kms of sponge.

amr-wind/equation_systems/temperature/source_terms/PertForcing.H
hgopalan: This number represents how often we add the perturbations. If we add too often it will cause numerical instability. Also if it is a large number the structures dissipate.

amr-wind/equation_systems/icns/source_terms/VelocityFreeAtmosphereForcing.H
```diff
@@ -0,0 +1,46 @@
...
+        const int lev,
+        const amrex::MFIter& mfi,
+        const amrex::Box& bx,
+        const FieldState /*fstate*/,
+        const amrex::Array4<amrex::Real>& src_term) const override;
+
+private:
+    const SimTime& m_time;
+    const amrex::AmrCore& m_mesh;
+    std::string m_1d_rans;
```
marchdf, line 30: rename: `m_1d_rans_filename` or something like that.

amr-wind/equation_systems/icns/source_terms/VelocityFreeAtmosphereForcing.cpp
```diff
@@ -0,0 +1,152 @@
...
+        m_velocity.state(field_impl::dof_state(fstate))(lev).const_array(mfi);
+    const amrex::Real sponge_start = m_meso_start;
+    const amrex::Real meso_timescale = m_time.delta_t();
+    const auto vsize = m_wind_heights_d.size();
+    const auto* wind_heights_d = m_wind_heights_d.data();
+    const auto* u_values_d = m_u_values_d.data();
+    const auto* v_values_d = m_v_values_d.data();
+    const auto* w_values_d = m_w_values_d.data();
+    const bool has_terrain = this->m_sim.repo().field_exists("terrain_height");
+    if (has_terrain) {
```
marchdf, line 78: let's consolidate these 2 branches since they look so similar. I think the only difference is the calculation of z? I would move the `if has_terrain` into the parallel for. It's not ideal but I would rather than then the code duplication.

marchdf: Or do it like you do below with the perturbation forcing

amr-wind/equation_systems/temperature/source_terms/PerturbationForcing.cpp
```diff
@@ -0,0 +1,77 @@
...
+                    (has_terrain) ? height_arr(i, j, k, 0) : 0.0;
+                const amrex::Real blank_arr_cell =
+                    (has_terrain) ? blank_arr(i, j, k, 0) : 0.0;
+                const amrex::Real x = prob_lo[0] + (i + 0.5) * dx[0];
+                const amrex::Real y = prob_lo[1] + (j + 0.5) * dx[1];
+                const amrex::Real z = std::max(
+                    prob_lo[2] + (k + 0.5) * dx[2] - height_arr_cell,
+                    0.5 * dx[2]);
+                const amrex::RealVect point{x, y, z};
+                if (pert_box.contains(point)) {
```
marchdf, line 68: nicely done

amr-wind/equation_systems/temperature/source_terms/TemperatureFreeAtmosphereForcing.cpp
```diff
@@ -0,0 +1,163 @@
...
+    const amrex::Real meso_timescale = m_meso_timescale;
+    const auto vsize = m_theta_heights_d.size();
+    const auto* theta_heights_d = m_theta_heights_d.data();
+    const auto* theta_values_d = m_theta_values_d.data();
+    const bool has_terrain = this->m_sim.repo().field_exists("terrain_height");
+    if (has_terrain) {
+        auto* const m_terrain_height =
+            &this->m_sim.repo().get_field("terrain_height");
+        const auto& terrain_height = (*m_terrain_height)(lev).const_array(mfi);
+        amrex::ParallelFor(
```
marchdf, line 73: this also looks like the one on L144. Think you can consolidate that?

marchdf (changes requested):

marchdf:
Good work, just had some thoughts for avoiding some duplicate code. I am running the tests on GPU now.

amr-wind/equation_systems/tke/source_terms/KransAxell.cpp
```diff
@@ -58,57 +94,70 @@ void KransAxell::operator()(
...
             std::max((z - sponge_start) / (probhi[2] - sponge_start), 0.0);
+        if (zi > 0) {
+            ref_tke = (vsize > 0) ? interp::linear(
+                                        wind_heights_d, wind_heights_d + vsize,
+                                        tke_values_d, z)
+                                  : tke_arr(i, j, k, 0);
+        }
         const amrex::Real sponge_forcing =
-            zi * zi * (tke_arr(i, j, k) - ref_tke);
+            1.0 / dt * (tke_arr(i, j, k) - ref_tke);
```
moprak-nrel, line 133: Is the `1/dt` fixing a previous bug? The dimensions of the `sponge_forcing` array should now be different right?

moprak-nrel (commented):
Good work! Have a few quick questions on the actual model.

amr-wind/equation_systems/temperature/source_terms/TemperatureFreeAtmosphereForcing.H
```diff
@@ -0,0 +1,54 @@
...
+
+private:
+    const amrex::AmrCore& m_mesh;
+    const Field& m_temperature;
+    amrex::Vector<amrex::Real> m_theta_heights;
+    amrex::Vector<amrex::Real> m_theta_values;
+    amrex::Gpu::DeviceVector<amrex::Real> m_theta_heights_d;
+    amrex::Gpu::DeviceVector<amrex::Real> m_theta_values_d;
+    amrex::Real m_meso_start{600};
+    amrex::Real m_meso_timescale{30};
```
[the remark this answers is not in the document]
hgopalan: the 30 seconds is sort of commonly used.

amr-wind/equation_systems/tke/source_terms/KransAxell.cpp
```diff
@@ -58,57 +94,70 @@ void KransAxell::operator()(
...
         amrex::Real bcforcing = 0;
-        const amrex::Real ux = vel(i, j, k, 0);
-        const amrex::Real uy = vel(i, j, k, 1);
         const amrex::Real z = problo[2] + (k + 0.5) * dx[2];
         if (k == 0) {
+            const amrex::Real ux = vel(i, j, k + 1, 0);
+            const amrex::Real uy = vel(i, j, k + 1, 1);
             const amrex::Real m = std::sqrt(ux * ux + uy * uy);
-            const amrex::Real ustar = m * kappa / std::log(z / z0);
+            const amrex::Real ustar = m * kappa / std::log(3 * z / z0);
```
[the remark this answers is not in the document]
hgopalan: Because we use the forcing term to set the boundary condition, cannot use the value of the cell in which the forcing term is set and we use something above it.

amr-wind/equation_systems/tke/source_terms/KransAxell.cpp
hgopalan: Changing to 1/dt helped in speeding the convergence and reducing oscillation in the free atmosphere

moprak-nrel: Does the dimensional change not matter? It changed from `[No-units] * tke` to `[1/T] * tke`? or is the `1` some type of dimensional constant?

hgopalan: The original zi also had a model constant which basically returns 1/s. It was never explicitly mentioned.

amr-wind/equation_systems/tke/source_terms/KransAxell.H
```diff
@@ -38,16 +38,33 @@ private:
...
+    //! Reference temperature
+    std::unique_ptr<ScratchField> m_ref_theta;
+    bool m_horizontal_sponge{false};
+    amrex::Real m_sponge_strength{1.0};
+    amrex::Real m_sponge_density{1.0};
+    amrex::Real m_sponge_distance_west{-1000};
+    amrex::Real m_sponge_distance_east{1000};
+    amrex::Real m_sponge_distance_south{-1000};
+    amrex::Real m_sponge_distance_north{1000};
+    int m_sponge_west{0};
```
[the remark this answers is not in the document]
hgopalan: I kept them as ints for future modifications when the different boundaries can have other non-zero positive values when we use different driving winds

amr-wind/equation_systems/tke/source_terms/KransAxell.cpp
```diff
@@ -117,15 +166,84 @@ void KransAxell::operator()(
...
+                ;
             });
+        if (m_horizontal_sponge) {
+            const amrex::Real sponge_strength = m_sponge_strength;
+            const amrex::Real sponge_density = m_sponge_density;
+            const amrex::Real start_east = probhi[0] - m_sponge_distance_east;
+            const amrex::Real start_west = problo[0] - m_sponge_distance_west;
+            const amrex::Real start_north = probhi[1] - m_sponge_distance_north;
+            const amrex::Real start_south = problo[1] - m_sponge_distance_south;
+            const int sponge_east = m_sponge_east;
```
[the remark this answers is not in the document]
hgopalan: Same as above. Currently it is 1/0 but there will be future changes when they can be positive integers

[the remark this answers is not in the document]
moprak-nrel: one last question: would that int work with 

```
                    amrex::Real xi_start =
                        (start_west - x) / (start_west - problo[0]);
                    xi_start = sponge_west * std::max(xi_start, 0.0);
                    xstart_damping =
                        sponge_west * sponge_strength * xi_start * xi_start;
```

That int seems to be multiplied as a cubed power in the damping calculation

marchdf (approved):

moprak-nrel (approved):

Issue #1417, opened 2024-12-19
Particles reader in Paraview 5.13.1 is broken


rthedin:
The latest Paraview version 5.13.1 is no longer able to open sampling output in native/particle format. It works fine in Paraview 5.11.*. I have not tested 5.12. It could be that Paraview has updated their readers and they broke compatibility. 

Here's a screenshot of the error:
![image](https://github.com/user-attachments/assets/e866fbdd-ae9c-437e-bf87-2410332e5bc4)

moprak-nrel:
If the same file is working within Paraview `5.11.x` and failing in `5.13.x`, I'm inclined to say this is a Paraview issue and not an amr-wind issue.

To debug this a bit more, which version of amr-Wind did you run this case with? Have you had other successful instances where things worked with Paraview `5.13.x`?

rthedin:
I agree with you. I've chatted with Michael and he suggested I posted the issue here so it's documented.

To your questions, I used AMR-Wind 3.2.0. I'm not a heavy Paraview user (let alone latest versions) and I'm unable to comment further on other things working. This bug has been brought to my attention by the TotalEnergies team.

marchdf:
I pinged the amrex slack channel to see if anyone has seen this.

marchdf:
FWIW, I see the same. 5.10 seems fine and 5.13 does same as you.

<img width="739" alt="Screenshot 2024-12-19 at 3 45 54 PM" src="https://github.com/user-attachments/assets/7c8db5af-cd2f-4a0d-93fb-06405d85e695" />

mbkuhn:
> I pinged the amrex slack channel to see if anyone has seen this.

Thanks Marc, that was my plan: getting this recorded here and then sharing with the amrex folks who may also have contacts w/ Paraview.

marchdf:
I have reported this issue to a developer at Kitware and created an issue on their site: https://gitlab.kitware.com/paraview/paraview/-/issues/22835.  Closing this issue as being tracked elsewhere.

PR #1435, opened 2025-01-08
Add a plot header file for sampling particles

marchdf:
## Summary

Add a `Header` file for sampling particles. Long story of why is below.

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [x] on CPU 
- the regression tests
  - [ ] on GPU 
  - [x] on CPU 

## Additional background

Issue #1417 brought to our attention that Paraview >= 5.12 could not read our sampling particles any more. 

This was reported to ParaView here: https://gitlab.kitware.com/paraview/paraview/-/issues/22835. 

It turns out that Paraview in this [PR](https://gitlab.kitware.com/vtk/vtk/-/merge_requests/9698) started requiring a Header file that was written by the WritePlotFile routine (independently of the particles). In the majority of amrex applications, particles are written out in conjunction with a plot file (and in the same directory) so this is typically not an issue because the plot file header is available. However, the AMR-Wind sampling particles are independent (can be written at a different frequency, etc) so our particle directories did not have a Header file. 

This PR adds a Header file to the sampling particle directories. I can now use ParaView 5.13 to viz our particles. The added bonus is that it now recognizes the time stamp associated to those particles. 

Screenshot 2025-01-08 at 9 47 48 AM

Tagging @rthedin as interested.

### DIFF
--- a/amr-wind/utilities/sampling/Sampling.H
+++ b/amr-wind/utilities/sampling/Sampling.H
@@ -118,7 +118,7 @@ protected:
 
     SamplingContainer& sampling_container() { return *m_scontainer; }
 
-    amrex::Vector<std::string> int_var_names()
+    static amrex::Vector<std::string> int_var_names()
     {
         return {"uid", "set_id", "probe_id"};
     };
--- a/amr-wind/utilities/sampling/Sampling.H
+++ b/amr-wind/utilities/sampling/Sampling.H
@@ -8,6 +8,7 @@
 #include "amr-wind/utilities/PostProcessing.H"
 #include "amr-wind/utilities/sampling/SamplerBase.H"
 #include "amr-wind/utilities/sampling/SamplingContainer.H"
+#include <AMReX_PlotFileUtil.H>
 
 /**
  *  \defgroup sampling Data-sampling utilities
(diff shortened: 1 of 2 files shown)


marchdf:
This does make is so that the `sampling_info` file I introduced in #1387 is redundant because it contains time information that is also in the Header file. But I think I will keep that sampling_info file since it seems generally a useful place to store additional information.

mbkuhn (approved):

rthedin:
I would request that we do keep the `sampling_info` file. I have starting adding more info to it on a local-to-me branch to test some things out. Mainly, as of right now, I'm adding a map between the string used for a sampling and the index. That will be needed to ensure post-processing is not dependent on an unmodified input file. I'm planning on doing some more tests and checking in with one of you on how to make my additions to the `sampling_info` more appropriate (right now it's very much not great coding as I am not familiar enough with AMR-Wind codebase). But it seems like now time is part of the Header file, which is fine to me.

PR #1440, opened 2025-01-13
Improve Ocean Waves Fill

mbkuhn:
## Summary

I had an oversight in #1424 regarding the modification of MAC velocities at valid boundary cells (set_inflow_sibling_fields). Other fills haven't exactly implemented this properly (MPL, boundary planes), but this one should be correct now. Discovered the problem because cases relying on #1415 became less stable when incorporating all the ow fill changes. I also made the tolerance match what is typically used for VOF checking.

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [x] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [x] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU

### DIFF
--- a/amr-wind/equation_systems/icns/icns_advection.cpp
+++ b/amr-wind/equation_systems/icns/icns_advection.cpp
@@ -177,10 +177,8 @@ void MacProjOp::set_inflow_velocity(amrex::Real time)
         velocity.set_inflow_sibling_fields(lev, time, mac_vec);
         if (m_phy_mgr.contains("OceanWaves")) {
             auto& ow = m_phy_mgr.get<amr_wind::ocean_waves::OceanWaves>();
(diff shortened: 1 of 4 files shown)


marchdf (approved):

marchdf:
Do we need to fix those other ones? And if so, should we have an issue tracking that?

mbkuhn:
> Do we need to fix those other ones? And if so, should we have an issue tracking that?

Not exactly sure what you mean, but I'll try to answer. #1424 is already merged, so this PR is a follow-on to it. The ow cases in the reg tests don't make the fill flaws super obvious, but when current is introduced, which is #1415, the flaws are really clear. So the fixes in this PR should make #1415 work properly.

PR #1450, opened 2025-01-16
Ocean Waves waves2amr bug fixes - levelset field and initial read

mbkuhn:
## Summary

Fixing 2 bugs. Because the w2a routines only update the ow fields where necessary, the values outside of these areas need to be reasonable. Setting levelset = 0 is bad because gradients can lead to intermediate vof values (between 0 and 1). Also, the modes still need to be read when the waves are not initialized over the entire domain.

Added a regression test that blows up without the first fix and segfaults without the second.

No diffs caused for the existing w2a reg tests because this fixes things that they did not cover.

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [x] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [x] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU


amr-wind/ocean_waves/relaxation_zones/waves2amr_ops.H
```diff
@@ -339,59 +339,51 @@ struct ReadInputsOp<W2AWaves>
                 std::to_string(flag));
         }
 
-        // If init_wave_field is activated and initialization will be done, get
-        // modes on every processor
-        if (wdata.init_wave_field && data.sim().time().time_index() == 0) {
```
mbkuhn, line 344: this looks like a lot of changes here, but really the only change within ReadInputsOp is removing this conditional. The first part needed to be remove to fix a segfault under certain conditions, and the second part did nothing because time_index always == 0 during the read inputs phase of the algorithm because the current time hasn't been calculated yet.

marchdf (approved):

PR #1470, opened 2025-01-29
w2a fields - replace fillpatch with FillBoundary

### LINKED ISSUE #1469: FPE in `abl_w2a_terrain`
## Bug description

https://my.cdash.org/tests/249847834 is showing an FPE in `abl_w2a_terrain`.  Can be seen here too: https://my.cdash.org/tests/249674702

mbkuhn:
## Summary

It's better to rely on the directly populated ghost boundary cells than use fillpatch to fill them. 

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [ ] on CPU 
- the regression tests
  - [ ] on GPU 
  - [x] on CPU  - just the w2a cases

## Additional background

Creates tiny diffs in the w2a cases. The one exception is the ow_w2a_nonperiodic case, but I am willing to trust my understanding of the boundary conditions that things are better without the fillpatch.

Putting an epsilon with the norm also fixes the issue, but when I dug deeper I found that the issue was taking place in ghost cells and that the zero norm wouldn't have been prior to this fillpatch operation.

Issue Number: closes #1469

### DIFF
--- a/amr-wind/ocean_waves/relaxation_zones/waves2amr_ops.H
+++ b/amr-wind/ocean_waves/relaxation_zones/waves2amr_ops.H
@@ -679,18 +679,17 @@ struct UpdateTargetFieldsOp<W2AWaves>
             }
 
             // Average down to get fine information on coarse grid where
-            // possible (may be unnecessary)
+            // possible and update ghost cells (may be unnecessary)
             for (int lev = nlevels - 1; lev > 0; --lev) {
                 amrex::average_down(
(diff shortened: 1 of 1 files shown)


marchdf (approved):

marchdf:
Nice work! Can confirm it gets rid of the FPE.

PR #1475, opened 2025-01-31
Underwater terrain-aware wave forcing

mbkuhn:
## Summary

Turn off wave forcing where there is terrain, and modify ocean wave boundary to also be aware.

## Pull request type

- [x] Bugfix
- [x] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [x] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [x] on CPU 
- the regression tests
  - [ ] on GPU 
  - [x] on CPU  -- the ow_* and *terrain* tests, specifically

## Additional background

For continuing collaboration with water power


mbkuhn:
Once this gets a regression test, we can upgrade it to "ready for review".

mbkuhn:
Main point of PR: allow terrain drag to do its thing underwater.

Wave forcing modifies velocity fields within the relaxation zones -- now it will leave cells untouched that are "blanked" (inside terrain) or marked as drag cells (velocity is modified by drag term).

Wave boundary fill sets the boundary velocity to the target velocity of the wave profile -- now it will set the velocity to 0 in boundary cells that are adjacent to "blanked" cells so no flow is pushing directly into terrain at the boundary.

mbkuhn:
Screenshot 2025-02-03 at 10 52 37 AM

From reg test after 10 steps. Enlarged the wave to be able to see it. In application, test would typically be run with some mesh refinements.

amr-wind/ocean_waves/boundary_ops/OceanWavesBoundary.cpp
```diff
@@ -98,13 +110,26 @@ void OceanWavesBoundary::set_velocity(
...
             amrex::ParallelFor(
                 bx, [=] AMREX_GPU_DEVICE(int i, int j, int k) noexcept {
                     for (int n = 0; n < numcomp; n++) {
                         if (targ_vof(i, j, k) > constants::TIGHT_TOL) {
                             arr(i, j, k, dcomp + n) =
                                 targ_arr(i, j, k, orig_comp + n);
                         }
+                        if (terrain_and_vof) {
+                            // Terrain-blanked adjacent cell means 0 velocity
+                            const amrex::IntVect current_iv{i, j, k};
```
marchdf, line 126: you could just move this to the first line of the parallel for `const amrex::IntVect iv{i, j, k};` and replace all `(i,j,k,x)` by `(iv, x)`.

amr-wind/ocean_waves/boundary_ops/OceanWavesBoundary.cpp
```diff
@@ -238,6 +264,10 @@ void OceanWavesBoundary::set_inflow_sibling_velocity(
 
         const int idir = ori.coordDir();
         const auto& domain_box = geom.Domain();
+
+        amrex::IntVect shift_to_interior = {0, 0, 0};
```
marchdf, line 268: no that this matters much but just in case, you can use `TheDimensionVector` or something like that to get a 1 in the `dir` location. You would still need to negate it after to get the `-1` with something like `shift_to_interior *= -1` if ori is low.

marchdf (commented):
Cool work!

hgopalan:
We should consider adding a walk through using Will's case in the future as this adds many different capabilities in a single solve.

mbkuhn:
> We should consider adding a walk through using Will's case. This PR adds many different capabilities in a single solve.

That's a good idea! I'm getting the capability in for the version release, but I think there will be more development and documentation as Matt and Will continue to look into this.

hgopalan (approved):

amr-wind/ocean_waves/boundary_ops/OceanWavesBoundary.cpp
```diff
@@ -73,6 +79,8 @@ void OceanWavesBoundary::set_velocity(
     const int nghost = 1;
     const auto& domain = geom.growPeriodicDomain(nghost);
 
+    const bool terrain_and_vof = m_terrain_exists && m_vof_exists;
```
moprak-nrel, line 82: really minor nit, can we have this be `terrain_and_vof_exists` or something similar?

moprak-nrel (commented):
Cool work! Have just a super minor comment

marchdf (approved):

moprak-nrel (approved):

PR #1297, opened 2024-10-15
Changes for coupling with ERF using MultiBlock

mukul1992:
## Summary

These changes allow coupling with ERF via the MultiBlock mechanism. It includes maintaining a pointer to the multiblock container which is defined in the coupling driver. For reading ERF data, it uses the ABL boundary plane mechanism by defining a `read_erf` function in the driver and passing it to a function pointer in AMR-Wind.

## Pull request type

Please check the type of change introduced:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe):

## Checklist

The following is included:

- [ ] new unit-test(s)
- [ ] new regression test(s)
- [ ] documentation for new capability

This PR was tested by running:

- the unit tests
  - [ ] on GPU 
  - [x] on CPU (MacOS M3 arch, GCC 13)
- the regression tests
  - [ ] on GPU 
  - [x] on CPU (MacOS M3 arch, GCC 13)

PR #1478, opened 2025-02-05
Changes for coupling with ERF using MultiBlock

mukul1992:
## Summary

Attempt 2 after the PR #1297 expired.
These changes allow coupling with ERF via the [MultiBlock driver](https://github.com/erf-model/erf-amrwind-driver). It includes maintaining a pointer to the multiblock container which is defined in the coupling driver. For reading ERF data, it uses the ABL boundary plane mechanism by defining a `read_erf` function in the driver and passing it to a function pointer in AMR-Wind.

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [x] on CPU (MacOS M3 arch, GCC 13)
- the regression tests
  - [ ] on GPU 
  - [x] on CPU (MacOS M3 arch, GCC 13)

## Additional background

I believe most issues highlighted in the last expired PR #1297 are now resolved, except the use of raw pointers. I did explore the use of unique_ptr, however it may not work due to how the pointer is being used. I am open to hearing suggestions.

### DIFF
--- a/amr-wind/CFDSim.H
+++ b/amr-wind/CFDSim.H
@@ -68,7 +68,7 @@ public:
     void set_mbc(MultiBlockContainer* mbc) { m_mbc = mbc; }
     MultiBlockContainer** mbc() { return &m_mbc; }
 
-    void set_read_erf(ReadERFFunction fcn) { m_read_erf = fcn; }
+    void set_read_erf(ReadERFFunction fcn) { m_read_erf = std::move(fcn); }
     ReadERFFunction* get_read_erf() { return &m_read_erf; }
 
     //! Return the field repository
--- a/amr-wind/CFDSim.H
+++ b/amr-wind/CFDSim.H
@@ -9,6 +9,10 @@
 #include "amr-wind/core/MeshMap.H"
 #include "amr-wind/helics.H"
 
+#ifdef ERF_USE_MULTIBLOCK
+class MultiBlockContainer;
+#endif
+
 /** AMR-Wind
  *
  *  All C++ code in AMR-Wind is organized within the amr_wind namespace.
@@ -61,6 +65,11 @@ public:
     //! Return simulation time control
     SimTime& time() { return m_time; }
     const SimTime& time() const { return m_time; }
+
+#ifdef ERF_USE_MULTIBLOCK
+    void set_mbc(MultiBlockContainer* mbc) {m_mbc = mbc;}
+    MultiBlockContainer** mbc() {return &m_mbc;}
+#endif
 
     //! Return the field repository
     FieldRepo& repo() { return m_repo; }
@@ -170,6 +179,10 @@ private:
 
     // State of solver - know if during an overset timestep or not
     bool m_during_overset_advance{false};
+
+#ifdef ERF_USE_MULTIBLOCK
+    MultiBlockContainer* m_mbc = nullptr;
+#endif
 };
 
 } // namespace amr_wind
--- a/amr-wind/CFDSim.H
+++ b/amr-wind/CFDSim.H
@@ -10,6 +10,7 @@
 #include "amr-wind/helics.H"
 
 #ifdef ERF_USE_MULTIBLOCK
+#include "amr-wind/wind_energy/ABLReadERFFcn.H"
 class MultiBlockContainer;
 #endif
 
@@ -67,8 +68,11 @@ public:
     const SimTime& time() const { return m_time; }
 
 #ifdef ERF_USE_MULTIBLOCK
-    void set_mbc(MultiBlockContainer* mbc) {m_mbc = mbc;}
-    MultiBlockContainer** mbc() {return &m_mbc;}
+    void set_mbc(MultiBlockContainer* mbc) { m_mbc = mbc; }
+    MultiBlockContainer** mbc() { return &m_mbc; }
+
+    void set_read_erf(ReadERFFcn fcn) { m_read_erf = fcn; }
+    ReadERFFcn* get_read_erf() { return &m_read_erf; }
 #endif
 
     //! Return the field repository
@@ -182,6 +186,7 @@ private:
 
 #ifdef ERF_USE_MULTIBLOCK
     MultiBlockContainer* m_mbc = nullptr;
+    ReadERFFcn m_read_erf {nullptr};
 #endif
 };
 
--- a/amr-wind/CFDSim.H
+++ b/amr-wind/CFDSim.H
@@ -9,7 +9,7 @@
 #include "amr-wind/core/MeshMap.H"
 #include "amr-wind/helics.H"
 
-#ifdef ERF_USE_MULTIBLOCK
+#ifdef ERF_AW_MULTIBLOCK
 #include "amr-wind/wind_energy/ABLReadERFFcn.H"
 class MultiBlockContainer;
 #endif
@@ -67,7 +67,7 @@ public:
     SimTime& time() { return m_time; }
     const SimTime& time() const { return m_time; }
 
-#ifdef ERF_USE_MULTIBLOCK
+#ifdef ERF_AW_MULTIBLOCK
     void set_mbc(MultiBlockContainer* mbc) { m_mbc = mbc; }
     MultiBlockContainer** mbc() { return &m_mbc; }
 
@@ -184,7 +184,7 @@ private:
     // State of solver - know if during an overset timestep or not
     bool m_during_overset_advance{false};
 
-#ifdef ERF_USE_MULTIBLOCK
+#ifdef ERF_AW_MULTIBLOCK
     MultiBlockContainer* m_mbc = nullptr;
     ReadERFFcn m_read_erf {nullptr};
 #endif
--- a/amr-wind/CFDSim.H
+++ b/amr-wind/CFDSim.H
@@ -9,10 +9,8 @@
 #include "amr-wind/core/MeshMap.H"
 #include "amr-wind/helics.H"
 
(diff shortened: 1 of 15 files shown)


marchdf:
@mukul1992 thanks for doing this! It looks like most of the code changes are cmake things. Can you comment on why those are needed? I added @jrood-nrel as a reviewer and some context would probably help him understand the changes.

amr-wind/wind_energy/ABLReadERFFcn.H
```diff
@@ -0,0 +1,17 @@
+#ifndef ABLREADERFFCN_H
```
marchdf, line 1: Can we make the name of this file and the ifdefs a bit more explicity? `ABLREADERFFUNCTION` maybe?

amr-wind/incflo.cpp
```diff
@@ -140,7 +146,7 @@ void incflo::prepare_for_time_integration()
         return;
     }
 
-    if (m_do_initial_proj || m_initial_iterations > 0) {
+    if (m_initial_iterations > 0) {
```
marchdf, line 149: Did you mean to remove the `do_initial_proj` check? If so, can you explain why? @mbkuhn might also have thoughts on that.

amr-wind/incflo.cpp
```diff
@@ -23,6 +23,12 @@ incflo::incflo()
     // constructor. No valid BoxArray and DistributionMapping have been defined.
     // But the arrays for them have been resized.
 
+#ifdef ERF_AMR_WIND_MULTIBLOCK
+    amrex::Print() << std::endl
+                   << "AMR-Wind is initializing.. " << std::endl
```
marchdf, line 28:
```suggestion
                   << "Initializing AMR-Wind." << std::endl
```

amr-wind/incflo.H
```diff
@@ -151,6 +160,11 @@ public:
 
     void ReadCheckpointFile();
 
+#ifdef ERF_AMR_WIND_MULTIBLOCK
+    void SetMultiBlockPointer(MultiBlockContainer* mbc) { m_sim.set_mbc(mbc); }
```
marchdf, line 164: All these raw pointers are making me a bit nervous. Do you think you could use `unique_ptr` here and in the other spots?

amr-wind/wind_energy/ABLBoundaryPlane.cpp
```diff
@@ -1062,6 +1102,19 @@ void ABLBoundaryPlane::read_file(const bool nph_target_time)
     const amrex::Real time =
         nph_target_time ? m_time.current_time() + 0.5 * m_time.delta_t()
                         : m_time.new_time();
+
+#ifdef ERF_AMR_WIND_MULTIBLOCK
+    if (m_out_fmt == "erf-multiblock") {
+        ReadERFFunction read_erf = *m_read_erf;
+        if (read_erf) {
```
marchdf, line 1109: the linters like to be explicit:
```suggestion
        if (read_erf != nullptr) {
```

amr-wind/CFDSim.H
```diff
@@ -9,6 +9,11 @@
 #include "amr-wind/core/MeshMap.H"
 #include "amr-wind/helics.H"
 
+#ifdef ERF_AMR_WIND_MULTIBLOCK
```
marchdf, line 12: I am wondering about the necessity of an additional compile time definition. We like to reduce these to reduce the potential build matrix. Do you think it would be worth considering removing this ifdef? Basically always building with this on?

mukul1992:
> @mukul1992 thanks for doing this! It looks like most of the code changes are cmake things. Can you comment on why those are needed? I added @jrood-nrel as a reviewer and some context would probably help him understand the changes.

So the two major changes related to cmake are:

amr-wind/incflo.cpp
mukul1992: Yes, we discussed this over slack a couple of months ago but the messages are now lost. I believe we had discussed how the initial projection should not require triggering the `prepare_boundaries()` here because we would only need the boundary values at time 0.

In context of the coupling, when AMR-Wind is doing this, the ERF solution has not yet advanced so it cannot provide the new time values. But it should also not be needed for AMR-Wind by itself.

Please let me know if this is incorrect and I can implement a work-around.

amr-wind/CFDSim.H
mukul1992: Yes, that would not cause any issues.

Our thought perhaps was to isolate the coupling related changes as currently it is kind of in an experimental phase, and especially as it has these raw pointers, function pointer etc. But I understand your concern too.

I can go ahead and remove these if you are okay with it. We can also discuss during the meeting tomorrow.

amr-wind/incflo.H
mukul1992: I did add an extra nullptr check in `ABLBoundaryPlane.H`.

jrood-nrel:
Does this really need an external amrex? I removed that option here https://github.com/Exawind/amr-wind/pull/1363

mukul1992:
> Does this really need an external amrex? I removed that option here #1363

The idea is to have a flat dependency structure where the driver code itself and ERF and AMR-Wind (and Hydro) depend on the same AMReX repo. Having them point to their own internal AMReX would create conflicts. Having this flat structure instead of just using AMReX from either the ERF or AMR-Wind submodule provides better control of what version of amrex we can check out, and also simplifies the submodule setup as we do not have to have recursive submodules. This setup has been working pretty well.

PXL_20240828_193445031 copy

mukul1992:
> Can you replace those fixme comments that are tripping the CI with just an abort if the conditions don't work out?

@marchdf I just pushed [these changes](https://github.com/Exawind/amr-wind/pull/1478/commits/bb84d3ed3b419c6b974afa0af43cf22a7571dad0). I removed the FIXME related to generalizing lev=0 as I think it is fair to keep this restricted to level 0 for now considering the complexity of coupling. I also removed the other FIXME, which was being carried over from older code and I did not fully understand. However the "native" boundary code for `read_header()` is identical so I think it should no longer be an issue. I also made the code section for "erf-multiblock" similar to that for "native" just to be consistent.

amr-wind/CFDSim.H
mukul1992: I've removed these.

amr-wind/incflo.cpp
mukul1992: @mbkuhn are you okay with this?

amr-wind/wind_energy/ABLBoundaryPlane.cpp
```diff
@@ -856,6 +865,45 @@ void ABLBoundaryPlane::read_header()
...
+                            (fld->bc_type()[ori] != BC::mass_inflow_outflow));
+                    })) {
+                continue;
+            }
+
+            m_in_data.define_plane(ori);
+
+            // restrict to level 0 for now for multiblock
+            const int lev = 0;
+
```
marchdf, line 896: do you want to add an abort if nlev > 0

marchdf:
Looks like you have a minor thing to fix in the linter. Do you have a regtest or unit test for this? It would be ideal though I get why it would be hard... 

I am running the tests with this branch on CPU/GPU.

jrood-nrel:
Does just putting `SetMultiBlockPointer` in https://github.com/Exawind/amr-wind/blob/38d1b9fd0b70aab4a01fd507f039750c2508bd1c/.clang-tidy#L76 solve the clang-tidy issue?

amr-wind/wind_energy/ABLBoundaryPlane.cpp
mukul1992: I was thinking that these lines downstream in `populate_data` would take care of it?
https://github.com/Exawind/amr-wind/blob/fa1e36535b1378c41d87444d532cb07cb12ef9ed/amr-wind/wind_energy/ABLBoundaryPlane.cpp#L1268-L1271 

Because technically we are okay with having multiple levels, we just don't want the finer levels to touch the boundary. Adding an Abort would prevent us from testing the former case in the medium term, and the latter case is taken care of by the above lines.

mukul1992:
> Does just putting `SetMultiBlockPointer` in
> 
> https://github.com/Exawind/amr-wind/blob/38d1b9fd0b70aab4a01fd507f039750c2508bd1c/.clang-tidy#L76
> solve the clang-tidy issue?

okay I just did that, that should certainly work..
I was trying to be more comprehensive by ignoring anything with "MultiBlock" in it, but messed things up because I misunderstood the syntax of the clang-tidy configuration file. I'll stick with the simpler solution for now.

mukul1992:
> Looks like you have a minor thing to fix in the linter. 
should be fixed now..

> Do you have a regtest or unit test for this? It would be ideal though I get why it would be hard...
Yeah that would be hard to come up with.. I can't think of a simple way to test the features, without actually coupling with ERF.

marchdf:
All tests are passing CPU and GPU.

marchdf (approved):

jrood-nrel (approved):

mukul1992:
Thank you for your efforts in reviewing the PR and providing suggestions. Are we ready to merge it?

mukul1992:
Great! Thanks again, and especially to @jmsexton03 for implementing the build system for the coupling driver, including the changes to AMR-Wind that were a part of this PR.

PR #1513, opened 2025-03-02
Enforce solvability AFTER multiplying rho0 for anelastic

mukul1992:
## Summary

This PR corrects the earlier code which enforced solvability for velocity before multiplying rho0 for the anelastic case. There are no diffs expected as there are no tests with both inflow-outflow BCs and anelastic, but I confirmed that such a case works as expected.

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [x] on CPU: macOS M3, g++-13
- the regression tests
  - [ ] on GPU 
  - [x] on CPU: macOS M3, g++-13

### DIFF
--- a/amr-wind/projection/incflo_apply_nodal_projection.cpp
+++ b/amr-wind/projection/incflo_apply_nodal_projection.cpp
@@ -320,12 +320,8 @@ void incflo::ApplyProjection(
 
     // Need to apply custom Neumann funcs for inflow-outflow BC
     // after setting the inflow vels above
-    // and then enforce solvability by matching outflow to inflow.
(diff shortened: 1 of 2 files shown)


marchdf:
Can you add that case to the reg tests? Seems like a valuable thing to have on hand.

mukul1992:
> Can you add that case to the reg tests? Seems like a valuable thing to have on hand.

Okay, I just changed the existing inflow-outflow ABL test to anelastic instead of adding a new test.

marchdf (approved):

PR #1528, opened 2025-03-13
make mass_inflow_outflow compatible with VOF

mbkuhn:
## Summary

The mass_inflow_outflow branch was previously omitted from some of the conditionals in VOF BCs. This PR includes this BC type.

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [x] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [ ] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU 

## Additional background

Issue Number:


mbkuhn:
@marchdf I'd like your thoughts on the reg test for this PR. I'm trying to add a case that is a version of abl_multiphase_laminar with a mass_inflow_outflow BC. Because of the mass balance that takes place in the mass_inflow_outflow, I have to have 2 of those, not just one, so I would end up using xlo and xhi. That would require adding "xhi" to the abl_multiphase_laminar case because that one generates the boundary planes. But then, there would be two planes in the boundary files, which removes the coverage that we had for a single plane (which led to this bug fix: https://github.com/Exawind/amr-wind/pull/1436). My other thought, if I wanted to leave it at a single plane in the files, was to specify a constant, uniform velocity for one of the inflow_outflow planes and not need to read it. But for that to work, then the check from the other PR would need to see which planes are specified as boundary planes in the input file and not just check the BC type.

mbkuhn:
I followed Marc's advice on the reg test, but the resulting reg test still does not run. There is an issue with the mass_inflow_outflow BC (direction-dependent within AMReX-Hydro). Because the current implementation uses the level 0 domain indices to check the domain boundaries at every level, in this reg test it only detects the inflow boundary on lev = 1 (because that is the same on every level, i.e., i = -1) and does not detect the outflow boundary on lev = 1 (because the indices don't line up).

See https://github.com/AMReX-Fluids/AMReX-Hydro/pull/164
This change (first commit) makes the simulation run because it can actually find the outflow. However, I am not sure that it is correct because it seems to check inflows against outflows on the same level, which doesn't make sense. It should be allowed to have an inflow on level = 1 and no outflow on level = 1, provided there is still outflow on level = 0. I have to look into it more.

marchdf (approved):

Issue #1462, opened 2025-01-27
Handing Turbulence in Multiphase Flow


hgopalan:
The ABL setup for multiphase flow is setup to run under laminar model and this needs to be modified in the future. There are three components which needs to be consistently setup to avoid oscillation and unrealistic values of turbulent viscosity at the air-water interface. 

1. Initial velocity profile in air and water: This can generate initial gradients vertically and with the density ratio, can cause the turbulent viscosity to grow.
2. Density difference: The difference in density between air and water causes unrealistic turbulent viscosity growth at interface and can also result in thermal oscillations in the water layer. 
3. Wave energy production: The movement of waves can generate velocity gradients, which are a solid body type production term and should not generate turbulent viscosity. This is more complicated for a linear model and may require non-linear models to handle. 

I tried to address issues 2. and partially 3. using a modification to the Smagorinsky model. We replace 

Shear Production = Dissipation 

by Shear Production + Density Gradient Production = Dissipation and recompute the turbulent viscosity. Initial results are shown below for the abl_multiphase_laminar flow case with Smagorinsky model replacing laminar model. The modification is still in the testing phase and I am hoping to have more ideas and discussion for implementation. 

Wind Speed 

<img width="1728" alt="Image" src="https://github.com/user-attachments/assets/7865c916-e4d0-4074-b982-21dd2dc96c93" />

Turbulent Viscosity 

<img width="1728" alt="Image" src="https://github.com/user-attachments/assets/3035171a-db0e-4921-935e-6a739b252cb8" />

Turbulence has been currently turned off in the water since the initial shear between air and water profile generates vertical winds which causes oscillations in the temperature in the water region.

mbkuhn:
linking #987 because that was an early attempt

PR #1530, opened 2025-03-17
Kosovic Model - Multiphase Support 

hgopalan:
## Summary

The existing LES models  overpredict  turbulent viscosity near the interface for multiphase flow simulations. This arises due to the inconsistency between the ABL profile and the profile specified in the ocean. To enable turbulence in the ocean, there has to be a consistent initial profile between atmosphere and ocean to avoid over-prediction of the turbulence diffusion terms due to initial condition generated shear and buoyancy production. 

The multi-phase model handles this incompatibility and ensures that the diffusion term do not cause a drop in the time step. The following changes are made 

1. A new density production term is added and added to the calculation of the turbulent viscosity. This term ensures that the turbulent viscosity is driven to zero at the interface where there is a vertical density gradient. 
2.  A horizontal damping term to turn off turbulent viscosity within the strong wave region. Wave motion is close to solid body motion and we cannot have the turbulence diffusion term inside it. 
3.  A water density term to turn off the turbulence diffusion in the water region. Currently a uniform current speed and temperature value is specified in the water region. This causes inconsistency in the diffusion term at the interface. This term ensures that the inconsistency does not cause solver divergence or reduction in time step.

The multiphase solver runs at 1/3rd of the time step when compared to the single phase immersed forcing wave model.  The maximum possible Aspect Ratio is ~ 2-3. Solver fails at AR=4 with JONSWAP  spectrum. 

- [ X] Bugfix
- [X ] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [ ] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU 

## Additional background

Linear Waves 

Streamwise (1)

Spanwise (1)

Turbulence (1)

Issue Number: #1462


hgopalan:
@rybchuk  Thank you for the test cases. It was very helpful.

hgopalan:
Adding the changes to the Kosovic model would make the code too cluttered. This was the reason for a separate model.

hgopalan:
Added test case

hgopalan:
documentation done

PR #1538, opened 2025-03-26
Check density bounds upon initialization - for constant density single-phase and two-phase

mbkuhn:
## Summary

After a fillpatch has been applied in post_init, this density check evaluates the maximum and minimum density in the domain, including the ghost cells. If the simulation is supposed to be constant density, but the minimum and the maximum densities do not match, then the code aborts. This works in the case that the BC density does not match the density from the input file, or the simulation restarts from a checkpoint file with a different density. For two-phase cases, this checks the max and min of the density in cells that are fully liquid or fully gas.

This PR also changes some reg tests because of this issue.

## Pull request type

- [x] Bugfix
- [x] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [ ] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU 

## Additional background

Issue Number:

### DIFF
--- a/amr-wind/utilities/diagnostics.cpp
+++ b/amr-wind/utilities/diagnostics.cpp
@@ -4,88 +4,6 @@
 
 using namespace amrex;
 
-void amr_wind::diagnostics::get_mfab_extrema(
-    amrex::Real& mfab_max_val,
-    amrex::Real& mfab_min_val,
-    const amrex::MultiFab& mfab,
-    const int nghost)
-{
-    int ncomp = mfab.n_comp;
-    mfab_max_val = amrex::ReduceMax(
-        mfab, nghost,
-        [=] AMREX_GPU_HOST_DEVICE(
-            amrex::Box const& bx,
-            amrex::Array4<amrex::Real const> const& f_arr) -> amrex::Real {
-            amrex::Real max_fab = -1e8;
-            amrex::Loop(bx, [=, &max_fab](int i, int j, int k) noexcept {
-                for (int n = 0; n < ncomp; ++n) {
-                    max_fab = amrex::max(max_fab, f_arr(i, j, k, n));
-                }
-            });
-            return max_fab;
-        });
-
-    mfab_min_val = amrex::ReduceMin(
-        mfab, nghost,
-        [=] AMREX_GPU_HOST_DEVICE(
-            amrex::Box const& bx,
-            amrex::Array4<amrex::Real const> const& f_arr) -> amrex::Real {
-            amrex::Real min_fab = 1e8;
-            amrex::Loop(bx, [=, &min_fab](int i, int j, int k) noexcept {
-                for (int n = 0; n < ncomp; ++n) {
-                    min_fab = amrex::min(min_fab, f_arr(i, j, k, n));
-                }
-            });
-            return min_fab;
-        });
-}
-
-void amr_wind::diagnostics::get_mfab_extrema(
-    amrex::Real& mfab_max_val,
-    amrex::Real& mfab_min_val,
-    const amrex::MultiFab& mfab,
-    const amrex::MultiFab& mfab_mask,
-    const amrex::Real mask_val,
-    const int nghost)
-{
-    int ncomp = mfab.n_comp;
-    mfab_max_val = amrex::ReduceMax(
-        mfab, mfab_mask, nghost,
-        [=] AMREX_GPU_HOST_DEVICE(
-            amrex::Box const& bx, amrex::Array4<amrex::Real const> const& f_arr,
-            amrex::Array4<amrex::Real const> const& mask_arr) -> amrex::Real {
-            amrex::Real max_fab = -1e8;
-            amrex::Loop(bx, [=, &max_fab](int i, int j, int k) noexcept {
-                for (int n = 0; n < ncomp; ++n) {
-                    max_fab = amrex::max(
-                        max_fab, std::abs(mask_arr(i, j, k) - mask_val) <
-                                         constants::TIGHT_TOL
-                                     ? f_arr(i, j, k, n)
-                                     : max_fab);
-                }
-            });
-            return max_fab;
-        });
-
-    mfab_min_val = amrex::ReduceMin(
-        mfab, mfab_mask, nghost,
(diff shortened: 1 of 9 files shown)


amr-wind/utilities/diagnostics.cpp
```diff
@@ -1,8 +1,148 @@
...
+#include "constants.H"
 
 using namespace amrex;
 
+void amr_wind::diagnostics::get_mfab_extrema(
+    amrex::Real& mfab_max_val,
+    amrex::Real& mfab_min_val,
+    const amrex::MultiFab& mfab,
+    const int nghost)
+{
```
marchdf, line 12: Could we use MultiFab min() and max()? Or even BaseFab::minmax()? And then for the masked version we could do a Multiply to get the mfab to do the reduction on?

amr-wind/utilities/diagnostics.cpp
```diff
@@ -1,8 +1,148 @@
...
+    int ncomp = mfab.n_comp;
+    mfab_max_val = amrex::ReduceMax(
+        mfab, nghost,
+        [=] AMREX_GPU_HOST_DEVICE(
+            amrex::Box const& bx,
+            amrex::Array4<amrex::Real const> const& f_arr) -> amrex::Real {
+            amrex::Real max_fab = -1e8;
```
marchdf, line 19: I think you can use `LOW_NUM` from the constants.

amr-wind/utilities/diagnostics.cpp
```diff
@@ -1,8 +1,148 @@
...
+            });
+            return max_fab;
+        });
+
+    mfab_min_val = amrex::ReduceMin(
+        mfab, mfab_mask, nghost,
+        [=] AMREX_GPU_HOST_DEVICE(
+            amrex::Box const& bx, amrex::Array4<amrex::Real const> const& f_arr,
+            amrex::Array4<amrex::Real const> const& mask_arr) -> amrex::Real {
+            amrex::Real min_fab = 1e8;
```
marchdf, line 75: and `LARGE_NUM` here.

amr-wind/utilities/diagnostics.cpp
```diff
@@ -1,8 +1,148 @@
...
+void amr_wind::diagnostics::get_field_extrema(
+    amrex::Real& field_max_val,
+    amrex::Real& field_min_val,
+    const amr_wind::Field& field,
+    const bool include_ghosts)
+{
+    const int finest_level = field.repo().num_active_levels() - 1;
+
+    // Initial values
+    amrex::Real max_val_lev{-1e8};
```
marchdf, line 98: same for these values.

marchdf (changes requested):

marchdf:
I think you also need to fix some other tests:
`ow_current_bathymetry` and `ow_w2a_nonperiodic` show this:
```
Creating MomentumSource instance: GravityForcing
Creating MomentumSource instance: DragForcing
  what():  Density check failed. Gas density maximum is too different from gas density minimum.
rho_g_max = 1.225000, rho_g_min = 1.000000
SIGABRT
```

mbkuhn:
@marchdf let me know what you think of the new approach -- if it's good, I'll take out the mfab_extrema functions

amr-wind/utilities/diagnostics.cpp
```diff
@@ -1,8 +1,196 @@
...
+                                       ? 1.0
+                                       : set_val;
+        });
+}
+
+void amr_wind::diagnostics::get_field_extrema(
+    amrex::Real& field_max_val,
+    amrex::Real& field_min_val,
+    const amr_wind::Field& field,
+    const int icomp,
```
marchdf, line 111: Only thing I would change, just to be consistent with other amrex-like things is to specify the start and end component instead of having logic on `icomp`.

marchdf:
New way is great! Thanks! Just had a minor comment.

amr-wind/utilities/diagnostics.cpp
mbkuhn: good idea. I think I'll do the starting component and the number of components, like in the MultiFab Multiply function

amr-wind/utilities/diagnostics.cpp
```diff
@@ -1,8 +1,109 @@
...
+            max_val_lev =
+                field(lev).max(n, include_ghosts ? field.num_grow()[n] : 0);
+            min_val_lev =
+                field(lev).min(n, include_ghosts ? field.num_grow()[n] : 0);
+            field_min_val = amrex::min(field_min_val, min_val_lev);
+            field_max_val = amrex::max(field_max_val, max_val_lev);
+        }
+    }
+
+    // Do additional parallelism stuff
```
marchdf, line 51:
```suggestion
```

amr-wind/utilities/diagnostics.cpp
```diff
@@ -1,8 +1,109 @@
...
+    const amr_wind::Field& field_mask,
+    const amrex::Real mask_val,
+    const int comp,
+    const int ncomp,
+    const bool include_ghosts)
+
+{
+    const int finest_level = field.repo().num_active_levels() - 1;
+
+    // Initial values
```
marchdf, line 69:
```suggestion
```

amr-wind/utilities/diagnostics.cpp
```diff
@@ -1,8 +1,109 @@
...
+            max_val_lev = mask_multiplier_max.max(
+                n, include_ghosts ? field.num_grow()[n] : 0);
+            min_val_lev = mask_multiplier_min.min(
+                n, include_ghosts ? field.num_grow()[n] : 0);
+            field_min_val = amrex::min(field_min_val, min_val_lev);
+            field_max_val = amrex::max(field_max_val, max_val_lev);
+        }
+    }
+
+    // Do additional parallelism stuff
```
marchdf, line 102:
```suggestion
```

mbkuhn:
whoops, changing it now

marchdf:
Looks good now!

marchdf (approved):

PR #1539, opened 2025-03-27
Correcting ConvectingTaylorVortex error calculation for the hybrid solver

itopcuoglu:


amr-wind/physics/ConvectingTaylorVortex.cpp
```diff
@@ -296,12 +297,31 @@ amrex::Real ConvectingTaylorVortex::compute_error(const Field& field)
                 return cell_vol * mask_bx(i, j, k) * (u - u_exact) *
                        (u - u_exact);
             });
+
+        amrex::Real error_vol;
+        error_vol += amrex::ParReduce(
+            amrex::TypeList<amrex::ReduceOpSum>{},
+            amrex::TypeList<amrex::Real>{}, fld, amrex::IntVect(0),
+            [=] AMREX_GPU_DEVICE(int box_no, int i, int j, int k)
```
marchdf, line 305: I don't think you need a seperate loop here. you can use the one above but have the return type be `amrex::GpuTuple<amrex::Real, amrex::Real>` and then `return {cell_vol * mask_bx(i, j, k) * (u - u_exact) * (u - u_exact), cell_vol * mask_bx(i, j, k)};`

Does that make sense?

marchdf (changes requested):

amr-wind/physics/ConvectingTaylorVortex.cpp
itopcuoglu: Is there a way to use the += operator when you return a GpuTuple?

marchdf: no but you can store to a temp: `const auto vals = ...` and then `error_vol += vals[0] etc`

itopcuoglu: I converted the return type to GpuTuple, and extracted the tuple elements with `amrex::get<>()`.

marchdf: well done!

marchdf:
Nice work! Does it actually work in overset world now?

marchdf (approved):

PR #1498, opened 2025-02-20
Use multiply_rho true for potential temperature evolution

marchdf:
## Summary

Use multiply_rho true for potential temperature evolution and adapt for multiphase sims.

## Pull request type

Please check the type of change introduced:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe):

## Checklist

The following is included:

- [ ] new unit-test(s)
- [ ] new regression test(s)
- [ ] documentation for new capability

This PR was tested by running:

- the unit tests
  - [ ] on GPU 
  - [ ] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU 

Closes #1493

PR #1541, opened 2025-03-27
Draft PR to fix issues with multiply_rho

mpolimeno:
## Summary

Continuing the work started [here](https://github.com/Exawind/amr-wind/pull/1498) to fix the impact that multiply_rho has on some regression tests

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [ ] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU 

## Additional background

Issue Number:

### DIFF
--- a/amr-wind/equation_systems/DiffusionOps.H
+++ b/amr-wind/equation_systems/DiffusionOps.H
@@ -149,10 +149,33 @@ struct DiffusionOp<
         auto tau_state = std::is_same<Scheme, fvm::Godunov>::value
                              ? FieldState::New
                              : fstate;
+
         amrex::MLMG mlmg(*this->m_applier);
         mlmg.apply(
             this->m_pdefields.diff_term.state(tau_state).vec_ptrs(),
             this->m_pdefields.field.vec_ptrs());
+
+        const auto& repo = this->m_pdefields.repo;
+        const int nlevels = repo.num_active_levels();
+        auto& divtau = this->m_pdefields.diff_term.state(tau_state);
+        auto& m_density = this->m_pdefields.repo.get_field("density");
+        const auto& density = m_density.state(FieldState::NPH);
+        if (!PDE::multiply_rho) {
+            for (int lev = 0; lev < nlevels; ++lev) {
+                const auto& divtau_arrs = divtau(lev).arrays();
+                const auto& rho_arrs = density(lev).const_arrays();
+
+                amrex::ParallelFor(
+                    divtau(lev), [=] AMREX_GPU_DEVICE(
+                                     int nbx, int i, int j, int k) noexcept {
+                        amrex::Real rhoinv = 1.0 / rho_arrs[nbx](i, j, k);
+                        divtau_arrs[nbx](i, j, k, 0) *= rhoinv;
+                        divtau_arrs[nbx](i, j, k, 1) *= rhoinv;
+                        divtau_arrs[nbx](i, j, k, 2) *= rhoinv;
+                    });
+            }
+            amrex::Gpu::streamSynchronize();
+        }
     }
 };
 


mpolimeno:
Will need to tweak a few more things in the code still, but getting there, @mbkuhn.

Will check the reg tests with bigger-than-expected diffs listed [here](https://github.com/Exawind/amr-wind/pull/1498)

amr-wind/equation_systems/DiffusionOps.H
```diff
@@ -149,10 +149,33 @@ struct DiffusionOp<
...
+        auto& divtau = this->m_pdefields.diff_term.state(tau_state);
+        auto& m_density = this->m_pdefields.repo.get_field("density");
+        const auto& density = m_density.state(FieldState::NPH);
+        if (!PDE::multiply_rho) {
+            for (int lev = 0; lev < nlevels; ++lev) {
+                const auto& divtau_arrs = divtau(lev).arrays();
+                const auto& rho_arrs = density(lev).const_arrays();
+
+                amrex::ParallelFor(
+                    divtau(lev), [=] AMREX_GPU_DEVICE(
```
marchdf, line 169: I think it would be worth making this a component loop as well.`divtau.nComp()` and then `divtau_arrs[nbx](i, j, k, n)`.

mpolimeno: Yeah, absolutely.

mpolimeno:
[fcompare_out_diff_type_one.txt](https://github.com/user-attachments/files/19509963/fcompare_out_diff_type_one.txt)

Attaching here the diffs for abl_stable reg test between this PR and @marchdf [PR](https://github.com/Exawind/amr-wind/pull/1498), using `incflo.diffusion_type=1` and the default branches' options for `multiply_rho`

marchdf:
that's awesome. exactly what we wanted to see. Thanks!

amr-wind/equation_systems/DiffusionOps.H
```diff
@@ -149,10 +149,32 @@ struct DiffusionOp<
...
+        if (!PDE::multiply_rho) {
+            for (int lev = 0; lev < nlevels; ++lev) {
+                const auto& divtau_arrs = divtau(lev).arrays();
+                const auto& rho_arrs = density(lev).const_arrays();
+
+                amrex::ParallelFor(
+                    divtau(lev), amrex::IntVect(0), divtau.num_comp(),
+                    [=] AMREX_GPU_DEVICE(
+                        int nbx, int i, int j, int k, int n) noexcept {
+                        amrex::Real rhoinv = 1.0 / rho_arrs[nbx](i, j, k);
```
marchdf, line 172:
```suggestion
                        const amrex::Real rhoinv = 1.0 / rho_arrs[nbx](i, j, k);
```

amr-wind/equation_systems/DiffusionOps.H
```diff
@@ -149,10 +149,32 @@ struct DiffusionOp<
...
+        auto& divtau = this->m_pdefields.diff_term.state(tau_state);
+        const auto& density =
+            this->m_pdefields.repo.get_field("density").state(FieldState::NPH);
+        if (!PDE::multiply_rho) {
+            for (int lev = 0; lev < nlevels; ++lev) {
+                const auto& divtau_arrs = divtau(lev).arrays();
+                const auto& rho_arrs = density(lev).const_arrays();
+
+                amrex::ParallelFor(
+                    divtau(lev), amrex::IntVect(0), divtau.num_comp(),
```
marchdf, line 169: What about: 
```suggestion
                    divtau(lev), divtau.num_grow(), divtau.num_comp(),
```

mbkuhn:
Just to add context from the equations:
From the transport models, the diffusion constants are always dynamic quantities, i.e. mu, rho*alpha, etc. despite what the names may be called (mu is correct in the code, but the other quantities are called diffusivity, which is a bit of a misnomer) This approach is correct, but it means that simply calculating the explicit diffusion term (div_tau) does not make the term ready to be added to the equation.

If the PDE has multiply_rho = true, then nothing needs to be done (i.e., for Navier-Stokes, the momentum equation has div(tau) on the RHS, which ends up being mu * lap(u) for constant properties). But if the PDE has multiply_rho = false, then the diffusion term needs to be divided by density, because in this scenario the density is not included in the temporal derivative and the entire RHS has to be divided by it (i.e. for Navier-Stokes, the velocity equation has 1/rho * div(tau) on the RHS, which ends up being nu * lap(u) for constant properties).

The other terms on the RHS (i.e. source terms) have their own ways of dealing with this, which is a bit confusing when you think about it. The source terms are formulated from the perspective of multiply_rho = false, and then they are multiplied by rho afterwards if multiply_rho = true.

mbkuhn:
for dividing by rho in this PR, we know that for the godunov scheme this should be a division by rho^(n+1/2). but for MOL, that state might not be available, so that may be an issue when we look at those reg tests. There are ways to get around the issue if that comes up.

mbkuhn:
@marchdf @moprak-nrel Should we try to get this merged soon? I think we will want to look at the reg tests. We should see the same ones fail as the other multiply_rho PR

marchdf:
I am all for merging this

marchdf (approved):

moprak-nrel (approved):

PR #1548, opened 2025-04-01
First commit

hgopalan:
Adds the drag force for two-phase solver based on VOF. 

- [ ] Bugfix
- [X ] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [X ] new regression test(s)

### DIFF
--- a/amr-wind/equation_systems/icns/source_terms/CMakeLists.txt
+++ b/amr-wind/equation_systems/icns/source_terms/CMakeLists.txt
@@ -17,4 +17,5 @@ target_sources(${amr_wind_lib_name} PRIVATE
   DragForcing.cpp
   ForestForcing.cpp
   VelocityFreeAtmosphereForcing.cpp
+  MultiphaseDragForcing.cpp
   )
--- a/amr-wind/equation_systems/icns/source_terms/MultiphaseDragForcing.H
+++ b/amr-wind/equation_systems/icns/source_terms/MultiphaseDragForcing.H
@@ -1,0 +1,44 @@
+#ifndef MULTIPHASEDRAGFORCING_H
+#define MULTIPHASEDRAGFORCING_H
+
+#include "amr-wind/equation_systems/icns/MomentumSource.H"
+#include "amr-wind/core/SimTime.H"
+#include "amr-wind/CFDSim.H"
+
+namespace amr_wind::pde::icns {
+
+/** Adds the forcing term to include the presence of immersed boundary for multiphase flows
+ *
+ *  \ingroup icns_src
+ *
+ *
+ */
+class MultiphaseDragForcing
+    : public MomentumSource::Register<MultiphaseDragForcing>
+{
+public:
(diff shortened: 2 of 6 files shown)


mbkuhn:
This is a good start, but it's going to need some work. Mainly just to get rid of some of the stuff that was assumed in terrain drag that doesn't apply here. For example, in the terrain drag / terrain_is_waves context, the target levelset and velocity fields are the actual wave quantities enforced globally. But in the multiphase context, those fields are only valid in the wave forcing region (near the inlet). To do something globally, it should be based on the actual velocity and the vof field. I can get into the code and start making some of these changes, if that's ok with you.

mbkuhn:
I have a more big-picture question. This change adds the form drag induced by the waves. Since we are representing the waves on the mesh, are we adding this twice? The solver should naturally produce the drag from the waves.

I know I incorporated this from Manuel's work into the terrain approach, but I'm having my doubts about that as well. Not about the validity of Manuel's approach, but his is typically applied as a BC to a domain that has a flat bottom boundary. We actually have the waves in the domain, so this would seem redundant. Like if you were solving for the flow around a sphere and you also added the theoretical sphere drag to the flow.

hgopalan:
It depends on the wave amplitude and the refinement used. I assumed we need 6-12 points to represent the wave amplitude/length. This term would account for cases in which we do not sufficiently resolve the wave amplitude. If we have sufficient points to resolve the wave amplitude, we can ignore the form drag portion.

mbkuhn:
as a result of our discussion of the form drag and the merging of #1558, we should remove the form drag from this implementation. The viscous drag is still good to have, but I think in the case of multiphase simulations, there is no scenario where adding the analytical form drag makes sense.

I'll go ahead and take it out.

mbkuhn:
@hgopalan could you remind me of the difference between Dxz, Dyz and bc_forcing_x, bc_forcing_y?

hgopalan:
Dxz adds the shear stress at the surface while the bc forcing forces the wind speed at the computation point to approach the log-law.

mbkuhn:
I think the implementation is set up now. I haven't done much investigation into how things look in a simulation, though. I'm noticing that the linear solvers are struggling to converge in the reg test. The same thing happens in the ordinary ow_linear reg test, though. I'll try to figure that out separately to make sure that the drag model isn't introducing any problems.

mbkuhn:
> I think the implementation is set up now. I haven't done much investigation into how things look in a simulation, though. I'm noticing that the linear solvers are struggling to converge in the reg test. The same thing happens in the ordinary ow_linear reg test, though. I'll try to figure that out separately to make sure that the drag model isn't introducing any problems.

Figured out that I needed to specify a lower absolute tolerance for the mac projection, now no convergence issues. Feel free to check this and try it out, @hgopalan.

hgopalan:
Thank you for doing this.

PR #1549, opened 2025-04-01
F/superellipsebody

gantech:
## Summary

This is a draft pull request meant to initiate a discussion on if and how best to integrate this capability into AMR-Wind. Creates a capability to stop the flow inside a super ellipse defined by 

$$
\left (\frac{(x-x_{loc})}{x_c} \right )^6 + \left (\frac{(y-y_{loc})}{y_c} \right)^6 + \left (\frac{(z-z_{loc})}{z_c} \right)^6 = 1.
$$

Uses the terrain blanking approach and updates the same field name for lack of a better solution. Thanks to @hgopalan for help with this feature.

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [ ] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU 

## Additional background

This image shows the feature in action.

image

Issue Number:

### DIFF
--- a/amr-wind/physics/SuperEllipseBody.cpp
+++ b/amr-wind/physics/SuperEllipseBody.cpp
@@ -31,6 +31,7 @@ SuperEllipseBody::SuperEllipseBody(CFDSim& sim)
     , m_repo(sim.repo())
     , m_mesh(sim.mesh())
     , m_body_blank(sim.repo().declare_int_field("terrain_blank", 1, 1, 1))
+    , m_terrain_drag(sim.repo().declare_int_field("terrain_drag", 1, 1, 1))
 {
 
     amrex::ParmParse pp(identifier());
@@ -38,7 +39,10 @@ SuperEllipseBody::SuperEllipseBody(CFDSim& sim)
     m_dim = parse_vector(pp, "dimensions");
 
     m_sim.io_manager().register_output_int_var("terrain_blank");
+    m_sim.io_manager().register_output_int_var("terrain_drag");
     m_body_blank.setVal(0.0);
+    m_terrain_drag.setVal(0.0);
+
 }
 
 void SuperEllipseBody::initialize_fields(int level, const amrex::Geometry& geom)
--- a/amr-wind/physics/SuperEllipseBody.cpp
+++ b/amr-wind/physics/SuperEllipseBody.cpp
@@ -1,0 +1,230 @@
+#include "amr-wind/physics/SuperEllipseBody.H"
+#include "amr-wind/CFDSim.H"
+#include "AMReX_iMultiFab.H"
+#include "AMReX_MultiFabUtil.H"
+#include "AMReX_ParmParse.H"
+#include "AMReX_ParReduce.H"
+#include "amr-wind/utilities/trig_ops.H"
+#include "amr-wind/utilities/IOManager.H"
+#include "amr-wind/utilities/io_utils.H"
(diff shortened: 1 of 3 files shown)


hgopalan:
Glad it worked. One thing we can do is also add a field refinement and test to make sure we can capture the sharper interface, if necessary. We have an option of adding field refinement using terrain blank as follows: 

tagging.labels                                     = mast1terrain
tagging.mast1terrain.type                          = FieldRefinement
tagging.mast1terrain.field_name                    = terrain_blank
tagging.mast1terrain.grad_error                    = 0.1 0.1 0.1 0.1 0.1 0.1
tagging.mast1terrain.box_lo                        = -500 -500 361.674
tagging.mast1terrain.box_hi                        = 500 500 461.674

This will refine the interface upto amr.max_level while the regions internal and external will still be coarser levels and you can capture the sharp flow separation features. One point to ponder is if we should define the function as a separate routine. In this way if functions need to be added in future, it can use your framework while just adding a new function.

amr-wind/physics/SuperEllipseBody.cpp
```diff
@@ -0,0 +1,102 @@
...
+#include "amr-wind/utilities/trig_ops.H"
+#include "amr-wind/utilities/IOManager.H"
+#include "amr-wind/utilities/io_utils.H"
+#include "amr-wind/utilities/linear_interpolation.H"
+
+namespace amr_wind::superellipsebody {
+
+namespace {
+
+//! Utility function to parse inputs and return a vector instance
```
marchdf, line 16: We already have this in BoxRefiner. I would say let's remove this and then move the other one to a utility or something.

amr-wind/physics/SuperEllipseBody.cpp
```diff
@@ -0,0 +1,102 @@
...
+    amrex::ParmParse pp(identifier());
+    pp.query("body_file", m_body_file);
+    m_dim = parse_vector(pp, "dimensions");
+
+    m_sim.io_manager().register_output_int_var("terrain_blank");
+    m_sim.io_manager().register_output_int_var("terrain_drag");
+    m_sim.io_manager().register_io_var("terrainz0");
+    m_body_blank.setVal(0.0);
+    m_terrain_drag.setVal(0.0);
+    m_terrainz0.setVal(1e-4);
```
marchdf, line 47: Is 1e-4 meaningful or just meant to be a small number? If the latter, let's use something in `constants`

amr-wind/physics/SuperEllipseBody.cpp
```diff
@@ -0,0 +1,102 @@
...
+             >> m_orient.zz();
+    }
+    file.close();
+
+    const auto& dx = geom.CellSizeArray();
+    const auto& prob_lo = geom.ProbLoArray();
+    auto& blanking = m_body_blank(level);
+    auto levelBlanking = blanking.arrays();
+
+    vs::Vector dim = m_dim;
```
marchdf, line 73: you can const these three.

marchdf (changes requested):

marchdf:
Thanks for putting this together! This may be obvious for you but what's the motivation for this? Is this for a tower representation? Did you not want to use the normal IB pathway? It feels more natural than hacking the terrain blanking.

PR #1569, opened 2025-04-16
Use fcompare in CI

jrood-nrel:
This adds checks by comparing gold files generated from the `main` branch against results from each PR change.


jrood-nrel:
Would be nice to only be warned if the reference tests fail due to diffs, but there seems to be no straightforward way to allow failures and still see the step failed: https://github.com/orgs/community/discussions/15452 . This feature request has been ignored for 5 years so far.

marchdf (approved):

PR #1584, opened 2025-04-18
Fix RK2 approach for explicit diffusion

mbkuhn:
## Summary

Looks like there is a bug in the RK2 formulation for explicit diffusion. Does not account properly for density, which is a problem for equations that have multiply_rho = true, and does not account for masking, which is a problem for overset simulations.

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [ ] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU 

## Additional background

This is only a problem for diffusion_type = 0, i.e., fully explicit.

Issue Number:

### DIFF
--- a/amr-wind/equation_systems/CompRHSOps.H
+++ b/amr-wind/equation_systems/CompRHSOps.H
@@ -262,6 +262,42 @@ struct ComputeRHSOp
         amrex::Gpu::streamSynchronize();
     }
 
+    void improve_explicit_diff(const amrex::Real dt)
+    {
+        explicit_RK2_post_hoc(dt);
+    }
+
+    void explicit_RK2_post_hoc(const amrex::Real dt)
+    {
+        // divtau must be the difference between new and old before calling this
+        const auto& d_divtau = fields.diff_term;
+
+        auto& dof = fields.field;
+        const auto& repo = fields.repo;
+        const auto& mask_cell = repo.get_int_field("mask_cell");
+
+        const int nlevels = repo.num_active_levels();
+        for (int lev = 0; lev < nlevels; ++lev) {
+            auto f_arrs = dof(lev).arrays();
+            const auto& d_diff_arrs = d_divtau(lev).const_arrays();
+            const auto& mask_arrs = mask_cell(lev).const_arrays();
+            const auto& rho_arrs = density(lev).const_arrays();
+
+            amrex::ParallelFor(
+                dof(lev), amrex::IntVect(0), dof.num_comp(),
+                [=] AMREX_GPU_DEVICE(
+                    int nbx, int i, int j, int k, int n) noexcept {
+                    auto factor =
+                        0.5 * dt * (amrex::Real)mask_arrs[nbx](i, j, k);
+                    if (PDE::multiply_rho) {
+                        factor /= rho_arrs[nbx](i, j, k);
+                    }
+                    f_arrs[nbx](i, j, k, n) +=
+                        factor * d_diff_arrs[nbx](i, j, k, n);
+                });
+        }
+    }
+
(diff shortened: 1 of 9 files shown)


mbkuhn:
Here is a writeup on the problem.
[ExplicitDiffusionRK2_AMRWind_bug.docx](https://github.com/user-attachments/files/19816743/ExplicitDiffusionRK2_AMRWind_bug.docx)

TL;DR:
The RK2 approach tries to subtract and add diffusion terms from the PDE field itself. However, when multiply_rho = true, the diffusion term is not what directly modifies the PDE field, it's actually the diffusion term divided by density.

mbkuhn:
The first commit in the PR replaces the explicit diffusion approach for only the scalar equations. Since (currently) multiply_rho = false for temperature, I did comparisons against main for the `abl_meso_tendency` test, and I got only machine diffs, which is expected.

Reg tests that use explicit diffusion but also have another scalar with multiply_rho = true (e.g., tke) will have diffs, which I confirmed using hbl_godunov.

Once I added the second commit, all explicit diffusion reg tests with nonunity density should have diffs because multiply_rho = true for the icns PDE.

There is a variety of reg tests comparisons we could do to check this, with different equations set to multiply_rho = true or not. However, with such a change, I think we should seek a verification test. CTV or burgraff with nonunity density and explicit diffusion. One set to confirm that the current implementation is wrong, and one to confirm this PR fixes it. Maybe there is a better test, but those two come to mind.

I'll do a representative unit test, too.

marchdf:
amazing work!

marchdf (approved):

moprak-nrel (approved):
Great job tracking this down and for the excellent accompanying doc with the fix!

mbkuhn:
**abl_godunov_explicit** => machine diffs against main. velocity ~ 1e-15, temperature ~ 1e-13
(diffusion_type = 0, unity density. should have no diffs)
**abl_meso_tendency** => decent velocity diff against main (velocityx ~ 1e-7), small temperature diff against main (1e-10)
(density = 1.2)
* In this branch and main, velocity is a multiply_rho = true equation, so there should be consequences for velocity. Temperature, however, is still multiply_rho = false for both this branch and main, so there should be no difference to the diffusion calculations, but there are greater than machine diffs because the velocity is different
* On the main branch, diffs between diffusion_type = 0 (fully explicit) and diffusion_type = 1 (Crank-Nicolson) are 2.1e-9 for temperature and 1.5e-7 for velocityx. Using this PR branch for diffusion_type = 0, the diffs remain 2.1e-9 for temperature and become 8e-11 for velocityx. The smaller difference could potentially indicate this PR branch is more accurate. Hard to tell with such a small difference.
**ow_linear_init_waves** with viscosity set to 1e-2 in both phases
(density ranges from 1 to 1000)
* On the main branch, the velocityx diffs between diffusion_type = 2 (fully implicit) and diffusion_type = 1 (Crank-Nicolson) are 2e-3. The vof diffs are 1e-5. However, the velocityx diffs increase when comparing diffusion_type = 0 (fully explicit) to diffusion_type = 1, going up to 5e-3, even though these diffusion types should be more similar (both are 2nd order). The vof diffs are as big as 1e-3. Using this PR branch for diffusion_type = 0 and comparing to diffusion_type = 1, the velocityx diffs are instead only 1e-4, and the vof diffs are 3e-6. This shows much less of a discrepancy between the different discretizations and much less of an impact on the free surface evolution, which is expected.

PR #1616, opened 2025-05-20
Ability to Have Multiple Instances of Time Averaging and Specify an Averaging Interval

mchurchf:
## Summary

This adds the ability to have multiple time averaging instances running at once along with the ability to select an averaging update frequency.  The ability to select an averaging update frequency allows one to "phase-average" periodic flows if the period is known.  Examples include, flow over waves, flow through a rotor spinning at known frequency, or flow past a floating turbine oscillating at known frequency.

## Pull request type

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [ ] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU 

## Additional background

Issue Number:


mbkuhn:
@mchurchf Is it ok if I replace the use of "frequency" with the word "interval"? not a big deal, it's just nice to be consistent that way.

amr-wind/utilities/averaging/TimeAveraging.cpp
```diff
@@ -71,25 +76,39 @@ const std::string& TimeAveraging::add_averaging(
...
     }
 
     const amrex::Real elapsed_time = (cur_time - m_start_time);
+    m_last_avg_time = cur_time;
+    amrex::Print() << "TimeAveraging: " << m_label << std::endl;
     for (const auto& avg : m_averages) {
-        (*avg)(time, m_filter, elapsed_time);
+        amrex::Print() << "Accumulating average.  Last accumulation done at: "
+                       << m_last_avg_time << ", elapsed time: " << elapsed_time
+                       << std::endl;
```
mbkuhn, line 110: @mchurchf Is there a benefit to having this print statement within the loop? It seems like it would just be repeating the same information for every avg.

mbkuhn:
I could also add "averaging_time_interval" so that it doesn't have to be specified in terms of time steps. The user could choose. I've added similar stuff to other parts of the code in the past, which Regis and Alex appreciated.

mbkuhn:
> "interval" is fine. Yes, having the ability to set the interval as a physical time instead of steps would be better. Ultimately, that's what most users would be doing indirectly through the number of time steps given a fixed dt. If it is not a fixed dt, how do you handle it?

For a variable dt, we end up using the first timestep that meets or exceeds the time interval. So it isn't exact, but it should be close when dt is much smaller than the averaging interval. For plotfiles there is even the option to modify dt so that the time interval occurs exactly, but I'm not sure if we want to add that here. We could potentially add that in the future; once #1592 is merged it'll be more straightforward.

mbkuhn (approved):

PR #1340, opened 2024-11-11
Clipped outflow made default for momentum equation

mbkuhn:
## Summary

This change should have been in #1287, but it was missed. Because of that, inflow at outflow was still automatically allowed in the momentum equation (ICNS). This removes that; changing the default to a clipped outflow (clipping velocities going into the domain)

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe):

## Checklist

The following is included:

- [ ] new unit-test(s)
- [ ] new regression test(s)
- [ ] documentation for new capability

This PR was tested by running:

- the unit tests
  - [ ] on GPU 
  - [ ] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU 

## Additional background

This should actually cause the diffs I expected in the other PR...

PR #1624, opened 2025-05-23
remove hard-coded inflow-on-outflow flag

mbkuhn:
## Summary

I thought this had already been done in #1340 (which followed after #1287), but that was only for the MAC velocity calculation. There should now be no flags turning on allow_inflow_on_outflow; it should only be turned on by an input file argument.

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [ ] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU 

## Additional background

Might cause some small diffs for reg tests that use weno schemes

### DIFF
--- a/amr-wind/equation_systems/icns/icns_advection.H
+++ b/amr-wind/equation_systems/icns/icns_advection.H
@@ -333,11 +333,6 @@ struct AdvectionOp<ICNS, fvm::Godunov>
                     limiter_type = PPM::default_limiter;
                 }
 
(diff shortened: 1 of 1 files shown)


mbkuhn:
@mpolimeno whaddaya know, I think this is why we're seeing flow coming in at the pressure outflow.

mpolimeno (approved):

PR #1628, opened 2025-05-28
Fix docs CI

jrood-nrel:


moprak-nrel:
Jon, I think the error is from incompatible `pybtex` (dependency of `sphinxcontrib-bibtex`) with newer versions of `setuptools` due to a deprecated API call.

moprak-nrel:
Unfortunately, the last release of `pybtex` was in [2021](https://pypi.org/project/pybtex/#history). There is an [issue](https://bitbucket.org/pybtex-devs/pybtex/issues/169/replace-pkg_resources-with) to get a new release, but it appears the package is not actively maintained. A hacky fix for now might be to pin `setuptools` like the CI warning suggests, but I don't like it.

PR #1634, opened 2025-06-02
Separate free surface calculation for diffuse interfaces (overset)

mbkuhn:
## Summary

Instead of using a geometric calculation of the interface location (planar intersection based on VOF reconstruction), this allows a basic trilinear interpolation of VOF to be used to find where 0.5 value occurs. This only turns on where iblank_cell = -1, i.e., parts of the domain that are covered by the Nalu-Wind mesh.

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [x] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [x] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU 

## Additional background

Issue Number:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,12 +18,12 @@ jobs:
     - name: Clone
       uses: actions/checkout@v4
     - name: Check formatting
-      uses: DoozyX/clang-format-lint-action@v0.18.2
+      uses: DoozyX/clang-format-lint-action@v0.20
       with:
         source: './amr-wind ./unit_tests ./tools/utilities'
         exclude: '.'
         extensions: 'H,h,cpp'
-        clangFormatVersion: 18
+        clangFormatVersion: 20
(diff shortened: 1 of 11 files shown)


mbkuhn:
In addition to the unit test, I confirmed that this works well using the sloshing tank case. Now that things can be handled differently in the overset region, the previous errors are eliminated.

[SloshingTankCompare.pdf](https://github.com/user-attachments/files/20577319/SloshingTankCompare.pdf)

mbkuhn:
Also confirmed that the sampler is correct in an off-center location (x=0.2, y=0.1) to make sure the x and y gradients are used.

[SloshingTankCompare_OffCenter.pdf](https://github.com/user-attachments/files/20577525/SloshingTankCompare_OffCenter.pdf)

PR #1651, opened 2025-06-11
Make initial iterations compatible with boundary planes

mbkuhn:
## Summary

Ever since the boundary fills were updated to be more accurate (filling at n+1/2 and n+1), this broke compatibility with the initial iterations because these go forward in time and back. This would lead to having data at n+1/2 when things needed to be filled at n, which isn't possible with the current boundary approach. By changing the calculation of time at n+1/2 to be 0.5*(current time + new time) instead of current time + 0.5*dt, things are made compatible again. During the initial iterations, the new time is the same as the current time, which eliminates the conflicts while still getting the same value during the actual simulation.

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [ ] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU 

## Additional background

Issue Number:

### DIFF
--- a/amr-wind/equation_systems/PDE.H
+++ b/amr-wind/equation_systems/PDE.H
@@ -137,7 +137,9 @@ public:
     {
         BL_PROFILE(
             "amr-wind::" + this->identifier() + "::pre_advection_actions");
-        m_adv_op->preadvect(fstate, m_time.delta_t(), m_time.current_time());
+        m_adv_op->preadvect(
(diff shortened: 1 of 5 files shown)


mbkuhn:
Need to run the reg tests for this

PR #1670, opened 2025-06-23
Document tools; add tools for converting native sampler outputs

mbkuhn:
## Summary

one tool is to create a single txt file for every sample directory (every output time). this would make sense for when there are a lot of sample points

one tool is to create a single txt file for a single sample point, constructing the time series. this would make sense for when there is one, or a few, sample points.

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [ ] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU 

## Additional background

Issue Number:


mbkuhn:
I have "amr-wind path" as an argument for both routines, but that is just to access the amrex particle routines. that may not be necessary, but it makes it easier for a user, I think.

mpolimeno (approved):

mbkuhn:
@marchdf I realized that many people don't know some of these tools exist. Though some are documented individually in different places, it makes sense to have a central list with a brief description. We could add internal links if needed, but I'm not sure that's necessary. While doing this, I noticed that 5 of the python routines are really old (6 years) and many of them interact with a type of output (line_plot) from AMR-Wind that doesn't exist anymore. Or maybe it is the ASCII output from ABLStats. Wondering if we should deprecate them.

marchdf:
Nice work! I wouldn't worry about links right now. I think we can deprecate them if we don't use line_plot anymore. Do you also want to link to the amr-wind-frontend repo as a source for more tools?

PR #1672, opened 2025-06-24
User-Defined Function Refinement 

hgopalan:
## Summary

As the coarse-fine grid boundary size does not matter a lot for  URANS solutions, the default field refinement levels are not always required. Also the existing refinement methods for wakes are limited to box or cylindrical regions. This PR defines a template for creating a blanking region which can be used to perform field refinement similar to the terrain_blank based field refinement for immersed boundaries. The current implementation is restricted to a cylindrical region as a proof of concept and the function can be modified in the future to incorporate analytical wake shape profiles or nacelle or other structures. This is expected to reduce the cell count significantly. 

- [ ] Bugfix
- [X ] Feature
- [ ] Build related changes

## Additional background

Sample Mesh with the blanking 

Screenshot 2025-06-24 at 4 01 47 PM (1)

### DIFF
--- a/amr-wind/physics/CMakeLists.txt
+++ b/amr-wind/physics/CMakeLists.txt
@@ -20,6 +20,7 @@ target_sources(${amr_wind_lib_name}
   ActuatorSourceTagging.cpp
   Intermittency.cpp
   ForestDrag.cpp
+  MetMastMarking.cpp
   )
 
 add_subdirectory(multiphase)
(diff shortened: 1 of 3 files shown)


marchdf:
Thanks for looking into this. I am sorry but I must be missing something. Why not just use https://github.com/Exawind/amr-wind/blob/main/amr-wind/utilities/tagging/CylinderRefiner.cpp ?

marchdf:
We chatted, I have an idea I want to try out.

hgopalan:
Generalized in #1673

PR #1673, opened 2025-06-25
User specified tagging criteria

marchdf:
## Summary

Provide functionality for the user to specify an arbitrary geometrical function for refinement.

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [x] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [ ] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU 

## Additional background

The example I threw in as a reg test (if the number of cells in x/y/z are increased to 128) show how one could refine inside the shell of a sphere, a further restrict it to the bottom half of that sphere:

Screenshot 2025-06-25 at 9 58 48 AM

Closes #1672

### DIFF
--- a/amr-wind/utilities/tagging/CMakeLists.txt
+++ b/amr-wind/utilities/tagging/CMakeLists.txt
@@ -13,4 +13,5 @@ target_sources(${amr_wind_lib_name} PRIVATE
   # Geometry based refinement options
   BoxRefiner.cpp
   CylinderRefiner.cpp
+  UDFRefiner.cpp
   )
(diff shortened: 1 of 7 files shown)


hgopalan:
Thanks for doing this. Makes it more generic than what I had proposed.

hgopalan (approved):

marchdf:
Just noting that AMReX used this as inspiration for: https://github.com/AMReX-Codes/amrex/pull/4534. We don't use `AMRErrorTag` but if we did one day switch to using that then we could simplify this code.

PR #1678, opened 2025-07-03
FreeSurface sampler: avoid issues when sampling within numerical beach

mbkuhn:
## Summary

Because of the volume fraction modification that takes place in numerical beaches and the geometric-VOF assumptions built into the FreeSurface sampler, things can look pretty crazy there. This enables the user to switch to a linear interpolation approach within a specified length from xhi, smoothing things out.

This PR also includes improvements to the linear interpolation approach for finding the free surface location, which translates to its usage in ExaWind simulations. (The linear approach is used in overset regions)

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [x] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [x] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU 

## Additional background

Issue Number:


mbkuhn:
Without the feature in this PR, the noisy vof field within a numerical beach can lead to a noisy free surface plot like this one

[damped_vs_undamped_stokes_second_order.pdf](https://github.com/user-attachments/files/21780339/damped_vs_undamped_stokes_second_order.pdf)

amr-wind/utilities/sampling/FreeSurfaceSampler.cpp
```diff
@@ -607,23 +638,23 @@ bool FreeSurfaceSampler::update_sampling_locations()
                                     const amrex::Real vof_c = vof_arr(i, j, k) +
                                                               slope_0 * dist_0 +
                                                               slope_1 * dist_1;
+                                    // Extrapolate to cell edge (0.5) plus a
+                                    // factor of safety (0.1) because
+                                    // reconstruction is not identical in
+                                    // neighboring cells
                                     const amrex::Real vof_r =
-                                        vof_c + 0.5 * slope_dir_r * dx[dir];
+                                        vof_c + 0.6 * slope_dir_r * dx[dir];
```
mpolimeno, line 646: Is the value of this safety factor (0.1) based on anything theoretical, or is primarily some heuristic based on testing?

mbkuhn: It's just based on testing. The edge cases causing the outliers are pretty rare, but they're caused by only checking for a 0.5 intersect between the current cell center (z) and its edge (z + 0.5*dz). The safety factor extends the check to go 10% into the neighboring cell. 

In the linear interpolation, I don't create a completely continuous reconstruction of vof. What I was seeing in my testing was that one cell might see the vof values range from 0.99999 to 0.500001, while the cell above it would see values from 0.0001 to 0.49999. So then neither cell sees the 0.5 intersect and it gets missed. If I use 1.0 instead of 0.6, it should never miss, but I might end up using a less accurate estimation of the vof location. Because cells get missed so rarely, it seems like the additional 0.1 is enough.

PR #1733, opened 2025-09-05
Add active wake control to Joukowsky Disk models

moprak-nrel:
## Summary

Add the ability to enable active wake control on individual Joukowsky Disk turbines. The implementation is based on: https://doi.org/10.3390/en17040865

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [x] on GPU 
  - [x] on CPU 
- the regression tests
  - [x] on GPU 
  - [x] on CPU 

## Additional background

Issue Number:

### DIFF
--- a/amr-wind/wind_energy/actuator/disk/Joukowsky.H
+++ b/amr-wind/wind_energy/actuator/disk/Joukowsky.H
@@ -40,8 +40,8 @@ struct JoukowskyData : public DiskBaseData
     // TODO: These might have to live in a different struct
     amrex::Real awc_amplitude{0.0};
     amrex::Real awc_angular_frequency{0.0};
-    amrex::Real awc_n{0.0};
-    amrex::Real awc_phi_clock{0.0};
+    amrex::Real awc_azimuthal_mode{0.0};
+    amrex::Real awc_clocking_angle{0.0};
 };
 
 struct Joukowsky : public DiskType
--- a/amr-wind/wind_energy/actuator/disk/Joukowsky.H
+++ b/amr-wind/wind_energy/actuator/disk/Joukowsky.H
@@ -37,7 +37,7 @@ struct JoukowskyData : public DiskBaseData
     vs::Vector disk_force{0.0, 0.0, 0.0};
     bool use_tip_correction{true};
     bool use_root_correction{true};
-    //TODO: These might have to live in a different struct
+    // TODO: These might have to live in a different struct
     amrex::Real awc_amplitude{0.0};
     amrex::Real awc_angular_frequency{0.0};
     amrex::Real awc_n{0.0};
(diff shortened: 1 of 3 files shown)


amr-wind/wind_energy/actuator/disk/Joukowsky.H
```diff
@@ -37,6 +37,10 @@ struct JoukowskyData : public DiskBaseData
     vs::Vector disk_force{0.0, 0.0, 0.0};
     bool use_tip_correction{true};
     bool use_root_correction{true};
+    amrex::Real awc_amplitude{0.0};
+    amrex::Real awc_angular_frequency{0.0};
+    amrex::Real awc_azimuthal_mode{0.0};
```
lawrenceccheung: The azimuthal mode number should be an integer (Boundary layer flows allow for a real spanwise mode number, but for circular geometries it needs to be a signed integer).

docs/sphinx/user/inputs_Actuator.rst
```diff
@@ -395,6 +395,40 @@ Example for ``TurbineFastLine``::
...
+.. input_param:: Actuator.WTG01.awc_amplitude
+
+   **type:** Real, optional, default=0
+
+   Sets the amplitude of the forcing term in AWC
+
+
+.. input_param:: Actuator.WTG01.awc_azimuthal_mode
+
+   **type:** Real, optional, default=0
```
lawrenceccheung: This should be type `int`.

amr-wind/wind_energy/actuator/disk/Joukowsky_ops.H
```diff
@@ -282,8 +286,15 @@ struct ComputeForceOp<Joukowsky, ActSrcDisk>
                 // eq 18
                 moment += x * ddata.radius() * f_theta * dArea;
 
+                const amrex::Real f_awc =
+                    ddata.awc_amplitude *
```
lawrenceccheung, line 290: I just realized that we might have sent you the wrong specification for `awc_amplitude`.  I meant for `awc_amplitude` to be given as a fraction of the total axial force (instead of absolute units).  So to make that change, I would set
```c++
   f_awc = ddata.awc_amplitude * f_z* ... 
```
or make that change below where
```c++ 
  grid.force[ip] = ( (f_z*(1.0 + f_awc) * ddata.normal_vec ...
```

docs/sphinx/user/inputs_Actuator.rst
```diff
@@ -395,6 +395,40 @@ Example for ``TurbineFastLine``::
...
+
+   Sets the temporal angular frequency for AWC (in radians)
+
+
+.. input_param:: Actuator.WTG01.awc_amplitude
+
+   **type:** Real, optional, default=0
+
+   Sets the amplitude of the forcing term in AWC
```
lawrenceccheung: If we do use definition of amplitude above, then we should say that this amplitude is relative to the total force `f_z`.

amr-wind/wind_energy/actuator/disk/Joukowsky_ops.H
moprak-nrel: Got it, just changed it to be relative

PR #1759, opened 2025-10-03
Couple to Kynema for ALM and templatize existing OpenFAST interface

mbkuhn:
## Summary

Needed to implement coupling with Kynema as an aero + turbine model. To avoid duplicating code and to ensure I was doing the right things, I templated the existing OpenFAST interface and made the Kynema interface another specialization of that template. Kynema does a lot less "under the hood" compared to OpenFAST, especially for initialization, so the template specialization is much more involved. The *Iface.cpp and *_types.H files are the main ones that have to be changed depending on the template; the others are simpler and mostly for instantiation.

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [x] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [ ] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU 

## Additional background

Issue Number:

### DIFF
--- a/amr-wind/utilities/FieldPlaneAveragingFine.cpp
+++ b/amr-wind/utilities/FieldPlaneAveragingFine.cpp
@@ -1,4 +1,5 @@
 #include "amr-wind/utilities/FieldPlaneAveragingFine.H"
+#include "amr-wind/utilities/constants.H"
 #include "AMReX_iMultiFab.H"
 #include "AMReX_MultiFabUtil.H"
 #include <algorithm>
@@ -205,6 +206,8 @@ void FPlaneAveragingFine<FType>::compute_averages(const IndexSelector& idxOp)
         ((g0.ProbHi(0) - g0.ProbLo(0)) * (g0.ProbHi(1) - g0.ProbLo(1)) *
          (g0.ProbHi(2) - g0.ProbLo(2)));
 
+    const bool periodic_dir = g0.periodicity().isPeriodic(m_axis);
+
     const auto& mesh = m_field.repo().mesh();
     const int finestLevel = mesh.finestLevel();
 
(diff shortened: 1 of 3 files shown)


mbkuhn:
At this point, the templates for OpenFAST are basically done, though they will probably be tweaked later in the process. Both act_uniform_alm reg tests pass!

mbkuhn:
Derek Slaughter already looked over the Kynema interface calls

amr-wind/wind_energy/actuator/turbine/turbine_types.H
```diff
@@ -26,6 +26,9 @@ struct TurbineInfo : public ActInfo
 using DeviceVecComponent = amrex::Gpu::DeviceVector<ComponentView>;
 struct TurbineBaseData
 {
+    //! Identifier string for turbine solver type
+    std::string solver_name{"Internal"};
```
marchdf, line 30: should we allow for lower case when checking solver_name?

amr-wind/wind_energy/actuator/turbine/external/external_base_types.H
```diff
@@ -0,0 +1,105 @@
...
+    int num_substeps;
+
+    //! Time step index for external
+    int time_index{0};
+
+    //! Checkpoint interval for external
+    int chkpt_interval;
+
+    //! Data access functions that have to be defined for each type
+    virtual float* position_at_vel(int dir) const = 0;
```
marchdf, line 88: `float` is the right type here?

mbkuhn: This is inherited from the types that openfast uses. I could cast openfast functions, I guess, but I just stuck with the same ones.

amr-wind/wind_energy/actuator/turbine/turbine_types.H
mbkuhn: Since camel case is the pattern for actuator types (e.g., TurbineFastLine), I think it makes sense to stick with the capital. But the solver name doesn't get checked against an input argument, it gets checked against the name in the code, so only a developer can make a case mistake, not a user.

PR #1770, opened 2025-10-14
Improve Robustness of Immersed Forcing Wall Function and Method 

hgopalan:
## Summary

In this PR, we make several changes to the immersed forcing method to improve robustness and also improve the accuracy of the wall function. 

1.  The terrain wall-function has been modified to provide wall function capability consistent with the flat terrain. 
2.  The hybrid version of the Kosovic model has also been modified to be terrain-aware. 
3.  The restart runs with a user-specified CFL assumes a flat bottom to compute the time step. However, when a terrain is added the initial steps at the terrain interface requires a smaller time step to stabilize. A delay time is added to ensure that for restart runs, the simulation does not start with CFl based dt but instead starts with an initial dt. Once the delay time elapses, the simulation switches back to CFL based dt or a fixed dt, depending on user preference. 

The time step of immersed forcing simulations are much higher now when we have multiple refinement levels. It is also possible to have terrain touch the outer x and y boundaries without worrying about time step reduction. 

- [X ] Bugfix
- [ ] Feature
- [ ] Build related changes

## Additional background

The existing regression tests and unit tests will be sufficient. 

Issue Number:

### DIFF
--- a/amr-wind/ocean_waves/OceanWaves.cpp
+++ b/amr-wind/ocean_waves/OceanWaves.cpp
@@ -71,6 +71,7 @@ void OceanWaves::post_init_actions()
     BL_PROFILE("amr-wind::ocean_waves::OceanWaves::post_init_actions");
     m_ow_bndry->post_init_actions();
     m_owm->update_target_fields(m_sim.time().current_time());
+    m_owm->update_target_volume_fraction();
     m_ow_bndry->record_boundary_data_time(m_sim.time().current_time());
     if (m_multiphase_mode) {
         m_owm->apply_relax_zones();
@@ -91,6 +92,7 @@ void OceanWaves::pre_advance_work()
     const amrex::Real adv_bdy_time =
         0.5 * (m_sim.time().current_time() + m_sim.time().new_time());
     m_owm->update_target_fields(adv_bdy_time);
+    m_owm->update_target_volume_fraction();
     m_ow_bndry->record_boundary_data_time(adv_bdy_time);
 }
 
@@ -100,6 +102,7 @@ void OceanWaves::pre_predictor_work()
     // Update ow values for boundary fills at new time
     const amrex::Real bdy_fill_time = m_sim.time().new_time();
     m_owm->update_target_fields(bdy_fill_time);
+    m_owm->update_target_volume_fraction();
     m_ow_bndry->record_boundary_data_time(bdy_fill_time);
 }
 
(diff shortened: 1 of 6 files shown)


amr-wind/turbulence/LES/Kosovic.cpp
```diff
@@ -25,7 +26,6 @@ Kosovic<Transport>::Kosovic(CFDSim& sim)
     m_Cs = std::sqrt(8 * (1 + m_Cb) / (27 * M_PI * M_PI));
     m_C1 = std::sqrt(960) * m_Cb / (7 * (1 + m_Cb) * m_Sk);
     m_C2 = m_C1;
-    pp.query("refMOL", m_refMOL);
```
marchdf, line 28: I like that you renamed this but I am worried about backwards compatability. Can you add an abort if someone tries to use this refMOL? There's a pp is present function you can use.

amr-wind/core/SimTime.cpp
```diff
@@ -248,12 +249,21 @@ void SimTime::set_current_cfl(
             dt_new = amrex::min(dt_new, m_initial_dt);
         }
 
-        m_dt[0] = dt_new;
+        if (m_cur_time <= m_delay_time) {
```
marchdf: switch to ternary expression

amr-wind/core/SimTime.cpp
```diff
@@ -248,12 +249,21 @@ void SimTime::set_current_cfl(
...
+            m_dt[0] = m_initial_dt;
+        } else {
+            m_dt[0] = dt_new;
+        }
 
     } else {
         // Ensure that we use user-specified dt. Checkpoint restart might have
         // overridden this
-        m_dt[0] = m_fixed_dt;
+        if (m_cur_time < m_delay_time) {
```
marchdf: switch to ternary expression, and fix the codeql warning

marchdf:
do you have a test you want to add to this? Also there are a lot of tests that still use `Kosovic.refMOL`, going to need to update those.

marchdf:
Just for posterity, these are the diffs I see:
```
         35 - abl_waves_terrain (Failed)                        no_ci regression
         36 - act_abl_joukowskydisk (Failed)                    no_ci regression
         37 - act_abl_uniformctdisk (Failed)                    no_ci regression
        101 - forest_drag (Failed)                              no_ci regression
        122 - abl_kosovic_neutral (Failed)                      no_ci regression
        124 - abl_kosovic_neutral_ib (Failed)                   no_ci regression
        125 - nrel_precursor (Failed)                           no_ci regression
        161 - nrel_terrain (Not Run)                            no_ci regression
        162 - nrel_terrain_amr (Not Run)                        no_ci regression
```

PR #1799, opened 2025-12-04
Drag Forcing changes and lateral damping added to Terrain Drag physics

hgopalan:
When Immersed Boundary Forcing method was developed, there was a need for having sponge layers to ensure that solution was not contaminated. However, with algorithm improvement, this feature became redundant and is being removed. On the other hand for strong inversions, gravity waves were being reflected back from the horizontal boundaries. A new lateral damping has been added within drag forcing to ensure that the gravity waves do not contaminate the solution. 

- [ ] Bugfix
- [X ] Feature
- [X ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ X] Documentation content changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [ ] new regression test(s)
- [X ] documentation for new capability

- the unit tests
  - [X ] on GPU Kestrel 
  - [X] on CPU Kestrel 
- the regression tests
  - [X ] on GPU Kestrel 
  - [ X] on CPU Kestrel

### DIFF
--- a/amr-wind/physics/BurggrafFlow.cpp
+++ b/amr-wind/physics/BurggrafFlow.cpp
@@ -45,7 +45,7 @@ BurggrafFlow::BurggrafFlow(const CFDSim& sim)
         std::ofstream f;
         f.open(m_output_fname.c_str());
         f << std::setw(m_w) << "time" << std::setw(m_w) << "L2_u"
-          << std::setw(m_w) << "L2_v" << std::setw(m_w) << "L2_w" << '\n';
+          << std::setw(m_w) << "L2_v\n";
         f.close();
     }
 }
@@ -175,7 +175,7 @@ void BurggrafFlow::output_error()
     if (amrex::ParallelDescriptor::IOProcessor()) {
         std::ofstream f;
         f.open(m_output_fname.c_str(), std::ios_base::app);
-        f << std::setprecision(12) << m_time.new_time() << std::setw(m_w)
+        f << std::setprecision(12) << std::setw(m_w) << m_time.new_time()
           << std::setw(m_w) << u_err << std::setw(m_w) << v_err << '\n';
         f.close();
     }
--- a/amr-wind/setup/init.cpp
+++ b/amr-wind/setup/init.cpp
@@ -32,7 +32,7 @@ void incflo::CheckAndSetUpDryRun()
         amrex::ParmParse pp("time");
         pp.add("max_step", 0);
         pp.add("plot_interval", 1);
-        pp.add("checkpoint_inteval", (-1));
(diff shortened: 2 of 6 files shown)


amr-wind/equation_systems/icns/source_terms/DragForcing.cpp
```diff
@@ -313,23 +368,27 @@ void DragForcing::operator()(
...
-             ((xstart_damping + xend_damping + ystart_damping + yend_damping) *
-              (uy1 - sponge_density * spongeVelY)));
+             (1 - blank(i, j, k)) * ((xstart_damping + xend_damping +
+                                      ystart_damping + yend_damping) *
+                                     (uy1 - sponge_density * spongeVelY)));
         src_term(i, j, k, 2) -=
             ((CdM * m * (uz1 - target_w) * blank(i, j, k)) +
-             ((xstart_damping + xend_damping + ystart_damping + yend_damping) *
-              (uz1 - sponge_density * spongeVelZ)));
+             (CdM * m * (uz1 - target_w) * drag(i, j, k)) +
```
mbkuhn, line 387: This is a change to the Drag Forcing formulation that has nothing to do with new damping features or sponges. It basically extends the no-penetration condition of a wall into the first non-wall cell. This is not unreasonable, but it's unclear what is the basis of this change.

amr-wind/equation_systems/icns/source_terms/DragForcing.cpp
```diff
@@ -301,8 +356,8 @@ void DragForcing::operator()(
                 (amr_wind::constants::EPS +
                  std::sqrt((ux2r * ux2r) + (uy2r * uy2r)));
             // BC forcing pushes nonrelative velocity toward target velocity
-            bc_forcing_x = -(uxTarget - ux1) / dt;
-            bc_forcing_y = -(uyTarget - uy1) / dt;
+            bc_forcing_x = -(uxTarget - ux1) / (time_factor * dt);
+            bc_forcing_y = -(uyTarget - uy1) / (time_factor * dt);
```
mbkuhn, line 360: These are also notable changes (also present in a couple other files) that affect the time scale of Drag Forcing by adding a default time factor of 5, which relaxes the forcing.

amr-wind/equation_systems/icns/source_terms/DragForcing.cpp
hgopalan: Because we use MOST assuming a flat surface, this usually has vertical velocity set to zero at the cell. This was the reason for the change. Setting it reduces the sharp velocity component at the stair Steppy interface.

PR #1808, opened 2025-12-12
Add inflow_scaling_factor for simple scaling/fine-tuning of ABL boundary velocity fields

manosser:
## Summary
Adds a new optional ABL input parameter:

ABL.inflow_scaling_factor = <real>    # default = 1.0

When reading boundary inflow files, the velocity components (u, v, w) are multiplied by this factor before being injected into the simulation (i.e., 1.1 = +10%, 0.9 = -10% velocity magnitude). All other fields remain unchanged. This enables simple scaling (or fine-tuning) of inflow velocity time-series without having to re-run a precursor, and is particularly useful for custom boundary files that are not generated from AMR-Wind. I have been using this feature here at the Technical University of Munich on our CPU based HPC system.

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [ ] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU 

Manual testing:
Compiled successfully on linux HPC system
Verified that the inflow_scaling_factor is correctly read from the input file
Confirmed that only the "velocity" field is scaled, and other fields remain unchanged (mass_inflow, pressure_outflow)
Confirmed default behavior (inflow_scaling_factor = 1.0) produces identical results to the baseline
Confirmed behavior when the parameter is missing (defaults to 1.0)
Confirmed that negative or zero values are reset to 1.0 safely

image

image

image

## Additional background
This feature allows rapid testing of different inflow magnitudes without regenerating boundary files. It hopefully adds minimal overhead, and preserves all existing solver behavior when the parameter is not set or set to 1.0.
Scaling the inflow in this way likely interferes with other physics in the boundary layer, but for fine tuning, or minimal wind speed changes, it seems to work well. I have only been able to test this on our CPU based system (LRZ supermuc) with OpenMP.
Open for any and all feedback, and please let me know if there is a better way to achieve this function. 
Thanks :)

### DIFF
--- a/amr-wind/core/MLMGOptions.cpp
+++ b/amr-wind/core/MLMGOptions.cpp
@@ -26,6 +26,7 @@ void MLMGOptions::parse_options(const std::string& prefix)
     pp.query("con_grid_size", m_lpinfo.con_grid_size);
     pp.query("max_coarsening_level", m_lpinfo.max_coarsening_level);
     pp.query("max_semicoarsening_level", m_lpinfo.max_semicoarsening_level);
+    pp.query("deterministic", m_lpinfo.deterministic);
 
     // LinOp options
     pp.query("max_order", max_order);
--- a/amr-wind/core/SimTime.H
+++ b/amr-wind/core/SimTime.H
@@ -261,6 +261,12 @@ private:
     //! Smallest absolute tolerance for enforcing dt, considering all
     amrex::Real m_force_dt_abs_tol{1e8};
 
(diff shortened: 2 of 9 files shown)


hgopalan:
Thank you for adding and testing the capability. @mchurchf Any suggestions for making this more general?

PR #1809, opened 2025-12-12
Fine ABL stats

### LINKED ISSUE #1041: Feature request: ABL statistics on level >0
With the recent ability to output boundary data at multiple levels at the same time, we got the ability to run grids that have coarse level 0, and have refinements towards the bottom of the domain. A problem with that is that the ABL statistics only get computed at the level 0. It would be nice to be able to have ABL stats to be calculated at the finest level for every height.

Additionally, I have recently ran into a problem where the sampling of the ABL stats might have an issue with a domain configured with these refinements. I had a case with 80 m level 0, and 3 refinements setup like so
```
3
1
-10000 -10000 0 10000 10000 1500
1
-10000 -10000 0 10000 10000 700
1
-10000 -10000 0 10000 10000 400
```
The potential temperature field had discontinuities, while everything else appears to be okay (including the temperature flux). They appear to happen right around the 400 and 700 m mark, where we have a change in resolution. Could be an issue on the sampling routine. See plot below for illustration
image

### LINKED ISSUE #1790: Missing Feature: ABL statistics on level >0
## Is this feature request related to a problem? Please describe.
This is a request for a missing feature.  Regis pointed out this long, long ago in request 1040, I believe.  I call this a missing feature because this is a feature we had pre-Exawind, so we're not asking for something new.  This feature is something we used in practice all the time with SOWFA.  People are wasting lots of time with work arounds.

## Describe the preferred solution
I know that an issue with this request is "what do we do if Levels > 0 are not carried through the domain horizontally?"  Why do we need to worry about that.  That's up to the user to use correctly.  We just have to assume some level of proficiency amongst the users.

## Additional context
Matt is happy to discuss and help and implemented such a thing before, so please draw upon the knowledge.

mbkuhn:
## Summary

Path laid out for this upgrade

- [x] add input parameter to FieldPlaneAveragingFine that can limit the "finest level" to compute averages on
- [x] replace FieldPlaneAveraging with FieldPlaneAveragingFine
- [x] pass regression tests with "finest level" set to 0 in all of the places that used to feature FieldPlaneAveraging
- [x] update second order, third order moment calculations to be suited to finer levels (add level masking, interpolation)
- [x] add unit tests for second order, third order moment calculations on finer levels
- [x] update instances with "finest level" set to 0 to use finer levels, either by default or by user input (including ABLStats)
- [x] add new regression or unit test for new capability

There are some additional features that could be added now that this is complete, and there may be compatibility issues in other (optional) parts of the code that could be addressed. But the PR is big enough at this point.

User summary: the bulk of the code changes are in FieldPlaneAveraging, but the main user takeaway is that ABL Stats can use levels above the coarsest level. The default behavior is the same, but by setting ABL.stats_max_level, the code will use a different level for the statistics. By changing the max level, the default log_law_height (zref) also changes automatically to the half the cell height on the specified level (still can be overridden by an input parameter if needed). The max level can also be set to -1, which tells the code to use the finest available level.

This is the most straightforward way to get ABL Stats to do things at finer levels. There are a lot of outputs calculated in ABL Stats, and for the vast majority of them, the resolution is based on the resolution of a single FieldPlaneAveraging instance. There are a couple of quantities that are always calculated at the finest available level instead of the default coarsest level -- I will try, in a follow-on PR, to add an option to also calculate things at the finest level and output those as well, which would allow the user to output 2 sets of statistics in a given run.

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [x] new unit-test(s)
- [x] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [ ] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU 

## Additional background

Issue Number: closes #1790, closes #1041


mbkuhn:
For the code as it is, there should be "no" diffs. There are very small diffs for the hbl_godunov case (which uses planar averages of velocity as part of HurricaneForcing) and the abl_godunov_mpl* cases (which uses planar averages of temperature as part of ABLMeanBoussinesq). I think these are small enough to be ignored.

mbkuhn:
Potential additional features:

- multiple instances of ABLStats to instantiate, output; each with a different max level
- along the same lines, specifying the "xhi" and "xlo" within the FieldPlaneAveraging (these would actually be zlo and zhi) to avoid always assuming the full extent of the domain along the normal axis (still always doing full extent in lateral directions)

Potential things to fix:

- assumptions about ABLStats always on the base level: ABLMeanBoussinesq, HurricaneForcing, etc.

jrood-nrel (approved):

rthedin (approved):
This looks good. Thanks for getting to this!
A couple of thoughts:
- You mentioned multiple ABLstats file for multiple levels on a follow on PR. This feature will be super useful
- If I request max level for stats, say, level2, and level2 only covers the bottom of the surface layer, then I will miss out (not be able to easily plot) on things happening above that. I can see such a scenario happening, where you really want finer resolution near the wall but can't afford to go much higher. Right now a single variable controls the actual ABL stat output _and_ what is being used internally for computations of other quantities (models, etc). I guess my point with this is back to my first bullet point above. We would be able to have a finer near-surface-layer layer of refinement, benefit from it, but still be able to obtain the whole vertical profile.

mbkuhn:
I get what you're saying, and that can still inform future changes. But this works slightly differently than what you're saying. Once the max level is set, that defines the resolution of ABLStats for the entire domain. In regions where that level is not available, it will interpolate the existing information to the desired point. E.g., for a mesh that has dz = 0.1 at the finest level, which is only present for the bottom 20% of the domain, and a base level dz = 0.4, ABL Stats will have a resolution of 0.1 for the 100% of the domain in z, even though most of that will be interpolated from the base level data.

The reliance on interpolation is key for when you want stats on a level that doesn't cover the entire xy-plane.

PR #1901, opened 2026-04-02
Changes to LES to enable terrain simulations

rthedin:
These are local changes by @hgopalan that I have used to successfully run the WFIP2 LES cases. They are out-of-date from main. Michael is on the loop regarding this PR. Submitting it so we can all take it from here and get the LES capabilities merged with the recent merge of the RANS one.
I'm happy to keep testing this branch as it evolves to ensure nothing is diverging.

### DIFF
--- a/amr-wind/equation_systems/temperature/temperature_ops.H
+++ b/amr-wind/equation_systems/temperature/temperature_ops.H
@@ -1,0 +1,53 @@
+#ifndef Temperature_OPS_H
+#define Temperature_OPS_H
+
+#include "amr-wind/equation_systems/temperature/temperature.H"
+#include "amr-wind/core/field_ops.H"
+
+namespace amr_wind::pde {
+
+/** Implementation of post-processing actions for temperature equation
+ *  \ingroup tke_eqn
+ */
+template <>
+struct PostSolveOp<Temperature>
+{
+    PostSolveOp(CFDSim& /*unused*/, PDEFields& fields) : m_fields(fields) {}
+
+    void operator()(const amrex::Real time)
+    {
+        // cppcheck-suppress constVariableReference
+        auto& field = m_fields.field;
+        const auto& repo = field.repo();
+        const int nlevels = repo.num_active_levels();
+        for (int lev = 0; lev < nlevels; ++lev) {
+            const auto& field_arrs = field(lev).arrays();
+             const bool has_terrain = repo.int_field_exists("terrain_blank");
+            if (has_terrain) {
+	      amrex::ParmParse pp("DragTempForcing");
+	      amrex::Real m_soil_temperature= 300.0;
+	      pp.query("soil_temperature", m_soil_temperature);
+                const IntField& m_terrain_blank =
+                    repo.get_int_field("terrain_blank");
+                const auto& blank_arrs = m_terrain_blank(lev).arrays();
+                amrex::ParallelFor(
+                    field(lev), [=] AMREX_GPU_DEVICE(
+                                    int nbx, int i, int j, int k) noexcept {
+                        field_arrs[nbx](i, j, k) =
+                            (blank_arrs[nbx](i, j, k) == 1)
+                                ? m_soil_temperature
+                                : field_arrs[nbx](i, j, k);
+                    });
+            }
+        }
+        amrex::Gpu::streamSynchronize();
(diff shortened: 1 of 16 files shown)


amr-wind/equation_systems/temperature/temperature_ops.H
mbkuhn, line 1: @hgopalan this file is effectively unused because it is commented out. I'm guessing this was an idea that didn't quite work out. Let me know if it needs to be included in the new PR for these changes.

amr-wind/turbulence/LES/Kosovic.cpp
```diff
@@ -210,17 +223,18 @@ void Kosovic<Transport>::update_alphaeff(Field& alphaeff)
     auto lam_alpha = (this->m_transport).alpha();
     auto& mu_turb = this->m_mu_turb;
     auto& repo = mu_turb.repo();
-    const amrex::Real muCoeff = (m_monin_obukhov_length < 0) ? 3.0_rt : 1.0_rt;
+    const amrex::Real muCoeff = 3.0;//(m_monin_obukhov_length < 0) ? 3 : 1;
```
mbkuhn, line 226: @hgopalan is this hard-coding necessary? Should I keep it for the future PR? Do you actually want to get rid of the conditional?

amr-wind/turbulence/LES/Kosovic.cpp
```diff
@@ -111,90 +113,101 @@ void Kosovic<Transport>::update_turbulent_viscosity(
...
+                    (has_terrain) ? 1 - blank_arrs[nbx](i, j, k, 0) : 1.0;
+		//                mu_arrs[nbx](i, j, k) *=
+		//                    rho * viscosityScale * turnOff * blankTerrain
+		//		mu_arrs[nbx](i, j, k) = (x3>400)? 1e-5:mu_arrs[nbx](i, j, k);
+        //
+		amrex::Real mut = mu_arrs[nbx](i, j, k) * mu_arrs[nbx](i, j, k);
+		amrex::Real stratification = 
+		  -(gradT_arrs[nbx](i, j, k, 0) * gravity[0] +
+		    gradT_arrs[nbx](i, j, k, 1) * gravity[1] +
+		    gradT_arrs[nbx](i, j, k, 2) * gravity[2])/300.0;
```
mbkuhn, line 163: @hgopalan this 300 is hard-coded. I'm assuming this is meant to be a reference temperature. Should I feed in the reference temperature here? Or is it just a tuning parameter?

mbkuhn:
after we address the comments here and make the associated changes to the newer PR, I'll close this PR.

amr-wind/turbulence/LES/Kosovic.cpp
```diff
@@ -111,90 +113,101 @@ void Kosovic<Transport>::update_turbulent_viscosity(
...
+		amrex::Real non_linear_coeff = 1.0;
+
+		if (stratification > 1e-10) {
```
mbkuhn, line 166: @hgopalan could you explain this conditional or point to a reference? This one has confused Regis and me. Especially with how it both checks and changes stratification.

amr-wind/turbulence/LES/Kosovic.cpp
hgopalan: No 3.0_rt can be a user input.

amr-wind/turbulence/LES/Kosovic.cpp
hgopalan: Yes. Reference Temperature from Boussinesq Buoyancy.

amr-wind/turbulence/LES/Kosovic.cpp
hgopalan: What we are checking is if we have a positive stratification. This happens at two places: surface layer and capping inversion. Surface layer mut is much larger so the effects of std::max(1.0e-10_rt, mut - 3.0_rt * stratification) is less. However, at the edge of the ABL near the capping inversion and within free atmosphere this effect is strong and will affect the results. I should have used a different variable name and instead just reused the stratification variable in lines 182 and 185.

mbkuhn:
The newer PR (#1916) is ready to take the place of this PR. Thank you!

PR #1911, opened 2026-04-23
Channel builder

mbkuhn:
## Summary

New physics module for initialization
- can set up segments of channels according to start and end points
- segment cross-sections can be either ellipses or trapezoids
- cross-sectional dimensions given at start and end of segments, linearly changing between
- can be single-phase or multiphase
- multiphase allows specifying a water level (air above, water below) and a land level (open channel above)
- velocity profile also specified using a flow speed (along segment) and profile can be uniform, linear, or parabolic

Some adjustments of DragForcing are also included for the sake of turning some things off.

This is meant to be a tool to practice channel-like flows. There will still be more PRs to follow - doing multiphase BCs is awkward with what we have right now. Documentation will follow as well.

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [x] new unit-test(s)
- [x] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [ ] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU 

## Additional background

Issue Number:


mbkuhn:
New reg test is a diverging channel in a periodic domain with a body force (pressure gradient) in x. This image shows the channel, with a threshold plotting operator active so that it only shows the fluid cells.
Screenshot 2026-04-24 at 3 22 21 PM

A slice through the channel looks like this, at initialization and then after 100 timesteps:
Screenshot 2026-04-24 at 3 16 10 PM
Screenshot 2026-04-24 at 3 15 22 PM

hgopalan:
Should we think about using EB to mark the regions and adding the flow conditions? Can give a more generic workflow

src/physics/multiphase/ChannelBuilder.cpp
```diff
@@ -0,0 +1,521 @@
...
+    // Dot product with plane normal at start and end
+    const amrex::Real p1 =
+        a * (x - start_x) + b * (y - start_y) + c * (z - start_z);
+    const amrex::Real p2 = a * (x - end_x) + b * (y - end_y) + c * (z - end_z);
+
+    // Point is within planes if it's on opposite sides or on a plane
+    return (p1 * p2 <= 0.0_rt);
+}
+
+[[nodiscard]] AMREX_GPU_HOST_DEVICE amrex::GpuArray<amrex::Real, 3>
```
jrood-nrel, line 67: Possible `AMREX_GPU_HOST_DEVICE` is necessary, but just want to mention we should prefer `AMREX_GPU_DEVICE` whenever possible.

jrood-nrel (approved):

PR #1916, opened 2026-05-01
LES (Kosovic) changes for simulations with terrain

mbkuhn:
## Summary

following after development work by Harish and Regis #1901

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Other (please describe): improvement

## Checklist

The following is included:

- [ ] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [ ] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU 

## Additional background

Issue Number:


mbkuhn:
Will need to address some hard-coded constants before merging, as well as comments posted on original PR.

hgopalan:
The number 3.0_rt can be an input from the file. It was hard coded to match WRF. Same with 1e-10 just a small number. Can also be user input. The 300.0_rt is the reference temperature from the BoussiensqBuoyancy TRef.

hgopalan (approved):

mbkuhn:
Thanks for the approval, @hgopalan. We'll merge this after Regis is done testing it.

mchurchf:
Guys, I went into this code to try to understand where mu_t gets modified near the immersed boundary.  The lack of comments makes it difficult to figure out what part of the code does what unless I were to decipher all of the parts.  Should someone really have to do that?  I think with 20 minutes of effort, the code could be commented in a way that makes it much easier for everyone else to understand the code.

hgopalan:
Line 201-227. We compute Ustar from k+1 and then derive the required mut at the face between k and k-1 and apply at the cell k.

rthedin:
I've been using this PR for the last couple of weeks. All good. Everything seems to be working.

mbkuhn:
the failing reg tests are due to expected changes in the Kosovic model. Good to go.

PR #1917, opened 2026-05-04
add single ghost cell to copy for boundary plane writing

mbkuhn:
## Summary

Important for some features coming down the line, but also fixes a bug for cases that should currently work, but don't.

When there is a mesh refinement that touches the boundary and expands to a larger size (tangentially) within the domain, this ghost cell is needed to prevent unfilled cells being used in the subsequent simulation.

Should not change any results for current working workflows: currently, the ghost cells get filled in a periodic domain because they "overlap" with the other side of the domain. With this, the boundary ghost cells are filled from the ghost cells of the domain, which have just been filled using the periodic boundary condition. Same result.

All regression tests pass.

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [x] on CPU 
- the regression tests
  - [ ] on GPU 
  - [x] on CPU 

## Additional background

Issue Number:


mbkuhn:
currently trying to accomplish the same thing for netcdf planes and haven't figured it out yet

mbkuhn:
after looking into it further, I don't think that it's worth the effort to fix this in the netcdf implementation

mbkuhn:
Screenshot 2026-04-20 at 9 59 07 AM

More background for what this fixes in the current implementation, not considering how it will be useful for future features. In cases where the refinement zone is smaller at the boundary than in the center, the boundary plane still ends up with refinements that span the same extent as in the center. This is due to the use of minBox(). When the boundary cells are initialized, they're given crazy high values to make sure that only filled cells get used in the subsequent simulation. Because the cells that actually border the boundary plane refinements are not actually refined there, those cells do not get filled with valid values from the domain. Then, they're still used as ghost cells in the subsequent simulation, which makes it blow up. Adding a ghost cell to the copy operation here fixes that problem by populating that single blue-colored cell within the boundary, preventing the unfilled neighboring cells from being used.

jrood-nrel (approved):

PR #1920, opened 2026-05-07
Reduce scratch field initializations

jrood-nrel:
## Summary

Move scratch field initializations to higher levels to be able to perform these operations less.

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [ ] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU 

## Additional background

Issue Number:

### DIFF
--- a/src/equation_systems/AdvOp_Godunov.H
+++ b/src/equation_systems/AdvOp_Godunov.H
@@ -4,6 +4,7 @@
 #include <type_traits>
 
 #include "src/convection/Godunov.H"
+#include "src/core/ScratchField.H"
 #include "src/equation_systems/SchemeTraits.H"
 #include "src/equation_systems/PDETraits.H"
 #include "src/equation_systems/PDEOps.H"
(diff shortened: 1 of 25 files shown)


mbkuhn:
I get that the scratch field creation needs to be in post init and post regrid for the cases of starting a simulation or restarting one. But what happens when you are actually running with adaptive mesh refinement and are frequently regridding?

With scratch field creations in post regrid ops, and the scratch fields not going out of scope, do new scratch fields get created every time there is a regrid, growing memory? Or does this implementation work fine?

jrood-nrel:
@marchdf

marchdf:
This should be fine on regrid. It's using a memory pool in either case. So it's less that the reinit of the scratch fields at each time step ws slow, it's more that all the checks around it were slow ("do I have enough memory in the pool/ok I can use it/no I can't/make sure the device in synced up/etc). So the memory footprint won't grow, it just reinits/reuses the memory pool.

mbkuhn (approved):
thanks for the explanation!

PR #1933, opened 2026-05-15
Fix actuator force over-applied by rho in uniform-Ct and Joukowsky disks

arthurpaulino:
The uniform-Ct and Joukowsky actuator-disk implementations wrote the *dimensional* aerodynamic force [N] into grid.force, while wing_ops and turbine_external_utils write the *kinematic* force (force / density).
ActuatorForcing adds grid.force (after Gaussian spreading) into the ICNS src_term, which is later passed through the shared multiply_rho step along with every other momentum source. Every other ICNS source
(Coriolis, BodyForce, ABLForcing, ...) is kinematic [m/s^2]; the two disk implementations were the outliers.

The result: on a constant-density case the disk source term picked up an extra factor of rho relative to the kinematic convention, over-applying the actuator force (a ~22.5% deeper wake at rho = 1.225 kg/m^3 for a
single uniform-Ct disk).

Fix: store the kinematic force in grid.force for both disk types so they match wing_ops and turbine_external_utils. The contribution then enters SrcTermOp in the same [m/s^2] units as every other source and the
existing multiply_rho step converts the whole mixture consistently.

  - uniform_ct_ops.H: aeroPressure no longer multiplies by density.
  - Joukowsky_ops.H: the disk-force expression no longer multiplies by density.

The thrust/power diagnostics (current_power, current_cp, the HELICS power export) keep their own density factor and are unchanged. This leaves the correct paths -- wing_ops, turbine_external_utils, and
ActuatorForcing -- and the FLLC lifting-line correction (which also consumes grid.force in kinematic units) untouched.


mbkuhn:
Thanks for this contribution! I looked deeper into the code to see if this fix would make sense. The actual situation is a bit more complicated than the text implies. There are a number of places in the code that calculate grid.force; some get it right, and some get it wrong. Because of that, this fix alone would cause other problems.

wing_ops calculates the kinematic force from aero tables without bringing in the density, while external_turbine_utils divides the force from an external solver in order to make it kinematic. These are correct, which is fortunate, especially since these were used in the AMR-Wind publication and the latter external pathway is the most frequently used turbine type.

However, as your PR points out, there are still issues with the force calculations in the uniform ct and joukowksy disk implementations. I see two possible methods to a proper fix:

1. Keep the suggested modification of ActuatorForcing.cpp, and make wing_ops and external_turbine_utils calculate the full force instead of the kinematic force (by multiplying by density for the former, by not dividing by density for the latter). - This approach enables grid.force to be true to its name
2. Do not modify ActuatorForcing.cpp, but remove the density factor from the two actuator disk implementations with the units conflict. - This approach leaves the "correct" parts of the code untouched

arthurpaulino:
Thanks! Good catch on the wing/external paths. Went with option 2.

Option 1 reaches further than the producer list shows: FLLCOp also consumes grid.force and assumes kinematic units (its Biot-Savart du ∝ dG/|v| has no density factor). Switching wing_ops to full force breaks FLLC, which would mean editing the lifting-line code + plumbing density into ComponentView/FLLCData. And the wing path would multiply by ρ then divide by ρ again in FLLC — a wash, which says kinematic is the natural internal unit anyway (matches every other ICNS source).

Option 2 only touches the two disk implementations we agree are wrong: wing_ops, turbine_external_utils, ActuatorForcing, FLLC all untouched, so the publication and external-turbine results stand.

- uniform_ct_ops.H / Joukowsky_ops.H — disk force no longer multiplies by density; current_power/current_cp keep theirs.
- Regression goldens for disk cases will shift (expected). total_disk_force output is now kinematic.

PR description updated to reflect the new approach.

mbkuhn (approved):
looks good. thank you!

src/wind_energy/actuator/disk/uniform_ct_ops.H
```diff
@@ -79,9 +79,12 @@ struct ComputeForceOp<UniformCt, ActSrcDisk>
         base::set_thrust_coefficient(ddata, uInfSqr);
 
         const int npts = ddata.num_force_pts;
-        const amrex::Real rho = ddata.density;
+        // Kinematic aero pressure [m^2/s^2]. grid.force is stored in
+        // kinematic units (force / density): ActuatorForcing adds it into
+        // the ICNS src_term in the same [m/s^2] convention as every other
+        // momentum source, ahead of the shared multiply_rho step.
         const amrex::Real aeroPressure =
```
mbkuhn:
```suggestion
        const amrex::Real aeroPressure = 0.5_rt * uInfSqr * ddata.current_ct;
```

src/wind_energy/actuator/disk/uniform_ct_ops.H
```diff
@@ -79,9 +79,12 @@ struct ComputeForceOp<UniformCt, ActSrcDisk>
...
-            0.5_rt * uInfSqr * rho * ddata.current_ct;
+            0.5_rt * uInfSqr * ddata.current_ct;
```
mbkuhn:
```suggestion
```

src/wind_energy/actuator/disk/Joukowsky_ops.H
```diff
@@ -301,9 +301,14 @@ struct ComputeForceOp<Joukowsky, ActSrcDisk>
...
+                // disk force enters ActuatorForcing in the same [m/s^2]
+                // convention as every other ICNS source (see uniform_ct_ops
+                // and ActuatorForcing.cpp). current_power / current_cp below
+                // keep their own density factor.
                 grid.force[ip] = ((1 + f_awc) * f_z * ddata.normal_vec +
                                   f_theta * theta_vec) *
-                                 ddata.density * uInfSqr * dArea;
+                                 uInfSqr * dArea;
 
                 ddata.disk_force = ddata.disk_force + grid.force[ip];
```
mbkuhn:
```suggestion
                ddata.disk_force =
                    ddata.disk_force + ddata.density * grid.force[ip];
```

PR #1934, opened 2026-05-15
Fix corrector MAC velocity-forces over-applied by factor rho

arthurpaulino:
incflo::ApplyCorrector enters with the ICNS src_term in mass-form units `rho * accel` -- the predictor's compute_source_term(NPH) at icns_advance.cpp:367 ends with SrcTermOp<ICNS>::multiply_rho(NPH) which lifts every accumulated kinematic source by density_nph for CompRHSOps::predictor_rhs consumption.

The corrector then calls icns().pre_advection_actions(New) which feeds src_term to HydroUtils::ExtrapVelToFaces in
icns_advection.H:198-203 as the `fq` velocity-forces argument. That half-step coupling expects kinematic acceleration `m/s^2`, not `rho * m/s^2`. Without the divide added here every corrector MAC half-step is over-applied by an extra factor of rho relative to a dimensionally-clean call, so each registered ICNS source (Coriolis, BodyForce, ABLForcing, ActuatorForcing, ...) drives the MAC face velocities harder than it should during the corrector pass.

The predictor handles the equivalent mass-to-kinematic conversion at incflo_advance.cpp:286 via

    kynema_sgf::field_ops::divide(
        icns().fields().src_term, density_old, 0, 0, 1,
        AMREX_SPACEDIM, 0);

before its pre_advection_actions(Old) call. The corrector was missing that step; mirror it here using `density_nph` because the last multiply_rho call lifted by `density_nph` (predictor's compute_source_term(NPH)).

After the corrector's pre_advection_actions, the corrector's compute_source_term(New) at incflo_advance.cpp:622 reassembles src_term from scratch (init + accumulate + multiply_rho), so this predivide does not need a matching post-MAC re-multiply -- the re-assembly overwrites whatever state src_term is left in.

Scope
-----

The fix is scoped to the corrector path only (ApplyCorrector is invoked only when m_use_godunov == false). Production Godunov runs are unaffected. MOL-with-corrector cases will see a different, dimensionally-correct corrector contribution; existing MOL validation goldens may shift slightly and require regen.

### DIFF
--- a/src/incflo_advance.cpp
+++ b/src/incflo_advance.cpp
@@ -542,6 +542,19 @@ void incflo::ApplyCorrector()
     const auto& density_old = density_new.state(kynema_sgf::FieldState::Old);
     auto& density_nph = density_new.state(kynema_sgf::FieldState::NPH);
 
+    // src_term is `rho * accel` after the predictor's multiply_rho;
+    // convert to kinematic for MAC velocity-forces.
+    kynema_sgf::field_ops::divide(
+        icns().fields().src_term, density_nph, 0, 0, 1, AMREX_SPACEDIM, 0);
+
+    const int nghost_force = 1;
+    amrex::IntVect ng(nghost_force);
+    icns().fields().src_term.fillpatch(m_time.current_time(), ng);
+
+    for (auto& eqn : scalar_eqns()) {
+        eqn->fields().src_term.fillpatch(m_time.current_time(), ng);
+    }
+
     // Extrapolate and apply MAC projection for advection velocities
     icns().pre_advection_actions(kynema_sgf::FieldState::New);
 


mbkuhn:
what's missed here is that there are two discretization schemes in the code: Godunov and MOL. MOL is not really supported anymore, and it is the only scheme that uses ApplyCorrector(). The density division in ApplyPredictor() is

1. exclusive to the godunov scheme (there's a conditional)
2. necessary for the godunov scheme to prepare source terms to be used in pre_advection_actions. The MOL scheme has no pre_advection_actions (see preadvect() in AdvOp_MOL.H)

both of these show this division is unnecessary. the source terms, as originally calculated with rho, are the correct formulation for the MOL corrector.

PR #1935, opened 2026-05-15
Fix ICNS diff_term units mismatch in scalar Laplacian variants

arthurpaulino:
ICNSDiffScalarOp::compute_diff_term and
ICNSDiffScalarSegregatedOp::compute_diff_term in icns_diffusion.H each apply MLABec to produce divtau = -beta * div(eta * grad u) = div(eta * grad u) (with beta = -1, mass-form, N/m^3), then run a conditional ParallelFor that divides divtau by rho whenever `fstate != FieldState::New` (i.e. for FieldState::Old and FieldState::NPH). The trailing /rho block leaves divtau in kinematic units m/s^2 for Old and NPH while leaving divtau in mass form N/m^3 for New, even though both states share the same downstream consumers.

The downstream consumers all expect mass-form divtau:

- ApplyPredictor:287-288 builds velocity_forces = src_term + divtau where src_term is mass-form (post compute_source_term(Old) + multiply_rho(Old)). With divtau kinematic, the sum mixes units. The subsequent field_ops::divide(src_term, density_old, ...) at :286 then under-applies the divtau contribution by an extra factor of rho.

- CompRHSOps::predictor_rhs (CompRHSOps.H:104-110) builds fld_new = (rho_o * fld_o + dt * (conv + det_j * src + factor * diff)) / rho with src and conv both in mass form (rho * accel) and diff in kinematic form (rho * accel / rho). The factor * diff term is therefore 1/rho smaller than the dimensionally-clean mass-form value, and the explicit half of CN diffusion is under-applied by factor rho.

Effect on a constant-density rho = 1.225 kg/m^3 ICNS run with Crank-Nicolson diffusion: the explicit-half divtau lands at 1/1.225 ~ 0.816 of its mass-form value, so the total explicit- plus-implicit CN diffusion is roughly 0.5 * 1.000 + 0.5 * 0.816 = 0.908 of physically correct. Visible as a noticeably sharper wake than the same case run with mass-form divtau throughout.

Fix
---

Remove the conditional /rho block from both compute_diff_term variants. divtau now stays in mass form for every FieldState, matching what every downstream consumer (velocity_forces add, predictor_rhs, corrector_rhs, explicit-RK2 diff_new) was already written to expect.

The companion ICNSDiffTensorOp::compute_diff_term (line 18-49) already produced mass-form divtau without the /rho block; the scalar variants now match.

Scope
-----

The fix applies whenever Godunov + CN (or Implicit) diffusion is active, which covers every production turbine LES configuration in this code base. Existing validation goldens encoded the under-applied diffusion behaviour and will need regeneration with the fixed (slightly more diffusive) wakes. The shift is largest near non-unity density (e.g. atmospheric runs with stratification); unit-density unit-tests are not affected.


mbkuhn:
This appears to be correct. When the handling of density was improved / changed in https://github.com/kynema/kynema-sgf/pull/1128, only one of these compute_diff_term instances was updated -- this PR addresses the other two that were missed.

mbkuhn (approved):

PR #1938, opened 2026-05-18
Reg test adjustments

### LINKED ISSUE #1918: Test dependency issue with abl_bndry_output_native
## Bug description
If netcdf is not enabled,  `abl_bndry_output_native` fails because it depends on an input file `abl_bndry_output` that is only turned on when built with netcdf. My recommendation would be to remove that depencency so the `native` always works.

Error (paste or attach):
```
75/125 Test  #75: abl_bndry_output_native .................***Failed    0.56 sec
                Kynema-SGF (https://github.com/kynema/kynema-sgf)

  Kynema-SGF version :: v3.9.0-16-g3e752144-CLEAN
  Kynema-SGF Git SHA :: 3e75214409bec5977de05eb38911f9dc1e13fbbc-CLEAN
  AMReX version    :: 26.03-110-g388b0db16adb

  Exec. time       :: Wed May  6 14:33:18 2026
  Build time       :: May  6 2026 14:11:23
  C++ compiler     :: GNU 13.3.0

  MPI              :: OFF
  GPU              :: ON    (Backend: CUDA)
  OpenMP           :: OFF

  No additional third-party libraries enabled

 See https://github.com/Kynema/kynema-sgf/blob/development/LICENSE for details.

amrex::Error::0::Couldn't open file: ../abl_bndry_output/abl_bndry_output.inp !!!
SIGABRT
See Backtrace.0 file for details
```

This triggers other failures:
```
The following tests FAILED:
         75 - abl_bndry_output_native (Failed)
         76 - abl_bndry_output_amr_native (Failed)
        117 - abl_bndry_input_native (Not Run)
        118 - abl_bndry_input_native_inout (Not Run)
        120 - abl_bndry_input_amr_native (Not Run)
        121 - abl_bndry_input_amr_native_xhi (Not Run)
        122 - abl_bndry_input_amr_native_mlbc (Not Run)
Errors while running CTest
```

mbkuhn:
## Summary

- Fixes cases where a basic (no netcdf dependency) test relied on netcdf-dependent tests
- Makes consistent the need for fcompare to be active when doing precursor-dependent tests "_red()"
- Puts more tests into the CI category

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Other (please describe): CI design

## Checklist

The following is included:

- [ ] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [ ] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU 

## Additional background

Issue Number: closes #1918


src/wind_energy/actuator/ActuatorContainer.cpp
```diff
@@ -260,10 +260,16 @@ void ActuatorContainer::populate_field_buffers()
         amrex::Gpu::deviceToHost, buff_device.begin(), buff_device.end(),
         buff_host.begin());
 #ifdef AMREX_USE_MPI
-    const int num_entires = static_cast<int>(buff_host.size());
+    const int num_entries = static_cast<int>(buff_host.size());
+#ifdef AMREX_USE_FLOAT
     MPI_Allreduce(
-        MPI_IN_PLACE, buff_host.data(), num_entires, MPI_DOUBLE, MPI_SUM,
```
jrood-nrel, line 265: I think for single precision we probably just want to cast to double for this.

mbkuhn: this was the change that enabled passing the single precision check, though - had a segfault otherwise. is there a different way to do something similar?

jrood-nrel: I think `amrex::ParallelDescriptor::Mpi_typemap<amrex::Real>::type()` could avoid the ifdef. It should return `MPI_DOUBLE` or `MPI_FLOAT` according to `amrex::Real`.

jrood-nrel (approved):

src/wind_energy/actuator/ActuatorContainer.cpp
mbkuhn: oh cool. I'll try that.

mbkuhn: good call. looks like that works.

jrood-nrel: So my initial thought on it was wrong as I didn't realize it was an enum, which explains my strange first comment, but then I assumed AMReX would already have something that exists to do this.

PR #1939, opened 2026-05-18
Adding Blanking using AMReX-EB 

hgopalan:
## Summary

The terrain marking module is based on a vertical height algorithm and cannot handle floating structures in domain. Also currently it is computed as a binary int field and this  causes stair-steppy behavior. 

This PR adds the capability to use two alternative drag methods: (i) a simple partial terrain drag to reduce the stair-steppy behavioe and (ii) the volume fraction option in EB to add shapes using EB implicit function, EB composite function or EB STL reader. 

These modules will operate independently to provide backwards compatibility and reduce if loops in the codes. 

- [ ] Bugfix
- [X ] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [ ] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU 

## Additional background

Picture1


hgopalan:
@rybchuk FYI

hgopalan:
DataCenter
Idealized Data Center

mbkuhn:
This is cool and impressive. It seems like a ton of duplicate code from what we have already though? Is there any of the code that could be put into a utility file so it can be reused and also make some of the individual files easier to read?

Also, are there any quantitative results from these capabilities that we could use to check their accuracy or implementation?

hgopalan:
I added it as a placeholder
Code for showing the feasibility of using EB for
Complex geometries. The reason I had it as a separate code is that for each new capability of EB implicit function, we may have a code change and the existing terrain drag already has too many ifs. 

We can discuss in detail on how the capability should be setup
And I can look into it.

PR #1945, opened 2026-05-20
more detail on delay and start time parameters

mbkuhn:
## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [ ] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU 

## Additional background

Issue Number:


rthedin (approved):
The extra info on the documentations is appreciated! I think it's pretty clear now and the example you have under plot_delay is useful.

As I read through it, I realized that for post-processing routines, we have `output_from_restart`, which to me seems equivalent to setting `[plot,checkpoint]_time_delay` and `[plot,checkpoint]_start_time` to the same start time (analogous for the time-step based variables). Could be worthwhile supporting some sort of `time.[plot,checkpoint]_output_from_restart` if there is a desire to keep consistency across inputs.

PR #1955, opened 2026-05-27
Subvolume: coarser grids, spanning cells

mbkuhn:
## Summary

Enable the subvolume output utility to create a distinct box for output, not identical to any box on the grid. This is to allow output of grids coarser than the coarsest mesh level and vary the mesh resolution (by integer multiples) from the mesh grid.

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [x] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [x] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU 

## Additional background

requested by @rthedin 

necessary for going from any Kynema-SGF simulations with non-unity integer aspect ratio to FAST.Farm, which requires unity aspect ratio


hgopalan:
I realized that there is no input file for reference for users who may not be familiar with the feature. Can we add one in regtest?

mbkuhn:
whoops - there were originally changes to the reg test to show that this can be done. I just added them back.

hgopalan (approved):

PR #1956, opened 2026-05-28
Sampling: interpolation options

mbkuhn:
## Summary

Two new similar options based on user requests

- interpolation order 0: sampling (of any sampler) will output the value of the nearest field value on the mesh (cell-centered value for most fields). This is requested at the top-level sampling instance, not at the level of individual samplers. Because this is only a choice about interpolation, there is no additional information output about the locations of the values being output. The default behavior is interpolation_order = 1 (linear interpolation), as it's always been.
- snap to cell center: option for static samplers (PlaneSampler, LineSampler, ProbeSampler, and VolumeSampler) to snap to mesh cell centers. Because this actually changes the locations of the points, the new locations will be output with the sampling data. Currently, this only takes place at initialization and the point locations remain the same for the entire simulation; they do not change in the case of an adaptive regrid.

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [ ] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU 

## Additional background

@rthedin


src/utilities/sampling/SamplingUtils.cpp
```diff
@@ -8,6 +9,20 @@ using namespace amrex::literals;
...
+amrex::Real snap_to_nearest_cell_center(
+    const amrex::Geometry& geom, const int dir, const amrex::Real x)
+{
+    const auto& plo = geom.ProbLoArray();
+    const auto& dx = geom.CellSizeArray();
+    const auto& dxi = geom.InvCellSizeArray();
+    const auto& dom = geom.Domain();
+
+    int idx = static_cast<int>(std::lround((x - plo[dir]) * dxi[dir] - 0.5_rt));
+    idx = std::max(dom.smallEnd(dir), std::min(dom.bigEnd(dir), idx));
```
jrood-nrel, line 21: I think we should use amrex::max and min to stay consistent. If there isn't an lround, don't bother changing that one. Maybe one day we could go back to std.

jrood-nrel (approved):

src/utilities/sampling/SamplingUtils.cpp
mbkuhn: whoops, I missed this one. I'll add it to a new PR

PR #1962, opened 2026-06-02
ABLStats: relative zref

mbkuhn:
## Summary

Code currently assumes that the wall is at z = 0 in ABLStats:

- this affects the computation of key averages within ABLStats, which are computed by default at z = dz/2 (dz of the coarsest mesh level) regardless of the actual location of the zlo boundary
- log_law_height can change the location of these averages, but this parameter is meant to be a height, not a position, and it is used elsewhere in the Monin-Obukhov model
- that prompted me to add a wall position argument, so the calculation of the averages can be modified if desired, without incorrectly changing the zref parameter within the MO part of the code.

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [ ] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU 

## Additional background

Issue Number:

### DIFF
--- a/docs/sphinx/user/inputs_ABL.rst
+++ b/docs/sphinx/user/inputs_ABL.rst
@@ -48,7 +48,7 @@ This section is for setting atmospheric boundary layer parameters.
    When not specified, this is taken from the lower problem-domain boundary in the
    selected normal direction. This value is used together with
    :input_param:`ABL.log_law_height` to determine the sampling location for the
-   planar averages used in thewall model.
+   planar averages used in the wall model.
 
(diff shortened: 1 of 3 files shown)


rthedin (approved):
Haven't tried running the code itself, but the change seems very straightforward. Thanks for fixing this little detail!

jrood-nrel (approved):

PR #1976, opened 2026-06-23
Heterogeneous Forest: point cloud, one forest per file

mbkuhn:
## Summary

Current forest implementation was very limited; only allowed for cylindrical forests with leaf area density varying with z. This PR adds a new forest capability where each forest is specified using a point cloud with leaf area density. The code interpolates among the nearest points to get the leaf area density at a given mesh cell.

The boundaries of a forest are determined by the following two checks calculated at a given mesh cell:

1. Lateral boundaries: check if the current location is within or on the boundary of the 2D convex hull of the forest point cloud
2. Vertical boundaries: check if the current location is above the heights of the closest points from the forest point cloud.

There are two parameters for modifying the algorithm: the number of nearest points to use for the leaf area density interpolation and a tolerance for proximity to a point where the point's value will be used instead of interpolated.

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [x] new unit-test(s)
- [x] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [x] on CPU 
- the regression tests
  - [ ] on GPU 
  - [x] on CPU 

## Additional background

Issue Number:


mbkuhn:
Screenshot 2026-06-23 at 5 00 46 PM

mbkuhn:
From meeting with NORDEX folks, one of the needs for this capability is being able to accommodate different data input formats. Best approach at the moment, though, is to get this merged and add different readers in the future.

hgopalan (commented):
The implementation is unbounded. In the empirical model the tree distribution was fitted to an approximate leaf area index ( LAI)

LAI = \int Lm F(z) dz over a vertical cylinder of tree height height and diameter D. 

So on coarse or fine grids we will recover an almost equivalent LAI. In here we have F(x,y,z) and there is no leaf area index. I think when they do a point cloud scanning they get individual LAD at different locations but also a LAI. We would need to do something like 

LAI = int F(x,y,z) dx dy dz to ensure that the tree distribution remains bounded for different grid resolutions

mbkuhn:
How is this unbounded? The boundaries of the forest are determined by the point cloud, which does not change with resolution. Because of this, the LAI, the integral of LAD, shouldn't have different bounds though it will look slightly different depending on the grid resolution (which is also true for the cylinder approach).

The researchers who asked for this would like to work purely in terms of LAD instead of LAI.

src/physics/ForestDrag.cpp
```diff
@@ -126,4 +346,149 @@ amrex::Vector<Forest> ForestDrag::read_forest(const int level) const
...
+            ForestPoint pt;
+            if (!(p_stream >> pt.m_x >> pt.m_y >> pt.m_z >> pt.m_lad)) {
+                amrex::Abort(
+                    "Malformed point entry in " + cloud_file +
+                    ". Expected: x y z lad");
+            }
+
+            points.push_back(pt);
+            xy_points.push_back({pt.m_x, pt.m_y});
+            xmin = std::min(xmin, pt.m_x);
```
jrood-nrel: I think we're still using `amrex::min` mostly because it has a better history on GPUs, so we should use that for consistency.

hgopalan:
The Lm in the empirical model is calculated from LAI / \int F(z) dz. The model is exact while numerical implementation wise there will be smearing depending on the vertical resolution. 

With point cloud my concern is we have a cloud of points which we interpolate on the cell using IDW and the distribution depends on the IDW kernel which performs the function of the Lm parameter and also possible asymmetrical and sparse point clouds. The simulation would run fine but we would need to come up with a way to quantify the optimal IDW to use to obtain the best possible solution. This is something I referred as unbounded since there is no exact expression to bound the total tree distribution. 

We can go ahead and test it and see what works best.

hgopalan (approved):

jrood-nrel (approved):

PR #1997, opened 2026-07-14
ChannelBuilder: apply setVal() only to the current level

mbkuhn:
## Summary

initialize_fields() is called for every level. Within that function, I was calling a setVal() for the whole field. That means when the function was called on the finest level, it was erasing work done on the coarser levels.

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [ ] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU 

## Additional background

Issue Number:

### DIFF
--- a/src/physics/multiphase/ChannelBuilder.cpp
+++ b/src/physics/multiphase/ChannelBuilder.cpp
@@ -174,7 +174,6 @@ ChannelBuilder::ChannelBuilder(CFDSim& sim)
 
     m_sim.io_manager().register_output_int_var("terrain_blank");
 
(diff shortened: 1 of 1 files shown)


mbkuhn:
Screenshot 2026-07-14 at 12 35 59 PM

Example of how the velocity field is screwed up with this bug; it's only correct at the finest level.

PR #2001, opened 2026-07-15
First prototype of the drone interface

tonyinme:
## Summary

Adding a new interface for drones to combine propellers. 

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [X] new unit-test(s)
- [X] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [x] on CPU 
- the regression tests
  - [ ] on GPU 
  - [x] on CPU 

## Additional background

Issue Number:


tonyinme:
@ndevelder , here is the first draft of the drone interface that we discussed.

tonyinme:
This PR is ready for review. 

The failed test is expected. This PR corrects the actuator-sector behavior, so the previous reference results are no longer valid and should be updated.

src/core/vs/quaternion.H
```diff
@@ -0,0 +1,173 @@
...
+/** Scalar-first unit quaternion used to interpolate and compose rotations. */
+struct Quaternion
+{
+    amrex::Real w{1.0_rt};
+    amrex::Real x{0.0_rt};
+    amrex::Real y{0.0_rt};
+    amrex::Real z{0.0_rt};
+};
+
+inline Quaternion normalized(Quaternion q)
```
jrood-nrel: I think we should just let the compiler do any inlining and not do it ourselves.

tonyinme: I rearranged the code. That inline was needed because the function definition was in the header file, but I have adjusted the definitions of the quaternion.

src/core/vs/quaternion.cpp
```diff
@@ -0,0 +1,102 @@
...
+#include <limits>
+
+namespace kynema_sgf::vs {
+
+using namespace amrex::literals;
+
+Quaternion& Quaternion::normalize()
+{
+    const amrex::Real norm = std::sqrt(w * w + x * x + y * y + z * z);
+    if (norm <= std::numeric_limits<amrex::Real>::epsilon()) {
```
d-montgomery: Minor and nitpicky, but this is included in the constants namespace `constants::EPS` and is already used elsewhere in this file.

d-montgomery (commented):
This is more of a general comment, but would it make sense to restructure the actuator module into a generic framework layer (`src/actuator/`) and separate domain-specific applications (`src/applications/wind_energy/`, `src/applications/aerospace/`)? This would make it clearer which components are reusable versus domain-specific, and would naturally place drone under aerospace rather than wind_energy. Something like this:

~~~
src/
├── actuator/
│   ├── disk/
│   ├── drone/ 
│   ├── motion/
│   ├── sector/
│   ├── turbine/
│   ├── wing/ 
│   ├── actuator_types.H  
│   └── ...
├── applications/ 
│   └── aerospace/ 
│       └── drone/
│   ├── hydrokinetic/
│   ├── wind_energy/
└── ...
~~~

Not something that needs to be addressed here, but worth considering as we expand SGF into different application domains.

src/core/vs/quaternion.cpp
tonyinme: Good catch, I have adjusted it and pushed the latest code. The tolerance epsilon was used in a few places and all are fixed now. 

I agree with the comment regarding the code structure, but that is a decision that the code maintainers need to make. This PR will use the current structure of the code.

src/core/vs/quaternion.cpp
```diff
@@ -0,0 +1,101 @@
...
+    amrex::Real cosine = dot(a, b);
+    // At exactly 180 degrees the interpolation path is not unique. Ignore a
+    // roundoff-sized negative dot product so equivalent RPY and quaternion
+    // inputs choose the same path.
+    if (cosine < -constants::EPS) {
+        b = {-b.w, -b.x, -b.y, -b.z};
+        cosine = -cosine;
+    } else if (cosine < 0.0_rt) {
+        cosine = 0.0_rt;
+    }
```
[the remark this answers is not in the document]
tonyinme: This has been fixed. There was an issue with the interpolation when there is 180deg separation between angles. The code will now give a warning when that is the case.

mbkuhn:
This produces diffs on the regression tests (act_sector, act_sector_quad) that were added in your last commit. Are those diffs expected?

tonyinme:
> This produces diffs on the regression tests (act_sector, act_sector_quad) that were added in your last commit. Are those diffs expected?

Yes, we made some changes in this PR that resulted in changes to the gold files. There were some changes in quaternion handling that leads to small differences and another bug fix in the actuator implementation. 

This is all correct as of this latest version and the gold files for those 2 tests need to be updated.

mbkuhn:
Sounds great. I'll skim over the code, and, unless I find something, it should be good to go. After it's approved, could you update it with the latest changes to main? Then it'll be ready to get merged.

We might to a follow-on PR with the restructuring that Dave proposed, so heads up for that. Though I don't know how that could affect your other PR in draft.

docs/sphinx/user/inputs_Actuator.rst
```diff
@@ -26,8 +26,8 @@ turbines as actuator disks and actuator line models.
 
    This string identifies the type of actuator to use. The ones currently
    supported are: ``UniformCtDisk``, ``JoukowskyDisk``, ``TurbineFastLine``,
-   ``TurbineFastDisk``, ``TurbineKynemaLine``, ``TurbineKynemaFMBLine``,
-   ``TurbineKynemaFMBDisk``, ``FixedWingLine``, and ``ActuatorSector``.
+   ``TurbineFastDisk``, ``TurbineKynemaLine``, ``FixedWingLine``,
+   ``ActuatorSector``, and ``Drone``.
```
mbkuhn:
```suggestion
   ``TurbineFastDisk``, ``TurbineKynemaFMBLine``, ``TurbineKynemaFMBDisk``, ``FixedWingLine``,
   ``ActuatorSector``, and ``Drone``.
```

mbkuhn (approved):

docs/sphinx/user/inputs_Actuator.rst
mbkuhn: Only thing I noticed was this minor docs list fix, which you should be able to commit directly from here

tonyinme:
Sounds good. Yes, the next PR builds on top of this one. 
As soon as this one is merged, I will rebase and mark #2011 as ready for review. 

Is it possible to do the restructuring after that? If not, we will need to restructure #2011

mbkuhn:
Yeah, I think that would be fine!

Could you update this branch? (Eliminate the "This branch is out-of-date..." note)

tonyinme:
> Yeah, I think that would be fine!
> 
> Could you update this branch? (Eliminate the "This branch is out-of-date..." note)

perfect, thanks! 

I have updated the branch.

PR #2005, opened 2026-07-21
ChannelBuilder: only re-initialize terrain-related fields 

mbkuhn:
(stop changing velocity after regrid)

## Summary

ChannelBuilder used TerrainDrag as a template, which applies initialize_fields within post_regrid_actions. This makes sense for the sake of the terrain related fields (terrain_blank, terrain_drag) so those fields have as much accuracy as possible given the current mesh. However, ChannelBuilder typically sets other fields within initialize_fields, which should not change after a regrid. This PR remedies that by disabling those fields from changing when called from post_regrid_actions.

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [ ] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU 

## Additional background

Should fix restart-related issues from ChannelBuilder.

### DIFF
--- a/src/physics/multiphase/ChannelBuilder.H
+++ b/src/physics/multiphase/ChannelBuilder.H
@@ -114,6 +114,7 @@ private:
     bool m_initialize_velocity{true};
     bool m_zero_blanked_velocity{true};
     bool m_initialize_drag{false};
+    bool m_terrain_fields_only{false};
     amrex::Real m_rho_init{1.0_rt};
     amrex::Real m_water_level{0.0_rt};
     amrex::Real m_land_level{0.0_rt};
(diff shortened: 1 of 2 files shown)


d-montgomery:
Thanks for adding this fix @mbkuhn! I was observing errors after restarting from check file and noticed the error was reverting back to the error at 10 seconds as shown in this first plot where the restart time is 393 seconds:
laminar_flow_slanted_before_PR

This PR resolved the issue:
laminar_flow_slanted_after_PR

d-montgomery (approved):

PR #2006, opened 2026-07-22
Logic operators between refinement tags

mbkuhn:
## Summary

Sometimes you want more complex tagging criteria than is available. One way to do this is use multiple tagging criteria with logic operations between. With this approach, you could set FieldRefinement to be on only inside or outside a particular geometric region.

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [x] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [x] on CPU 
- the regression tests
  - [ ] on GPU 
  - [x] on CPU 

## Additional background

Issue Number:


mbkuhn:
Should add a reg test

src/utilities/tagging/GeometryRefinement.cpp
```diff
@@ -62,10 +62,25 @@ void GeometryRefinement::operator()(
             const auto& bxa = utils::realbox_to_box(gg->bound_box(), geom);
             const auto& bxi = bx & bxa;
             if (bxi.isEmpty()) {
+                // Operators OR and AND_NOT will not be affected by an empty box
+                // Operators AND and OR_NOT require modification of tag array
+                if (tag_operator() == tagging::TaggingOperator::AND) {
+                    // Because entire box is false, set all tags to false
```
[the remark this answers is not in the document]
mbkuhn: Pretty sure that's how it's supposed to work.

mbkuhn:
I thought this might come in handy for the estuary flows. For example, if you want to refine the terrain interface only underwater, you could combine two field refinement criteria:

1. Gradient of terrain_blank
2. Value of vof

docs/sphinx/user/inputs_tagging.rst
```diff
@@ -72,66 +74,22 @@ Example::
...
-   will be active. By default the bounding box will span the entire domain.
-
 Refinement using geometry
 `````````````````````````
 
 This section controls refinement using pre-defined geometric shapes. Currently,
 two options are supported: 1. ``box`` -- refines the region inside a hexahedral
 block, 2. ``cylinder`` -- refines the region inside a cylindrical block, and
 3. ``udf`` -- refines along an analytical function depending on time and spatial coordinate.
+With the exception of the ``udf`` able to describe temporal variations, geometry-based refinement is static.
```
d-montgomery: This is being wordsmithy, but maybe something like this:
"With the exception of the ``udf``, _**which is able to prescribe temporal variations**,_ geometry-based refinement is static.

docs/sphinx/user/inputs_tagging.rst
```diff
@@ -287,25 +290,54 @@ Example::
...
+  tagging.hole.c1.start = 500.0 500.0 250.0
+  tagging.hole.c1.end = 500.0 500.0 750.0
+  tagging.hole.c1.outer_radius = 300.0
+
+  tagging.geometry.type = FieldRefinement
+  tagging.geometry.operator = and
+  tagging.geometry.field_name = density
+  tagging.geometry.grad_error = 0.1 0.1 0.1
+
+In this example, ``base`` creates an initial tagged region.
```
d-montgomery, line 339: Thanks for adding this summary of the example!

unit_tests/utilities/test_refinement.cpp
```diff
@@ -141,4 +179,94 @@ TEST_F(NestRefineTest, bbox_limits)
...
+    }
+    {
+        amrex::ParmParse pp("tagging.t_and");
+        pp.add("type", (std::string) "DummyOperatorRefinement");
+        pp.add("operator", (std::string) "and");
+    }
+    {
+        amrex::ParmParse pp("tagging.t_or");
+        pp.add("type", (std::string) "DummyOperatorRefinement");
+        pp.add("operator", (std::string) "OR");
```
d-montgomery, line 248: Same here, is this supposed to be `or`?

unit_tests/utilities/test_refinement.cpp
```diff
@@ -141,4 +179,94 @@ TEST_F(NestRefineTest, bbox_limits)
     EXPECT_EQ(bx.bigEnd(), big_end.diagShift(1));
 }
 
+TEST(RefinementTaggingLogic, string_to_operator)
+{
+    using kynema_sgf::tagging::string_to_operator;
+    using kynema_sgf::tagging::TaggingOperator;
+
+    EXPECT_EQ(string_to_operator("and"), TaggingOperator::AND);
+    EXPECT_EQ(string_to_operator("OR"), TaggingOperator::OR);
```
d-montgomery: Are these supposed to be `or` and `and_not`?

unit_tests/utilities/test_refinement.cpp
```diff
@@ -141,4 +179,94 @@ TEST_F(NestRefineTest, bbox_limits)
...
+    }
+    {
+        amrex::ParmParse pp("tagging.t_and_not");
+        pp.add("type", (std::string) "DummyOperatorRefinement");
+        pp.add("operator", (std::string) "and_not");
+    }
+    {
+        amrex::ParmParse pp("tagging.t_or_not");
+        pp.add("type", (std::string) "DummyOperatorRefinement");
+        pp.add("operator", (std::string) "Or_Not");
```
d-montgomery, line 258: `or_not`?

d-montgomery (commented):
This is a great idea, @mbkuhn! I left a few comments for you.

mbkuhn:
Great comments, @d-montgomery! I'll get those fixed. Thanks for the wordsmithing, too -- I've been relying on copilot a lot more lately for documentation, so those wording errors can pop up!

unit_tests/utilities/test_refinement.cpp
mbkuhn: This one is testing that, at the user input level, the operator is not case-sensitive.

unit_tests/utilities/test_refinement.cpp
mbkuhn: After I modified things based on the copilot review, these do need to change. Thanks!

d-montgomery (approved):
Looks good now. Thanks for adding this feature, @mbkuhn!

PR #2010, opened 2026-07-31
Avg fix

d-montgomery:
## Summary

This fixes a bug in the time averaging post processing. `m_accumulated_avg_time_interval += cur_dt` is passed as the `avg_time_interval` parameter to the operator `ReAveraging::operator()`. During restart from check file, `TimeAveraging::post_advance_work` is called before the time step is updated and the default is `cur_dt = -1`. This resulted in `m_accumulated_avg_time_interval` being set to -1, which then caused the resulting averaged field to be negative. This was fixed by adding a check in `TimeAveraging::post_advance_work` to not update `m_accumulated_avg_time_interval` if `cur_dt < 0`.

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [ ] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU 

## Additional background

Issue Number:


mbkuhn (approved):
nice find!

PR #2018, opened 2026-08-13
provide option to skip initializing terrain in ChannelBuilder

mbkuhn:
## Summary

For cases when there is a terrain file for the channel, thus TerrainDrag is going to be used, but there is not a good way to initialize the flow. ChannelBuilder can still be used for the sake of initializing the interface and the velocity. Though the channels won't necessarily match between the two physics classes, this should be a helpful way to give the flow a good start.

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [ ] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU 

## Additional background

Issue Number:


d-montgomery (approved):
Thanks for adding this, @mbkuhn!

PR #2021, opened 2026-08-19
Improvements to make field-based refinements work as expected at initialization (for estuary flows)

mbkuhn:
## Summary

For most multiphase flow cases, the initializer in a physics class sets up a levelset field (e.g., SloshingTank, ChannelBuilder, OceanWaves), then MultiPhase physics (in post_init) converts the levelset field to vof and then calculates the density from the vof field. That means that the vof field isn't calculated by the time it is being used for a field refinement, which is undesirable. The solution is to put those post_init actions direcly into initialize_fields(). This PR does that for ChannelBuilder, which we are currently using as the estuary flow initializer.

There is also a conditional in TerrainDrag that prevents cells within the bottom boundary as being identified as terrain. That leads to a gradient being identified there by gradient field refinement, which is undesirable. Turning that off - need to confirm with tests that it doesn't mess up terrain runs (I don't expect it to matter).

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [ ] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [ ] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU 

## Additional background

Issue Number:

### DIFF
--- a/src/physics/TerrainDrag.cpp
+++ b/src/physics/TerrainDrag.cpp
@@ -160,8 +160,7 @@ void TerrainDrag::initialize_fields(int level, const amrex::Geometry& geom)
             const amrex::Real terrainHt = interp::bilinear(
                 xterrain_ptr, xterrain_ptr + xterrain_size, yterrain_ptr,
                 yterrain_ptr + yterrain_size, zterrain_ptr, x, y);
-            levelBlanking[nbx](i, j, k, 0) =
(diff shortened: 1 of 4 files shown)


d-montgomery (approved):
Thanks for doing this reorganization @mbkuhn. Looks good to me!

PR #2033, opened 2026-09-10
Update vof values within terrain

mbkuhn:
## Summary

IBFM is not a true wall implementation, and the strong inertia or weight of water pushes vof into the wall cells, which affects the interface position for IBFM cases with waves or sea level rise. To address this, this PR fills in the vof values within the terrain to match the values in the non-terrain cells. It is effectively a Neumann condition for vof in the terrain cells. Also, because the vof determines density, this helps push back against hydrostatic forces pushing fluid into non-fluid cells.

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [x] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [ ] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU 

## Additional background

Will affect regression tests

### DIFF
--- a/docs/sphinx/user/inputs_multiphase.rst
+++ b/docs/sphinx/user/inputs_multiphase.rst
@@ -52,8 +52,8 @@ terrain surface.
 
 The extrapolation is performed after the vof solve and before the density is computed from vof:
 - Blanked cells with at least one lateral neighbor (in x or y) that is not blanked take the average
-  vof of those internal unblanked neighbors. Values on the x or y domain boundary are not used as
-  sources.
+   vof of those unblanked neighbors. Valid cells on the x or y domain boundary may be used as
+   sources, but ghost cells outside the domain are excluded.
 - Blanked cells that have no non-blanked lateral neighbors take the average of the laterally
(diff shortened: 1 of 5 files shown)


mbkuhn:
Non-unity vof is somehow originating in terrain cells well below the surface. As the vof values go toward 0, away from 1, bubbles form and float upward, spuriously contaminating the flow.

Screenshot 2026-09-11 at 12 58 51 PM

Screenshot 2026-09-11 at 12 56 53 PM

mbkuhn:
So far my guesses have been that non-unity vof comes from floating point errors and from coarser cells seeing non-unity values due to their size and extrapolating those outward. For the first problem, I added a snap to unity for vof near 1 (didn't work), and for the second, I restricted the method to only apply to non-covered cells (didn't work). Non-unity vof is still contaminating the terrain well below the surface. It would be good to replicate this issue on a smaller case.

Screenshot 2026-09-11 at 4 27 54 PM

PR #2038, opened 2026-09-14
ALM Chord-Based Point Distribution

mchurchf:
## Summary

OpenFAST's interface to CFD has the ability (from long ago) to use a non-uniform actuator line force point spacing that is based on blade chord.  This is useful when ALM force projection width (epsilon) is specified as a fraction of local blade chord.  As chord goes down, body force projection width decreases, meaning actuator points need to be spaced more closely.  This can be done using the uniform spacing that is currently the only option and choosing a spacing that is the smallest required, but this means far more actuator points are called for than necessary over the larger chord portions of the blades.  This is inefficient as the there are more blade points to search for within the CFD mesh and over which to loop for body force projection.  My calculations show that by using chord-based spacing, the number of actuator points can be reduced by a factor of 4-5.  This capability has been available for a long time in SOWFA, just never turned on in AMR-Wind.  This PR allows AMR-Wind to use that capability.

## Pull request type

Draft

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes

## Checklist

The following is included:

- [X] new unit-test(s)
- [X] new regression test(s)
- [X] documentation for new capability

On CPUs and plotting output and comparing to desired behavior.

- the unit tests
  - [ ] on GPU 
  - [ ] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU 

## Additional background

Issue Number:

### DIFF
--- a/src/equation_systems/temperature/source_terms/PerturbationForcing.cpp
+++ b/src/equation_systems/temperature/source_terms/PerturbationForcing.cpp
@@ -33,6 +33,10 @@ void PerturbationForcing::operator()(
         amrex::GpuArray<amrex::Real, AMREX_SPACEDIM> end{
             m_end[0], m_end[1], m_end[2]};
         amrex::RealBox pert_box(start.data(), end.data());
+        amrex::Print() << "Temperature perturbations:" << std::endl;
+        amrex::Print() << pert_box.lo(0) << " " << pert_box.lo(1) << " " << pert_box.lo(2) << std::endl;
+        amrex::Print() << pert_box.hi(0) << " " << pert_box.hi(1) << " " << pert_box.hi(2) << std::endl;
+        amrex::Print() << pert_box.volume() << std::endl;
         const bool has_terrain =
             this->m_sim.repo().int_field_exists("terrain_blank");
         const auto& geom = m_mesh.Geom(lev);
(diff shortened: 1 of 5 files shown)


rthedin (commented):
How does this compare with non-uniformly-spaced FLLC actuator points, in terms of both results and computational performance?

src/equation_systems/temperature/source_terms/PerturbationForcing.cpp
```diff
@@ -65,9 +69,10 @@ void PerturbationForcing::operator()(
                                       : 0.0_rt;
                     const amrex::Real x = prob_lo[0] + ((i + 0.5_rt) * dx[0]);
                     const amrex::Real y = prob_lo[1] + ((j + 0.5_rt) * dx[1]);
-                    const amrex::Real z = amrex::max<amrex::Real>(
-                        prob_lo[2] + ((k + 0.5_rt) * dx[2]) - height_arr_cell,
-                        0.5_rt * dx[2]);
+                    const amrex::Real z = prob_lo[2] + ((k + 0.5_rt) * dx[2]);
+                  //const amrex::Real z = amrex::max<amrex::Real>(
+                  //    prob_lo[2] + ((k + 0.5_rt) * dx[2]) - height_arr_cell,
+                  //    0.5_rt * dx[2]);
```
rthedin, line 75: This is incorrect. The old, more complicated expression is terrain-aware. I suggest we revert back.

